PR #1, opened 2015-12-04
[gerrit] Add backend implementation.

acs:


perceval/backends/gerrit.py
```diff
@@ -0,0 +1,315 @@
...
+#
+# Authors:
+#   Alvaro del Castillo San Felix <acs@bitergia.com>
+#
+
+'''Gerrit backend for Perseval'''
+
+
+from datetime import datetime
+from dateutil import parser
```
sduenas: This module is not used. It should be removed.

perceval/backends/gerrit.py
```diff
@@ -0,0 +1,315 @@
...
+from ..errors import BackendError, CacheError
+from ..utils import DEFAULT_DATETIME, str_to_datetime
+
+
+
+class Gerrit(Backend):
+
+    name = "gerrit"
+
+    def __init__(self, user = None, url = None, nreviews = None,
```
sduenas: It shouldn't be there any spaces around keywords and assignements.
For instance : `url=None`

perceval/backends/gerrit.py
```diff
@@ -0,0 +1,315 @@
...
+                 cache = None, **nouse):
```
sduenas: I think we should forbid the use of packed arguments (like `**nouse`) in any child of `Backend` class. Parameters should be passed explicitly.

perceval/backends/gerrit.py
```diff
@@ -0,0 +1,315 @@
...
+        super().__init__(cache=cache)
+        self.repository = url
+        self.nreviews = nreviews
+        self.last_item = None  # Start item for next iteration
+        self.more_updates = True  # To check if reviews are updates
+        self.number_results = self.nreviews  # To control if there are more items
+        self.client = GerritClient(self.repository, user, nreviews)
+
+
```
sduenas: The next two methods have to be removed because they don't make any sense on this backend.

perceval/backends/gerrit.py
```diff
@@ -0,0 +1,315 @@
...
+        reviews = self.get_reviews(from_date)
+        self._push_cache_queue(reviews)
+
+        while reviews:
+            issue = reviews.pop(0)
+            yield issue
+
+            if not reviews:
+                reviews = self.get_reviews(from_date)
+                self._push_cache_queue(reviews)
```
sduenas: As I've understood from the code below, you're storing the parsed items in the cache, not the raw data. The purpose of the cache is to store the raw items that are retrieved from the repositories.

perceval/backends/gerrit.py
```diff
@@ -0,0 +1,315 @@
...
+
+        :raises CacheError: raised when an error occurs accesing the
+            cache
+        """
+        if not self.cache:
+            raise CacheError(cause="cache instance was not provided")
+
+        cache_items = self.cache.retrieve()
+
+        while True:
```
sduenas: This whole loop is unnecessary. Just iterate over the `cache_items` generator and you got it.
```python
for item in cache_items:
    for review in item:
        yield item
```

perceval/backends/gerrit.py
```diff
@@ -0,0 +1,315 @@
...
+            yield issue
+
+            if not reviews:
+                reviews = self.get_reviews(from_date)
+                self._push_cache_queue(reviews)
+                self._flush_cache_queue()
+
+
+    def fetch_from_cache(self):
+        """Fetch the bugs from the cache.
```
sduenas: Bugs? I think it should be `reviews` or `patches`.

perceval/backends/gerrit.py
```diff
@@ -0,0 +1,315 @@
...
+                    break
+            except StopIteration:
+                break
+
+            for item in raw_items:
+                yield item
+
+
+
+    def get_reviews(self, from_date = None):
```
sduenas: This method mixes retrieval and parsing processes. IMHO both functionalities must be split in two. The parsing should be in a static class method, as we do in Bugzilla backend.

sduenas:
@acs , I made some comments to your code. I think you have to fix them before continuing with the review.

perceval/backends/gerrit.py
acs: This semantic should be in some place: howto get the id for the repository and howto get the unique id for the items. But probably it should be outside Perceval as you propose. I will rework the integration with p2o to understand the best place to put this logic/semantic in the toolchain.

perceval/backends/gerrit.py
acs: The need for this comes from how the backends are used in p2o. Take a look to:

https://github.com/acs/GrimoireELK/blob/master/utils/p2o.py#L69

p2o uses the backends in a very opaque way so the same code works for all backends. p2o knows that with a list of command line params to execute the tool, all needed arguments for the backend are included, but it does not know what params are for the backend and what params for other pieces. This is why this support is needed. The backend takes its needed params, that are specified in the __init__ (API level), and drops the rest of params.

perceval/backends/gerrit.py
acs: As I commented before, the code for the cache should be the same in most cases for backends. And this is why I have not changed the code, nor the params, to make clear that this code is exactly the same in gerrit and in github (and in most of the future backends).

perceval/backends/gerrit.py
acs: Yes, you are right. The cache should store the results from:

https://github.com/acs/perceval-1/blob/gerrit/perceval/backends/gerrit.py#L126

I need to fix that, and the logic related to cache loading/storing. Probably. gerrit logic for cache won't be generic anymore. Interesting. Good point!

perceval/backends/gerrit.py
acs: Great!

perceval/backends/gerrit.py
acs: Yes, I need to work on that. I have tried to isolate it. but looking with detail, yes, some of the parsing logic is mixed here with retrieval logic.

perceval/backends/gerrit.py
acs: Yes!

perceval/backends/gerrit.py
acs: Ok, this is the kind of issues that using pep8 before commit will be catched. In any case, my fault for not do it! Also, I will try to take care of this inside the IDE (Eclipse) to avoid this noise in review process.

acs:
@sduenas yes, it is clear that:

* cache should store real raw data
* retrieval and parsing logic should be isolated
* getting raw data from cache simplified
* get_field_id and get_id moved to other place in the toolchain

I will work on it probably next week. I will use this feedback also to review github backend to avoid loosing time in the review process.

acs:
After uploading new commits, some of the comments are outdated automatically which should not be, like the parse vs retrieval logic. The PEP-8 fixing modify a lot of parts of the code, so it makes harder the review process. In any case, in my comment above I have the summary or items to be completed.

perceval/backends/gerrit.py
```diff
@@ -0,0 +1,308 @@
...
+        """ Get all reviews from repository """
+
+        reviews = []
+
+        if self.number_results < self.nreviews or not self.more_updates:
+            # No more reviews after last iteration
+            return reviews
+
+        task_init = time()
+        raw_data = self.client.get_items(self.last_item)
```
acs: @sduenas you said that retrieval and parsing logic are mixed here. Not sure. The retrieval logic is inside self.client.get_items. We get the raw data here. And then the parsing logic starts. But here, the iteration for the retrieval of reviews is also here. And probably this is the error. get_reviews should read from a generator inside GerritClient and then parse tha raw data that comes from this generator. 

The same approach that should be done in the cache once the code is fixed to store real raw data in the cache.

acs:
@sduenas I have changed all code according to your review. For some reason the cache is not getting the data correctly, but the implementation now stores the raw data. I will fix that later.

Time for a new review?

The logic is now cleaner.

perceval/backends/gerrit.py
```diff
@@ -0,0 +1,297 @@
...
+    def fetch(self, from_date=DEFAULT_DATETIME):
+        self._purge_cache_queue()
+
+        self.last_item = None
+        reviews = self._get_reviews()
+
+        while reviews:
+            review = reviews.pop(0)
+            updated = datetime.fromtimestamp(review['lastUpdated'])
+            if updated <= from_date:
```
sduenas: Why do we need this check?

perceval/backends/gerrit.py
```diff
@@ -0,0 +1,297 @@
...
+        raw_data = self.client.get_items(self.last_item)
+        self._push_cache_queue(raw_data)
+        self._flush_cache_queue()
+        reviews = self._parse_reviews(raw_data)
+        self.last_item = self.client.get_next_item(self.last_item, reviews[-1])
+        logging.info("Received %i reviews in %.2fs" % (len(reviews),
+                                                       time()-task_init))
+        return reviews
+
+    def _parse_reviews(self, raw_data):
```
sduenas: This method should be static, so any user only interested on parsing reviews from Gerrit would be able to do it without creating an instance of the class.

perceval/backends/gerrit.py
```diff
@@ -0,0 +1,297 @@
...
+
+class GerritClient():
+
+    def __init__(self, repository, user, nreviews):
+        self.gerrit_user = user
+        self.nreviews = nreviews
+        self.repository = repository
+        self.project = None
+        self.version = None
+        self.gerrit_cmd = "ssh -p 29418 %s@%s" % (self.gerrit_user,
```
sduenas: On next versions, we should explore other ways to run an ssh connection instead of calling `subprocess`

perceval/backends/gerrit.py
```diff
@@ -0,0 +1,297 @@
...
+                                                  self.repository)
+        self.gerrit_cmd += " gerrit "
```
sduenas: I think it makes more sense to have this kind of strings as class strings. If something change, we only need to change the values of these variables and not look for where I use the string. For instance:
```
CMD_GERRIT = 'gerrit'
CMV_VERSION = 'version'
PORT = '29418'
```

perceval/backends/gerrit.py
```diff
@@ -0,0 +1,297 @@
...
+
+    def _get_version(self):
```
sduenas: This method should be renamed to `version`. Others outside from the class might consider to be useful to know the version of the gerrit server calling to this method.

perceval/backends/gerrit.py
```diff
@@ -0,0 +1,297 @@
...
+
+        if last_item is not None:
+            if gerrit_version[0] == 2 and gerrit_version[1] >= 9:
+                cmd += " --start=" + str(last_item)
+            else:
+                cmd += " resume_sortkey:" + last_item
+
+        return cmd
+
+    def get_items(self, last_item):
```
sduenas: I suggest to renamed it to `reviews`. `get_` prefix are from Java world ;-)

perceval/backends/gerrit.py
```diff
@@ -0,0 +1,297 @@
...
+        cmd = self._get_gerrit_cmd(last_item)
+
+        logging.debug(cmd)
+        raw_data = subprocess.check_output(cmd, shell=True)
+        raw_data = str(raw_data, "UTF-8")
+
+        return raw_data
+
+    def get_next_item(self, last_item, entry):
```
sduenas: Does the method return the next item? I think not. It should be renamed.

perceval/backends/gerrit.py
```diff
@@ -0,0 +1,297 @@
...
+        if self.number_results and self.number_results < self.nreviews:
+            # No more reviews
+            return reviews
+
+        task_init = time()
+        raw_data = self.client.get_items(self.last_item)
+        self._push_cache_queue(raw_data)
+        self._flush_cache_queue()
+        reviews = self._parse_reviews(raw_data)
+        self.last_item = self.client.get_next_item(self.last_item, reviews[-1])
```
sduenas: Is it necessary to store the last item in an attribute of the instance? It makes impossible to run two or more `fetch()` calls at the same time.

perceval/backends/gerrit.py
```diff
@@ -0,0 +1,297 @@
...
+        # Join isolated reviews in JSON in array for parsing
+        items_raw = "[" + raw_data.replace("\n", ",") + "]"
+        items_raw = items_raw.replace(",]", "]")
+        items = json.loads(items_raw)
+        reviews = []
+
+        for item in items:
+            if 'rowCount' in item.keys():
+                # stats data, not a review
+                self.number_results = item['rowCount']
```
sduenas: Same comment as with `self.last_item`

perceval/backends/gerrit.py
```diff
@@ -0,0 +1,297 @@
...
+
+        cmd = self.gerrit_cmd + " version "
+
+        logging.debug("Getting version: %s" % (cmd))
+        raw_data = subprocess.check_output(cmd, shell=True)
+        raw_data = str(raw_data, "UTF-8")
+        logging.debug("Gerrit version: %s" % (raw_data))
+
+        # output: gerrit version 2.10-rc1-988-g333a9dd
+        m = re.match("gerrit version (\d+)\.(\d+).*", raw_data)
```
sduenas: This regular exp pattern should be stored on a class variable. Use the prefix `r''` to tag it as a regular expression.

sduenas:
@acs, I've just added more comments to your patches. Please check them.

perceval/backends/gerrit.py
acs: The reviews in gerrit are retrieved using lastUpdated desc as order. When we find a review that the lastUpdated is older than the from_date, then we don't need to check any more reviews because the rest are older than the from_date (the initial date from which we want reviews).

perceval/backends/gerrit.py
acs: Ok, I will change it!

perceval/backends/gerrit.py
acs: This is the version after some iterations trying to find the best approach. But ok, it we find a better approach, let's use it!

perceval/backends/gerrit.py
acs: Ok, I will change it! It will be very strange that this change, but for example, if we use the same values in several places, it is better as you say to use those values as constants to reflect that it is the same values in all places.

perceval/backends/gerrit.py
acs: So remove "_" (make it public) and also rename from get_version to version?

perceval/backends/gerrit.py
acs: Ok, I will remove the "get_" :)

About the name, I like "items" in order to have a common API in all backends. What do you think?

perceval/backends/gerrit.py
acs: It is the "next_item" for the "next_iteration". So probably, a better name is:

next_iteration_item or next_bucket_item using "elastic" terminology and avoiding mixing iterators with iterations :)

Hmmm, maybe better: next_retrieve_group_item.

perceval/backends/gerrit.py
acs: Hmmm, maybe if you want more than one fetch, you could use different instances?

Maybe we can have a more complex data structure to support several fetches in the same instance (with an array of last_item for example), but I would prefer to keep this simple approach.

perceval/backends/gerrit.py
acs: Same response.

perceval/backends/gerrit.py
acs: Ok, I will change that!

perceval/backends/gerrit.py
sduenas: Yeah, I understand it but my point is that we call `get_next_item` to get which hash we have to ask for the next buch of reviews. So, as far as I understand, this should be removed, isn't it?

perceval/backends/gerrit.py
sduenas: I was thinking on a library but maybe this is the easiest way.

perceval/backends/gerrit.py
sduenas: Just `version`. get is for java coders.

perceval/backends/gerrit.py
sduenas: I like items too but I'm afraid it's impossible since we will have clients that won't return just items. Please check the code from Bugzilla backend. We have `buglist`, `bugs`, `bug_activity`. In GitHub backend we will have issues, pull_requests, even users!

perceval/backends/gerrit.py
sduenas: I think `last_item` can be stored outside of this method. In `fetch()`, for instance, and then,`get_reviews` will receive `last_item` as a parameter.

perceval/backends/gerrit.py
acs: No, we need it to detect when to stop parsing reviews. It is the stop condition.

perceval/backends/gerrit.py
acs: Hmmm, we can not change it if we don't find a way to manage:

```
self.number_results
self.last_item += 1
```
in other way.

perceval/backends/gerrit.py
acs: Done!

perceval/backends/gerrit.py
acs: Done!

perceval/backends/gerrit.py
sduenas: So, does it means that without this exit condition reviews are fetching till infinite? I'm I don't understand well how reviews are fetched. Can we fetch reviews from a giving date? or do we have to fetch them all and the check the dates?

perceval/backends/gerrit.py
sduenas: For instance on `fetch()`:
```python
  last_item = None
  reviews = self._get_reviews(last_item)
  while reviews:
      [...]
      last_item = self.client.get_next_item(last_item,  review)
      
      if not reviews:
          self._get_reviews(last_item)
```

And you will have to remove the calls to `get_next_item` on `__get_reviews()`

perceval/backends/gerrit.py
acs: We have to fetch them all in buckets and then check the dates and stop getting new buckets if the reviews are not updated after from_date.

perceval/backends/gerrit.py
acs: Finally we are using ```@property``` here. Much cleaner!

Done!

perceval/backends/gerrit.py
acs: Done!

perceval/backends/gerrit.py
acs: Done!

perceval/backends/gerrit.py
acs: Done!

perceval/backends/gerrit.py
acs: Done!

perceval/backends/gerrit.py
acs: Done!

perceval/backends/gerrit.py
acs: Done?

perceval/backends/gerrit.py
acs: Done!

perceval/backends/gerrit.py
acs: Fixed as you proposed @sduenas . Really good point!

Done!

acs:
@sduenas all your review reflected in the code now. A final review iteration? :)

acs:
Fixed the pending cache logic. @sduenas something more needed? The change about nreviews (number of reviews pero ssh api query), I think it should be an instance (or maybe a class) param because it is a param from command line (the same for all the program).

perceval/backends/gerrit.py
```diff
@@ -0,0 +1,295 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
+#
+# Authors:
+#   Alvaro del Castillo San Felix <acs@bitergia.com>
+#
+
+'''Gerrit backend for Perseval'''
```
sduenas: Use docstrings with double quotes as in others parts of the tool. Please update doc, too. There're plenty of comments with references to "bugs" when the word should be "reviews". In addition, you have to add some documentation to your classes and public methods.

perceval/backends/gerrit.py
```diff
@@ -0,0 +1,295 @@
...
+
+from ..backend import Backend, BackendCommand
+from ..cache import Cache
+from ..errors import BackendError, CacheError
+from ..utils import DEFAULT_DATETIME, str_to_datetime
+
+
+class Gerrit(Backend):
+
+    name = "gerrit"
```
sduenas: So far, this wasn't needed so remove it from the code.

perceval/backends/gerrit.py
```diff
@@ -0,0 +1,295 @@
...
+
+    def __init__(self, user=None, url=None, nreviews=None,
+                 cache=None, **nouse):
```
sduenas: I still think we have to get sticky to the plan and not use wildcards in backend classes, or at least, not like these. So, please remove `nouse` from the method. If we need it, we will figure out how to do include these things on the code.

perceval/backends/gerrit.py
acs: Done! Probably we will need it later for the backend Registry and also, in arthur. But we can remove it and add it when needed.

perceval/backends/gerrit.py
```diff
@@ -0,0 +1,295 @@
...
+        super().__init__(cache=cache)
+        self.repository = url
+        self.nreviews = nreviews
```
sduenas: I agree with you in `nreviews` should be a instance attribute but please, use `max_reviews` instead. Moreover, we need a default vaule, so declare it as `MAX_REVIEWS` on the top of this module. Check `bugzilla` backend that has a similar schema.

perceval/backends/gerrit.py
acs: Done! Some integrations issues wait when integrating with arthur, but I think in possible ways to avoid this use of wildcards. Done.

perceval/backends/gerrit.py
```diff
@@ -0,0 +1,295 @@
...
+        :raises CacheError: raised when an error occurs accesing the
+            cache
+        """
+        if not self.cache:
+            raise CacheError(cause="cache instance was not provided")
+
+        cache_items = self.cache.retrieve()
+
+        for raw_items in cache_items:
+            reviews = Gerrit.parse_reviews(raw_items)
```
sduenas: Use `self.parse_reviews` instead of `Gerrit.parse_reviews`. If you use `Gerrit` as the caller, you're forcing derived classes to work with Gerrit class method, so no overriding won't be possible.

perceval/backends/gerrit.py
acs: Converted all to use double quotes. About the doc, I have updated all "bugs" occurrences to "reviews.. And about the classes and public methods doc, I am not sure we need to copy and paste the same doc when documenting the public backend API, for example. What do you think? I can copy and paste bugzilla doc if you like this approach! (and adapt it if needed, but if it is backend API it should be exactly the same).

I will review other public API in the backend, it should be strange to have this kind of API, in order to document it!

perceval/backends/gerrit.py
```diff
@@ -0,0 +1,295 @@
...
+        last_item = self.client.next_retrieve_group_item()
+        reviews = self._get_reviews(last_item)
+        last_nreviews = len(reviews)
+
+        while reviews:
+            review = reviews.pop(0)
+            last_item += 1
+            updated = datetime.fromtimestamp(review['lastUpdated'])
+            if updated <= from_date:
+                logging.debug("No more updates for %s" % (self.repository))
```
sduenas: As I read on many sites, it is recommended to declare a logger at the top of the module and then, use it for logging instead of using the module because this way it writes on the root logger.
You have to create a logger using `logger = logging.getLogger(__name__)' and then write on it using `logger.debug()`, for instance.

perceval/backends/gerrit.py
acs: Good point. So self is changed to a class variable, Gerrit, and not to the instance, in this context. Done!

perceval/backends/gerrit.py
sduenas: El lun, 21-12-2015 a las 08:48 -0800, Alvaro del Castillo escribió:
> In perceval/backends/gerrit.py:
> > +# (at your option) any later version.
> > +#
> > +# This program is distributed in the hope that it will be useful,
> > +# but WITHOUT ANY WARRANTY; without even the implied warranty of
> > +# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
> > +# GNU General Public License for more details.
> > +#
> > +# You should have received a copy of the GNU General Public
> License
> > +# along with this program; if not, write to the Free Software
> > +# Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-
> 1307, USA.
> > +#
> > +# Authors:
> > +#   Alvaro del Castillo San Felix <acs@bitergia.com>
> > +#
> > +
> > +'''Gerrit backend for Perseval'''
> Converted all to use double quotes. About the doc, I have updated all
> "bugs" occurrences to "reviews.. And about the classes and public
> methods doc, I am not sure we need to copy and paste the same doc
> when documenting the public backend API, for example. What do you
> think? I can copy and paste bugzilla doc if you like this approach!
> (and adapt it if needed, but if it is backend API it should be
> exactly the same).
> I will review other public API in the backend, it should be strange
> to have this kind of API, in order to document it!
No, I meant you have to include documentation saying which the
attributes to initialize the Gerrit backend are needed or what the
fetch method does, etc.

perceval/backends/gerrit.py
acs: El lun, 21-12-2015 a las 08:53 -0800, Santiago Dueñas escribió:
> In perceval/backends/gerrit.py:
> > +                yield review
> > +
> > +
> > +    def fetch(self, from_date=DEFAULT_DATETIME):
> > +        self._purge_cache_queue()
> > +
> > +        last_item = self.client.next_retrieve_group_item()
> > +        reviews = self._get_reviews(last_item)
> > +        last_nreviews = len(reviews)
> > +
> > +        while reviews:
> > +            review = reviews.pop(0)
> > +            last_item += 1
> > +            updated =
> datetime.fromtimestamp(review['lastUpdated'])
> > +            if updated <= from_date:
> > +                logging.debug("No more updates for %s" %
> (self.repository))
> As I read on many sites, it is recommended to declare a logger at the
> top of the module and then, use it for logging instead of using the
> module because this way it writes on the root logger.
> You have to create a logger using logger =
> logging.getLogger(__name__)' and then write on it
> usinglogger.debug()`, for instance.

I think we can start with a global" Grimoire" for all of our platform,
and then breakdown if needed. I am afraid that logging config will be
complex if we start isolating in modules too much the logs.

> —
> Reply to this email directly or view it on GitHub.

-- 
Alvaro del Castillo San Félix
acs@bitergia.com - Chief Technical Officer (CTO)
http://www.bitergia.com
"Software metrics for your peace of mind"

sduenas: I don't see it. Perceval is a standalone tool that can work as library as well. When it is used as library, you need to know what the library is doing. If you use 'grimoire' as the logger name for each Grimoire library, it's impossible to filter or hard to know what each component is doing at a point.

acs: El mar, 22-12-2015 a las 03:19 -0800, Santiago Dueñas escribió:
> In perceval/backends/gerrit.py:
> > +                yield review
> > +
> > +
> > +    def fetch(self, from_date=DEFAULT_DATETIME):
> > +        self._purge_cache_queue()
> > +
> > +        last_item = self.client.next_retrieve_group_item()
> > +        reviews = self._get_reviews(last_item)
> > +        last_nreviews = len(reviews)
> > +
> > +        while reviews:
> > +            review = reviews.pop(0)
> > +            last_item += 1
> > +            updated =
> datetime.fromtimestamp(review['lastUpdated'])
> > +            if updated <= from_date:
> > +                logging.debug("No more updates for %s" %
> (self.repository))
> I don't see it. Perceval is a standalone tool that can work as
> library as well. When it is used as library, you need to know what
> the library is doing. If you use 'grimoire' as the logger name for
> each Grimoire library, it's impossible to filter or hard to know what
> each component is doing at a point.


Ok, as we chat in IRC, a channel log per module. You can configure the
logs using the full path to a module (perceval.backends.gerrit) or a
subpatch of it (perceval). Working in this way, it is ok for me!

-- 
Alvaro del Castillo San Félix
acs@bitergia.com - Chief Technical Officer (CTO)
http://www.bitergia.com
"Software metrics for your peace of mind"

acs: Done!

PR #2, opened 2015-12-04
[github] Add github backend for perceval.

acs:


perceval/backends/github.py
```diff
@@ -0,0 +1,250 @@
...
+class GitHub(Backend):
+    """GitHub backend for Perceval
+
+    This class allows the fetch the issues stored in GitHub
+    repository.
+    """
+
+    def __init__(self, owner=None, repository=None, token=None,
+                 cache=None):
+        """ Init a GitHub instance
```
sduenas: Put the documentation on the top of the class instead of the `__init__` because the documentation looks prettier when it is generated with `sphynx`.

perceval/backends/github.py
```diff
@@ -0,0 +1,250 @@
...
+
+        self._push_cache_queue(issues)
+        self._flush_cache_queue()
+
+        while issues:
+            issue = issues.pop(0)
+            yield issue
+
+            if not issues:
+                issues = self.client.get_issues()
```
sduenas: Why don't you use here the date from the last date you retrieved?

perceval/backends/github.py
```diff
@@ -0,0 +1,250 @@
...
+                yield issue
+
+
+class GitHubClient:
+
+    def __init__(self, owner, repository, token):
+        self.owner = owner
+        self.repository = repository
+        self.auth_token = token
+        self.last_page = self.page = 1  # pagination in items downloading
```
sduenas: Don't get used to declare variables this way. Here there's no problem because `1` is an immutable object but it won't work the same when the objects are mutable. For instance:
```
>>> a = b = []
>>> a.append('x')
>>> print(a)
['x']
>>> print(b)
['x']
```

perceval/backends/github.py
```diff
@@ -0,0 +1,250 @@
...
+        cache_items = self.cache.retrieve()
+
+        for issues in cache_items:
+            for issue in issues:
+                yield issue
+
+
+class GitHubClient:
+
+    def __init__(self, owner, repository, token):
```
sduenas: This client won't work with a GitHub enterprise instance because no URL is provided.

perceval/backends/github.py
```diff
@@ -0,0 +1,250 @@
...
+        self.owner = owner
+        self.repository = repository
+        self.auth_token = token
+        self.last_page = self.page = 1  # pagination in items downloading
+
+    def _get_url(self):
+        github_api = "https://api.github.com"
```
sduenas: Same comment as above regarding GitHub enterprise.

perceval/backends/github.py
```diff
@@ -0,0 +1,250 @@
...
+        url_params += "&direction=asc"  # first older pull request
+        if startdate:
+            startdate = startdate.isoformat()
+            url_params += "&since=" + startdate
+
+        url = url_issues + url_params
+
+        return url
+
+    def get_issues(self, start=None):
```
sduenas: In my opinion, this should work like a generator. Thus, multiple calls can be launched at the same time using the same client.

perceval/backends/github.py
```diff
@@ -0,0 +1,250 @@
...
+        :param from_date: obtain issues updated since this date
+
+        :returns: a generator of issues
+        """
+
+        self._purge_cache_queue()
+
+        issues = self.client.get_issues(from_date)
+
+        self._push_cache_queue(issues)
```
sduenas: You're storing here the JSON object, not the raw stream.

sduenas:
@acs please address my comments

perceval/backends/github.py
acs: Because for navigation, we use url links and the from_date is needed just for getting the first navigation link. After that, we don't need it anymore (it is included in navigation links generated in GitHub API server side).

perceval/backends/github.py
acs: It looks prettier but it is not in the right place no? Imagine that __init__ is not at the top of the class definition. What do you think? If you prefer in any case to put the doc for __init__ on the top of the class, I will change it.

perceval/backends/github.py
acs: Done!

perceval/backends/github.py
acs: No, I have not worked with GitHub enterprise and I don't know the API. Not sure if this should be a requirement for the first version of this GitHub backend, or we can add the support later. What do you think? Our main concern right now is support GitHub in the cloud, so I will focus initially on that.

perceval/backends/github.py
acs: Same answer as above.

perceval/backends/github.py
acs: Not sure. In this case, I think I will probably recommend creating different clients for accessing different calls. But if you have a strong opinion about it, I will change it.

perceval/backends/github.py
acs: Yes, fixing it!

Done!

perceval/backends/github.py
sduenas: It's prettier and a common practice, too.

perceval/backends/github.py
sduenas: So, this means, the client has state and that if I want to fetch a new list of issues...I need a new instance of the client and of the backend. IMHO, it doesn't look friendly for developers.

perceval/backends/github.py
sduenas: The API is the same. The only thing that changes is the base URL of the server. For now, it is OK but it would be nice to take into account to develop the backend as much closer as possible to how it will work for both platforms.

perceval/backends/github.py
sduenas: I wrote it above but as I understand, the client can only used once, the same as the backend. If you want to use both class to fetch a new buch of issues, you need to create a new backend and a new client.

perceval/backends/github.py
acs: El mar, 22-12-2015 a las 17:06 -0800, Santiago Dueñas escribió:
> In perceval/backends/github.py:
> > +        :raises CacheError: raised when an error occurs accessing
> the
> > +            cache
> > +        """
> > +        if not self.cache:
> > +            raise CacheError(cause="cache instance was not
> provided")
> > +
> > +        cache_items = self.cache.retrieve()
> > +
> > +        for issues in cache_items:
> > +            for issue in issues:
> > +                yield issue
> > +
> > +
> > +class GitHubClient:
> > +
> > +    def __init__(self, owner, repository, token):
> The API is the same. The only thing that changes is the base URL of
> the server. For now, it is OK but it would be nice to take into
> account to develop the backend as much closer as possible to how it
> will work for both platforms.

Great. Being that the case, we could change it now, but I have no
environment to test it. I will add a new base URL param optional.

> —
> Reply to this email directly or view it on GitHub.

-- 
Alvaro del Castillo San Félix
acs@bitergia.com - Chief Technical Officer (CTO)
http://www.bitergia.com
"Software metrics for your peace of mind"

perceval/backends/github.py
acs: El mar, 22-12-2015 a las 17:04 -0800, Santiago Dueñas escribió:
> In perceval/backends/github.py:
> > +        :returns: a generator of issues
> > +        """
> > +
> > +        self._purge_cache_queue()
> > +
> > +        issues = self.client.get_issues(from_date)
> > +
> > +        self._push_cache_queue(issues)
> > +        self._flush_cache_queue()
> > +
> > +        while issues:
> > +            issue = issues.pop(0)
> > +            yield issue
> > +
> > +            if not issues:
> > +                issues = self.client.get_issues()
> So, this means, the client has state and that if I want to fetch a
> new list of issues...I need a new instance of the client and of the
> backend. IMHO, it doesn't look friendly for developers.

Probably the real issue here is that fetch is no recallable now from
the same instance of the backend. The option is to create a new client
inside of it, so fetch could be recallable. Maybe is that a better
solution?

> —
> Reply to this email directly or view it on GitHub.

-- 
Alvaro del Castillo San Félix
acs@bitergia.com - Chief Technical Officer (CTO)
http://www.bitergia.com
"Software metrics for your peace of mind"

perceval/backends/github.py
acs: El mar, 22-12-2015 a las 17:08 -0800, Santiago Dueñas escribió:
> In perceval/backends/github.py:
> > +
> > +        url_issues = self._get_url() + "/issues"
> > +
> > +        url_params = "?per_page=" + str(github_per_page)
> > +        url_params += "&state=all"  # open and close pull requests
> > +        url_params += "&sort=updated"  # sort by last updated
> > +        url_params += "&direction=asc"  # first older pull request
> > +        if startdate:
> > +            startdate = startdate.isoformat()
> > +            url_params += "&since=" + startdate
> > +
> > +        url = url_issues + url_params
> > +
> > +        return url
> > +
> > +    def get_issues(self, start=None):
> I wrote it above but as I understand, the client can only used once,
> the same as the backend. If you want to use both class to fetch a new
> buch of issues, you need to create a new backend and a new client.

Thinking about it, maybe the best approach is that in the fetch method
in the backend, which should be recallable using the same instance as
in other backends, we can create a specific client instance each time.

What do you think?

> —
> Reply to this email directly or view it on GitHub.

-- 
Alvaro del Castillo San Félix
acs@bitergia.com - Chief Technical Officer (CTO)
http://www.bitergia.com
"Software metrics for your peace of mind"

perceval/backends/github.py
acs: El mar, 22-12-2015 a las 17:02 -0800, Santiago Dueñas escribió:
> In perceval/backends/github.py:
> > +from ..cache import Cache
> > +from ..errors import CacheError
> > +from ..utils import DEFAULT_DATETIME, str_to_datetime
> > +
> > +logger = logging.getLogger(__name__)
> > +
> > +class GitHub(Backend):
> > +    """GitHub backend for Perceval
> > +
> > +    This class allows the fetch the issues stored in GitHub
> > +    repository.
> > +    """
> > +
> > +    def __init__(self, owner=None, repository=None, token=None,
> > +                 cache=None):
> > +        """ Init a GitHub instance
> It's prettier and a common practice, too.

Great! So then, I will change it as your proposed (here and in gerrit
backend).

> —
> Reply to this email directly or view it on GitHub.

-- 
Alvaro del Castillo San Félix
acs@bitergia.com - Chief Technical Officer (CTO)
http://www.bitergia.com
"Software metrics for your peace of mind"

acs: Done!

perceval/backends/github.py
acs: Done!

perceval/backends/github.py
sduenas: If you think on it, the current version of the client, in some how, is implemented like an iterator. Your `get_issues()` is like calling `next()` on an iterator. So, if this is the case, why don't you implement it like a Python iterator? Morever, a generator is like an iterator for dummies, so why don't you implement it like a generator?

perceval/backends/github.py
sduenas: Same comment as above.

perceval/backends/github.py
acs: Yes, probably it could implemented as an iterator and also, as a generator. We can talk later about it, in order to be sure we add value doing that and reworking the logic.

perceval/backends/github.py
acs: The client is very small. We can talk later about it, but I don't see value offering the possibility of calling several times get_issues over the same client vs creating different issues with their own state, which helps for example in tracking the first/last pages and urls, simplifying the logic of the user of the client.

acs:
Ok, get_issues in GitHubClient is now a generator. So it can be called several times (getting different generators) in the same instance of GitHubClient. So I think all comments are now addressed. What do you think @sduenas ?

PR #4, opened 2016-01-04
PR to integrate Perceval and Arthur

acs:


acs:
The idea of this PR is to continue pushing commits related to the integration of Perceval and Arthur and discuss them. Probably we need to reopen it!

PR #7, opened 2016-01-21
[github] Add to issues user data from the GitHub User API (email, company ...)

acs:
Add also the cache support for users.

The users in an issue are: "user", "assignee". Both of them have now extra user data retrieved from GitHub User API.

### DIFF
--- a/perceval/backends/github.py
+++ b/perceval/backends/github.py
@@ -80,7 +80,7 @@ class GitHub(Backend):
 
         if field not in fields:
             logging.warning("Wrong user field %s" % (field))
-            return []
+            return {}
 
         # Add user data: user and assignee
         if issue[field]:
--- a/perceval/backends/github.py
+++ b/perceval/backends/github.py
@@ -36,6 +36,7 @@ from ..utils import DEFAULT_DATETIME, str_to_datetime, urljoin
 
 
 GITHUB_URL = "https://github.com/"
+GITHUB_API_URL = "https://api.github.com"
 
 logger = logging.getLogger(__name__)
 
@@ -68,6 +69,22 @@ class GitHub(Backend):
         self.owner = owner
         self.repository = repository
         self.client = GitHubClient(owner, repository, token, base_url)
+        self._users = {}  # internal users cache
+
+    def _add_users(self, issue):
+        """ Add the users data to the item """
+        # Add user data: user and assignee
+        user_raw = self.client.get_user(issue['user']['login'])
+        self._push_cache_queue(user_raw)
+        issue['user_data'] = json.loads(user_raw)
+        if issue['assignee']:
+            user_raw = self.client.get_user(issue['assignee']['login'])
+            issue['assignee_data'] = json.loads(user_raw)
+            self._push_cache_queue(user_raw)
+        else:
+            issue['assignee_data'] = None
+
+        self._flush_cache_queue()
 
     @metadata(get_update_time)
     def fetch(self, from_date=DEFAULT_DATETIME):
@@ -90,7 +107,19 @@ class GitHub(Backend):
             self._flush_cache_queue()
             issues = json.loads(raw_issues)
             for issue in issues:
+                self._add_users(issue)
                 yield issue
+
+    def _add_users_from_cache(self, issue):
+        """ Add the users data to the cache item """
+        # All user data is already in self._users
+        login = issue['user']['login']
+        issue['user_data'] = self._users[login]
+        if issue['assignee']:
+            login = issue['assignee']['login']
+            issue['assignee_data'] = self._users[login]
+        else:
+            issue['assignee_data'] = None
 
     @metadata(get_update_time)
     def fetch_from_cache(self):
@@ -111,13 +140,29 @@ class GitHub(Backend):
 
         cache_items = self.cache.retrieve()
 
-        for raw_issues in cache_items:
-            issues = json.loads(raw_issues)
-            for issue in issues:
-                yield issue
-
+        issues = []
+        # In the cache first is the item with issues and next users data
+        # Read issues and all their users before processing issues
+        for raw_item in cache_items:
+            item = json.loads(raw_item)
+            if 'login' in item:
+                self._users[item['login']] = item
+            else:
+                next_issues = item
+                for issue in issues:
+                    # Add user data to the issue
+                    self._add_users_from_cache(issue)
+                    yield issue
+                issues = next_issues
+        for issue in issues:
+            # Add user data to the issue
+            self._add_users_from_cache(issue)
+            yield issue
 
 class GitHubClient:
+    """ Client for retieving information from GitHub API """
+
+    _users = {}  # users cache
 
     def __init__(self, owner, repository, token, base_url = None):
         self.owner = owner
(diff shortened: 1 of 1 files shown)


perceval/backends/github.py
```diff
@@ -187,6 +232,22 @@ def get_issues(self, start=None):
                 logger.debug("Rate limit: %s" %
                      (r.headers['X-RateLimit-Remaining']))
 
+    def get_user(self, login):
+        user = None
+
+        if login in self._users:
+            user = self._users[login]
```
sduenas: You can return the user here, avoiding an extra `else` (and indentation) block

perceval/backends/github.py
```diff
@@ -68,6 +69,22 @@ def __init__(self, owner=None, repository=None, token=None,
         self.owner = owner
         self.repository = repository
         self.client = GitHubClient(owner, repository, token, base_url)
+        self._users = {}  # internal users cache
+
+    def _add_users(self, issue):
```
sduenas: Do not modify the `issue` object inside the method. For this case, I prefer to fetch the user inside the method and return it. Outside, you can assign it to `assignee_key`.

perceval/backends/github.py
```diff
@@ -90,8 +107,20 @@ def fetch(self, from_date=DEFAULT_DATETIME):
             self._flush_cache_queue()
             issues = json.loads(raw_issues)
             for issue in issues:
+                self._add_users(issue)
                 yield issue
 
+    def _add_users_from_cache(self, issue):
```
sduenas: Same comment that as above. Do not modify the given object inside the method. It makes difficult to reuse it.

perceval/backends/github.py
```diff
@@ -111,13 +140,29 @@ def fetch_from_cache(self):
...
-            for issue in issues:
-                yield issue
-
+        issues = []
+        # In the cache first is the item with issues and next users data
+        # Read issues and all their users before processing issues
+        for raw_item in cache_items:
+            item = json.loads(raw_item)
+            if 'login' in item:
+                self._users[item['login']] = item
```
sduenas: You may want to add a `continue` here.

perceval/backends/github.py
```diff
@@ -111,13 +140,29 @@ def fetch_from_cache(self):
...
+                issues = next_issues
+        for issue in issues:
+            # Add user data to the issue
+            self._add_users_from_cache(issue)
+            yield issue
 
 class GitHubClient:
+    """ Client for retieving information from GitHub API """
+
+    _users = {}  # users cache
```
sduenas: Should it be a class attribute? As I understand about how the class works, it makes more sense to change it by a instance attribute.

sduenas:
Reviewed it. @acs I miss the code where you add the information about who made a comment or a change.

perceval/backends/github.py
acs: I don't like in general to return inside if/else for/while logic because it is hard to follow the logic, but in this simple case, ok, I will do that.

perceval/backends/github.py
acs: Yes, I don't like also to modify dicts inside external functions, but all the logic in this method is the assign logic for user and assignee. If we move it to the main function, we can just remove this function and _add_users_from_cache also, and have a bigger fetch functions. If you are more comfortable with this approach, I will do that.

perceval/backends/github.py
acs: Ok, the if/else logic is pretty different so I like this approach.

perceval/backends/github.py
acs: As we talked in the office, it make sense to share the cache between several instances of GitHubClient.

perceval/backends/github.py
```diff
@@ -68,6 +69,34 @@ def __init__(self, owner=None, repository=None, token=None,
...
+        """ Add the users data to the item """
+
+        user = {}
+
+        fields = ['user', 'assignee']
+
+        if field not in fields:
+            logging.warning("Wrong user field %s" % (field))
+            return {}
+
```
sduenas: You don't need to include all this code if you just pass to the method the `login` value. Neither `issue` nor `field` are needed here to get the user information.

For instance:
```python
def _get_user(self, login):
    if not login:
        return {}

    user_raw = self.client.get_user(login)
    user_orgs_raw = self.client.get_user_orgs(login)
    [...]

File: perceval/backends/github.py [hunk 7]
```diff
@@ -90,8 +119,16 @@ def fetch(self, from_date=DEFAULT_DATETIME):
             self._flush_cache_queue()
             issues = json.loads(raw_issues)
             for issue in issues:
+                for field in ['user', 'assignee']:
+                    issue[field+"_data"] = self._get_user(issue, field)
                 yield issue
 
+    def _get_user_from_cache(self, issue, field):
```
sduenas: Same comment as above.

perceval/backends/github.py
```diff
@@ -111,13 +148,42 @@ def fetch_from_cache(self):
 
         cache_items = self.cache.retrieve()
 
-        for raw_issues in cache_items:
-            issues = json.loads(raw_issues)
+        issues = []
+        # In the cache first is the item with issues and next users data
+        # Read issues and their users and their orgs before processing issues
+        for raw_item in cache_items:
```
sduenas: IMO, it makes more sense to have a `while` loop here. When you find a user item in the cache, you know the next item will be an organization item. Therefore, you can read a pair of items in sequence to build a `user`data dict. In fact, you can build each item in the same loop, instead of splitting the process into three loops.

perceval/backends/github.py
acs: Good point! Done!

perceval/backends/github.py
acs: Done!

perceval/backends/github.py
acs: Done!

acs:
@sduenas all review addressed except adding user data for people doing comments and changes. I am not sure we should complete user data also for this actors. The resulting JSON could be huge and with a lot of duplicates. What do you think?

perceval/backends/github.py
```diff
@@ -111,13 +137,41 @@ def fetch_from_cache(self):
 
         cache_items = self.cache.retrieve()
 
-        for raw_issues in cache_items:
+        # In the cache first is the item with issues and next users data
```
sduenas: I have another version here. It checks some possible failures that you haven't taken into account. I think it's more clear what it does. I didn't check it but I think it should work as is:

```python
issues = None

while True:
    try:
        raw_item = next(cache_items)
    except StopIteration:
        if issues:
            for issue in self._build_issues(issues):
                yield issue
        break

    item = json.loads(raw_item)

    if 'login' in item:
        try:
            raw_orgs = next(cache_items)
        except StopIteration:
            # Fatal error. Cache should had stored an organizations item
            # per parsed user.
            cause = "cache is exhausted but more items were expected"
            raise CacheError(cause=cause)

        item['organizations'] = json.loads(raw_orgs)
        self._users[item['login']] = item
        continue

    # A new set of issues has been read. It means we already
    # have the enough information to build and return the
    # previous set
    if issues:
        for issue in self._build_issues(issues):
            yield issue

    # Next issues to parse
    issues = item

def _build_issues(self, issues):
    for issue in issues:
        for field in ['user', 'assignee']:
            issue[field + '_data'] = {}

            if issue[field]:
                issue[field + '_data'] = \
                    self._users[issue[field]['login']]
        yield issue
```

sduenas:
@acs you're right. We can keep these data away for the moment.

Can you check the code that I added? Do you think it's better than your proposal?

PR #10, opened 2016-02-09
[stackexchange] Add backend implementation

albertinisg:
Added first backend implementation, updated perceval description message and README.md.

### DIFF
--- a/perceval/backends/stackexchange.py
+++ b/perceval/backends/stackexchange.py
@@ -47,7 +47,7 @@ class StackExchange(Backend):
 
     This class retrieves the questions stored in any of the
     StackExchange sites. To initialize this class the
-    site and the tag must be provided.
+    site must be provided.
 
     :param site: StackExchange site
     :param tagged: filter items by question Tag
@@ -56,7 +56,7 @@ class StackExchange(Backend):
     """
     version = '0.1.0'
 
-    def __init__(self, site=None, tagged=None, token=None,
+    def __init__(self, site, tagged=None, token=None,
                  max_questions=None, cache=None):
 
         super().__init__(site, cache=cache)
@@ -284,9 +284,9 @@ class StackExchangeCommand(BackendCommand):
 
         group.add_argument("--site", required=True,
                            help="StackExchange site")
-        group.add_argument("--tagged", required=True,
+        group.add_argument("--tagged",
                            help="filter items by question Tag")
-        group.add_argument("--token", required=True,
+        group.add_argument("--token",
                            help="StackExchange token for the API")
         group.add_argument('--max-questions', dest='max_questions',
                            type=int, default=MAX_QUESTIONS,
--- a/perceval/backends/stackexchange.py
+++ b/perceval/backends/stackexchange.py
@@ -32,16 +32,7 @@ from ..cache import Cache
 from ..errors import CacheError
 from ..utils import str_to_datetime, DEFAULT_DATETIME, urljoin
 
-# Filters are immutable and non-expiring. This filter allows to retrieve all
-# the information regarding Each question. To know more, visit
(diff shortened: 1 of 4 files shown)


perceval/backends/stackexchange.py
```diff
@@ -0,0 +1,296 @@
...
+from ..cache import Cache
+from ..errors import CacheError
+from ..utils import str_to_datetime, DEFAULT_DATETIME, urljoin
+
+# Filters are immutable and non-expiring. This filter allows to retrieve all
+# the information regarding Each question. To know more, visit
+# https://api.stackexchange.com/docs/questions and paste the filter in the
+# whitebox filter. It will display a list of checkboxes with the selected
+# values for the filter provided.
+
```
sduenas: As these constants are related to the API client, move them to that class.

perceval/backends/stackexchange.py
```diff
@@ -0,0 +1,296 @@
...
+    site and the tag must be provided.
+
+    :param site: StackExchange site
+    :param tagged: filter items by question Tag
+    :param token: StackExchange access_token for the API
+    :param cache: cache object to store raw data
+    """
+    version = '0.1.0'
+
+    def __init__(self, site=None, tagged=None, token=None,
```
sduenas: According to the doc from above, `site` and `tagged` are mandatory. Then, why do they have a default value?. I guess that `site` is mandatory, but not `tagged` because maybe I want to have all the items without regard to which tag they have.

perceval/backends/stackexchange.py
```diff
@@ -0,0 +1,296 @@
...
+        logger.info("Looking for questions at site '%s', with tag '%s' and updated from '%s'",
+                    self.site, self.tagged, str(from_date))
+
+        self._purge_cache_queue()
+
+        whole_page = self.client.get_questions(from_date)
+
+        for questions in whole_page:
+            for question in questions:
+                self._push_cache_queue(question)
```
sduenas: What it should be stored in the cache is `whole_page` not `question`. This is a misconception of how we implemented backend clients so far. See my comments below on class `StackExchangeClient`.

perceval/backends/stackexchange.py
```diff
@@ -0,0 +1,296 @@
...
+            fetched = req.json()['total']
+        else:
+            fetched = req.json()['page_size']
+
+        self.__log_status(req.json()['quota_remaining'],
+                          req.json()['quota_max'],
+                          fetched,
+                          req.json()['total'])
+
+        while questions:
```
sduenas: This whole loop has to be outside, in the backen class. This is not the duty of the client, it's the duty of the backend to know when it has to asks for more items. The rationale behind this it's we want to have clients that can be used to ask for anything (for instance, to ask for the third page of a certain set of questions without having to start from the first page). This make the code reusable by others and easier to maintain and to fix.

sduenas:
Good work @albertinisg. Please address the comments as soon as you can.

perceval/backends/stackexchange.py
sduenas: Ok. After all talk, you're right that it has more sense to return all the items related to a query because modern APIs allow us to do so in an easy way. The only thing here is you have to return the raw response with the `yield` statement:

```python
req = requests.get(self.__build_base_url(),
                   params=self.__build_payload(page, from_date))
questions = req.text

while questions:
    yield questions
    
    data = req.json()
    if data['has_more']:
        # do the next requests
       [...]
```

albertinisg:
Changes commited @sduenas. Let me know if I missed something!

sduenas:
Thanks for the PR, @albertinisg

albertinisg:
You are welcome! :smiley:

PR #13, opened 2016-02-19
Gerrit 2.8 support

acs:
Gerrit 2.8 does not support using the query (status:open OR status:closed). It is needed to get open and closed in different queries, returning the items ordered using update time. This is implemented in this PR.

### DIFF
--- a/perceval/backends/gerrit.py
+++ b/perceval/backends/gerrit.py
@@ -69,6 +69,86 @@ class Gerrit(Backend):
         self.client = GerritClient(self.url, user, max_reviews,
                                    blacklist_reviews)
 
+    def _fetch_gerrit28(self, from_date=DEFAULT_DATETIME):
+        """ Specific fetch for gerrit 2.8 version.
+
+        Get open and closed reviews in different queries.
+        Take the newer review from both lists and iterate.
+        """
+
+        # Convert date to Unix time
+        from_ut = datetime_to_utc(from_date)
+        from_ut = from_ut.timestamp()
+
+        filter_open = "status:open"
+        filter_closed = "status:closed"
+
+        last_item_open = self.client.next_retrieve_group_item()
+        last_item_closed = self.client.next_retrieve_group_item()
+        reviews_open = self._get_reviews(last_item_open, filter_open)
+        reviews_closed = self._get_reviews(last_item_closed, filter_closed)
+        last_nreviews_open = len(reviews_open)
+        last_nreviews_closed = len(reviews_closed)
+
+        while reviews_open or reviews_closed:
+            if reviews_open and reviews_closed:
+                if reviews_open[0]['lastUpdated'] >= reviews_closed[0]['lastUpdated']:
+                    review_open = reviews_open.pop(0)
+                    review = review_open
+                else:
+                    review_closed = reviews_closed.pop(0)
+                    review = review_closed
+            elif review_closed and not reviews_open:
+                review_closed = reviews_closed.pop(0)
+                review = review_closed
+            elif reviews_open and not reviews_closed:
+                review_open = reviews_open.pop(0)
+                review = review_open
+
+            updated = review['lastUpdated']
+            if updated <= from_ut:
+                logger.debug("No more updates for %s" % (self.url))
+                break
+            else:
+                yield review
+
+            if not reviews_open and last_nreviews_open >= self.max_reviews:
+                last_item_open = self.client.next_retrieve_group_item(last_item_open, review_open)
+                reviews_open = self._get_reviews(last_item_open, filter_open)
+                last_nreviews_open = len(reviews_open)
+            if not reviews_closed and last_nreviews_closed >= self.max_reviews:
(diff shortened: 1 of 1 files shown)


acs:
@sduenas new version updated. Could you review it?

perceval/backends/gerrit.py
```diff
@@ -258,24 +323,38 @@ def next_retrieve_group_item(self, last_item=None, entry=None):
                 next_item = 0
             else:
                 next_item = last_item
+        elif gerrit_version[0] == 2 and gerrit_version == 9:
+            # https://groups.google.com/forum/#!topic/repo-discuss/yQgRR5hlS3E
+            raise RuntimeExecption("Gerrit 2.9 does not support pagination")
```
sduenas: This exception does not exist and you misspelled it. Change it by the generic `BackendError`.

perceval/backends/gerrit.py
```diff
@@ -343,6 +422,8 @@ def run(self):
             total = 0
             for bug in bugs:
                 obj = json.dumps(bug, indent=4, sort_keys=True)
+                # bug_date = datetime.datetime.fromtimestamp(bug['lastUpdated']).isoformat()
```
sduenas: You should remove these comments.

perceval/backends/gerrit.py
```diff
@@ -258,24 +323,38 @@ def next_retrieve_group_item(self, last_item=None, entry=None):
...
             if entry is not None:
                 next_item = entry['sortKey']
 
         return next_item
 
-    def _get_gerrit_cmd(self, last_item):
+    def _get_gerrit_cmd(self, last_item, filter_=None):
+
+        if filter_ and filter_ not in ['status:open','status:closed']:
+            raise RuntimeError("Filter not supported in gerrit %s" % (filter_))
```
sduenas: This should raise a `BackendError` too!

perceval/backends/gerrit.py
```diff
@@ -258,24 +323,38 @@ def next_retrieve_group_item(self, last_item=None, entry=None):
...
 
         cmd = self.gerrit_cmd + " query "
         if self.project:
             cmd += "project:"+self.project+" "
         cmd += "limit:" + str(self.max_reviews)
 
-        if self.blacklist_reviews:
-            blacklist_reviews = ' AND NOT (%s)' % (','.join(self.blacklist_reviews))
-            cmd += " '(status:open OR status:closed)%s' " % (blacklist_reviews)
+        if not filter_:
```
sduenas: Is it possible to reword this? The code is more or less the same inside the `if..else` block.

perceval/backends/gerrit.py
acs: Fixed!

perceval/backends/gerrit.py
acs: Fixed! Arggg!

perceval/backends/gerrit.py
acs: I have already reword it trying to simplify it. Maybe we can improve it a little more but not sure. There are subtle differences between filter and not filter commands.

perceval/backends/gerrit.py
```diff
@@ -69,19 +69,59 @@ def __init__(self, url,
...
+                if reviews_open[0]['lastUpdated'] >= reviews_closed[0]['lastUpdated']:
+                    review_open = reviews_open.pop(0)
+                    review = review_open
+                else:
+                    review_closed = reviews_closed.pop(0)
+                    review = review_closed
+            elif review_closed and not reviews_open:
+                review_closed = reviews_closed.pop(0)
+                review = review_closed
+            elif reviews_open and not reviews_closed:
```
sduenas: This check should be an `else` statement. If the code reaches here that will mean that  `reviews_closed` is empty.

perceval/backends/gerrit.py
```diff
@@ -69,19 +69,59 @@ def __init__(self, url,
...
+
+        while reviews_open or reviews_closed:
+            if reviews_open and reviews_closed:
+                if reviews_open[0]['lastUpdated'] >= reviews_closed[0]['lastUpdated']:
+                    review_open = reviews_open.pop(0)
+                    review = review_open
+                else:
+                    review_closed = reviews_closed.pop(0)
+                    review = review_closed
+            elif review_closed and not reviews_open:
```
sduenas: There's a bug here. It should be `reviews_closed` instead of `review_closed`. You can also remove the `and not` condition.

perceval/backends/gerrit.py
acs: Yes, but maybe it is more clear to put the elif explicitly.

perceval/backends/gerrit.py
acs: Fixed!

perceval/backends/gerrit.py
sduenas: Ok. We can keep it as is for now.

sduenas:
Please remove the `import datetime` line from the beginning of the code. It is not required anymore.

PR #15, opened 2016-02-24
[github] backend minor refactorization

albertinisg:


sduenas:
@albertinisg Most of the PR is OK but there's only one comment that you should address before merging it. Would you mind to make a new commit fixing it?

sduenas:
@albertinisg I alson think you have to update the section `Examples` on `README.md` file because the parameter `token` is not used any more when `github` backend is executed.

albertinisg:
@sduenas changes commited!

PR #23, opened 2016-03-18
[git] Fix carriage return not in end of line

### LINKED ISSUE #21: ^M) in git logs causing trouble
In some messages from the Linux kernel git log, the following line appears:

```
    ^M)
```

That is, four spaces followed by ASCII 13, Carriage Return, a parenthesis and the end of line. See an example of a one-commit git log with one of these commits in
[linux-gitlog.txt](https://github.com/grimoirelab/perceval/files/178808/linux-gitlog.txt) (line 9 in that file).

The regex in perceval/backends/git.py for parsing the message is looking for lines starting with four spaces:

``` python
    MESSAGE_LINE_PATTERN = r"[\s]{4}(?P<msg>.*)$"
```

But this fails miserably in this case: the CR is identified by the regex as the end of the line ($), and the reminding parenthesis as the beginning of the next line. Since that one does not start with four spaces (in fact, it starts with "(", it is not considered as a message line, and therefore the message ends. This causes the code to consider the commit up to this line, and then expect a new commit starting in line 10, which is not happening. Exception launched, end. You can reproduce (using the attached file, saved as /tmp/linux-gitlog.txt) as follows:

```
$ perceval -g git --git-log /tmp/linux-gitlog.txt http://github.com/torvalds/linux.git
[2016-03-17 23:51:36,617 - root - INFO] - Sir Perceval is on his quest.
[2016-03-17 23:51:36,618 - perceval.backends.git - INFO] - Fetching commits: 'http://github.com/torvalds/linux.git' git repository from 1970-01-01 00:00:00+00:00
[2016-03-17 23:51:36,618 - perceval.backends.git - DEBUG] - Invalid message format on line 10. Skipping.
[2016-03-17 23:51:36,619 - perceval.backends.git - DEBUG] - Invalid action format on line 10. Skipping.
[2016-03-17 23:51:36,619 - perceval.backends.git - DEBUG] - Commit 2d137c24e9f433e37ffd10b3d5f418157589a8d2 parsed
{
    "Author": "akpm@osdl.org <akpm@osdl.org>",
    "AuthorDate": "Sat Apr 16 15:23:55 2005 -0700",
    "Commit": "Linus Torvalds <torvalds@ppc970.osdl.org>",
    "CommitDate": "Sat Apr 16 15:23:55 2005 -0700",
    "__metadata__": {
        "backend_name": "Git",
        "backend_version": "0.1.0",
        "origin": "http://github.com/torvalds/linux.git",
        "timestamp": 1458255096.619283,
        "updated_on": 1113690235.0
    },
    "commit": "2d137c24e9f433e37ffd10b3d5f418157589a8d2",
    "files": [],
    "message": "[PATCH] arm: fix SIGBUS handling\n\n",
    "parents": [
        "baaa2c512dc1c47e3afeb9d558c5323c9240bd21"
    ],
    "refs": []
}
  File "/home/jgb/venvs/grimoirelab/lib/python3.5/site-packages/perceval/backends/git.py", line 204, in run
    for commit in commits:
  File "/home/jgb/venvs/grimoirelab/lib/python3.5/site-packages/perceval/backend.py", line 161, in decorator
    for item in func(self, *args, **kwargs):
  File "/home/jgb/venvs/grimoirelab/lib/python3.5/site-packages/perceval/backends/git.py", line 107, in fetch
    for commit in commits:
  File "/home/jgb/venvs/grimoirelab/lib/python3.5/site-packages/perceval/backends/git.py", line 150, in parse_git_log_from_file
    for commit in parser.parse():
  File "/home/jgb/venvs/grimoirelab/lib/python3.5/site-packages/perceval/backends/git.py", line 371, in parse
    parsed = self.handlers[self.state](line)
  File "/home/jgb/venvs/grimoirelab/lib/python3.5/site-packages/perceval/backends/git.py", line 413, in _handle_commit
    raise ParseError(cause=msg)
perceval.errors.ParseError: commit expected on line 10

  File "/home/jgb/venvs/grimoirelab/bin/perceval", line 163, in <module>
    main()
  File "/home/jgb/venvs/grimoirelab/bin/perceval", line 84, in main
    cmd.run()
  File "/home/jgb/venvs/grimoirelab/lib/python3.5/site-packages/perceval/backends/git.py", line 211, in run
    raise RuntimeError(str(e))
RuntimeError: commit expected on line 10
```

I guess the regex should be fixed to consider this case, or maybe CR should be filtered out somehow before using the regex.

jgbarah:
In some commit messages, there is a carriage return character
not in the end of a line. This caused a wrong line break which
caused the parser to misidentify the end of the message.

Fixes #21

### DIFF
--- a/perceval/backends/git.py
+++ b/perceval/backends/git.py
@@ -154,7 +154,8 @@ class Git(Backend):
         :raises OSError: raised when an error occurs reading the
             given file
         """
-        with open(filepath, 'r', errors='surrogateescape') as f:
+        with open(filepath, 'r', errors='surrogateescape',
+                  newline=os.linesep) as f:
             parser = GitParser(f)
 
             for commit in parser.parse():
--- a/tests/test_git.py
+++ b/tests/test_git.py
@@ -204,6 +204,21 @@ class TestGitBackend(unittest.TestCase):
         self.assertEqual(commit['commit'], 'cb24e4f2f7b2a7f3450bfb15d1cbaa97371e93fb')
         self.assertEqual(commit['message'], 'Calling \udc93Open Type\udc94 (CTRL+SHIFT+T) after startup - performance improvement.')
 
+    def test_git_cr_error(self):
+        """Test if mislocated carriage return chars do not break lines
+
+        In some commit messages, carriage return characters (\r) are found
+        in weird places. They should not be misconsidered as end of line.
+
+        Before fixing, this test raises an exception:
+        "perceval.errors.ParseError: commit expected on line 10"
+
+        """
+
+        commits = Git.parse_git_log_from_file("data/git_bad_cr.txt")
+        result = [commit for commit in commits]
+        self.assertEqual(len(result), 1)
+
     def test_git_parser_from_iter(self):
         """Test if the static method parses a git log from a repository"""
 
--- a/tests/data/git_bad_cr.txt
+++ b/tests/data/git_bad_cr.txt
@@ -1,0 +1,13 @@
+commit 2d137c24e9f433e37ffd10b3d5f418157589a8d2 baaa2c512dc1c47e3afeb9d558c5323c9240bd21
+Author:     akpm@osdl.org <akpm@osdl.org>
+AuthorDate: Sat Apr 16 15:23:55 2005 -0700
+Commit:     Linus Torvalds <torvalds@ppc970.osdl.org>
+CommitDate: Sat Apr 16 15:23:55 2005 -0700
+
+    [PATCH] arm: fix SIGBUS handling
+    
+    )
+    
+:100644 100644 29be1c0... e25b4fd... M	arch/arm/mm/fault.c
+36	44	arch/arm/mm/fault.c
+


perceval/backends/git.py
```diff
@@ -154,7 +154,8 @@ def parse_git_log_from_file(filepath):
         :raises OSError: raised when an error occurs reading the
             given file
         """
-        with open(filepath, 'r', errors='surrogateescape') as f:
+        with open(filepath, 'r', errors='surrogateescape',
+                  newline = os.linesep) as f:
```
sduenas: Please follow PEP-8 rules about the use of whitespaces on keyword arguments.
```
Don't use spaces around the = sign when used to indicate a keyword argument
or a default parameter value. 
```
Change `newline = os.linesep` by `newline=os.linesep`

tests/test_git.py
```diff
@@ -204,6 +204,25 @@ def test_git_encoding_error(self):
...
+        In some commit messages, carriage return characters (\r) are found
+        in weird places. They should not be misconsidered as end of line.
+
+        Before fixing, this test raises an exception:
+        "perceval.errors.ParseError: commit expected on line 10"
+
+        """
+
+        commits = Git.parse_git_log_from_file("data/git_bad_cr.txt")
+        try:
```
sduenas: Why do you need this `try..except` block here? If the error was fixed it shouldn't raise any exception.

tests/test_git.py
```diff
@@ -204,6 +204,25 @@ def test_git_encoding_error(self):
...
+            result = [commit for commit in commits]
+        except e:
+            self.fail("Exception raised: " + e)
+        self.assertEqual(len(result), 1)
+        commit = result[0]
```
sduenas: You should remove this line because the code does not do anything with the commit.

perceval/backends/git.py
jgbarah: OK, doing it.

tests/test_git.py
jgbarah: Just to make the error explicit. But I'm removing it anyway, since you seem to prefer that ;-)

tests/test_git.py
jgbarah: You're right. My error.

PR #24, opened 2016-03-19
[git] Make reading of git log output a real iterator.

jgbarah:
Instead of reading all stdout from git log into a list of strings,
read each line and yield. Use a thread for ensuring it doesn't
block because the stdout fills the buffer.

### DIFF
--- a/perceval/backends/git.py
+++ b/perceval/backends/git.py
@@ -25,6 +25,7 @@ import logging
 import os
 import re
 import subprocess
+import threading
 
 from ..backend import Backend, BackendCommand, metadata
 from ..errors import RepositoryError, ParseError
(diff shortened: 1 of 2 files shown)


jgbarah:
I had to touch some tests, because now an empty output from "git log" returns no line. In general, if a log ends with an empty line, it is not returned. This seems to be consistent with what happens when instead of getting the output from "git log", a file is used to read the log (before this patch, the behavior seemed to be different).

perceval/backends/git.py
```diff
@@ -632,25 +660,46 @@ def log(self, from_date=None, encoding='utf-8'):
...
         if from_date:
             dt = from_date.strftime("%Y-%m-%d %H:%M:%S %z")
             cmd_log.append('--since=' + dt)
+        env = {'LANG' : 'C', 'PAGER' : ''}
 
-        gitlog = self._exec(cmd_log, cwd=self.dirpath,
-                            env={'LANG' : 'C', 'PAGER' : ''})
-        gitlog = gitlog.split(b'\n')
-
+        self.failed_message = None
```
sduenas: I think it makes more sense to move all this code to `_exec` method because it's generic and can be useful for other (future) commands.

perceval/backends/git.py
```diff
@@ -632,25 +660,46 @@ def log(self, from_date=None, encoding='utf-8'):
...
+        try:
+            self.proc = subprocess.Popen(cmd_log,
+                                        stdout=subprocess.PIPE,
+                                        stderr=subprocess.PIPE,
+                                        cwd=self.dirpath,
+                                        env=env)
+        except OSError as e:
+            raise RepositoryError(cause=str(e))
+        # Spawn the thread for reading stderr
+        err_thread = threading.Thread (target = self._read_stderr, daemon = True)
```
sduenas: Don't use white spaces between the class and the `(`. Same comment for keyword arguments.

perceval/backends/git.py
```diff
@@ -632,25 +660,46 @@ def log(self, from_date=None, encoding='utf-8'):
...
+                      ' '.join(cmd_log), self.dirpath, str(env))
+        try:
+            self.proc = subprocess.Popen(cmd_log,
+                                        stdout=subprocess.PIPE,
+                                        stderr=subprocess.PIPE,
+                                        cwd=self.dirpath,
+                                        env=env)
+        except OSError as e:
+            raise RepositoryError(cause=str(e))
+        # Spawn the thread for reading stderr
```
sduenas: From my point of view, I don't need these comments, they are redundant. The code is telling what it is doing. There isn't anything difficult to understand. I think you can remove them.

perceval/backends/git.py
sduenas: As a general rule, try to add more blank lines to your code. Now it looks like a block. There are some parts that you can split like paragraphs and will make the code easier to read :-D. For instance, I usually add a blank lines between statements and `if` or `for` blocks. For instance, I think this is better:

```python

err_thread.start()

for line in self.proc.stdout:
    yield line.decode('utf-8', errors='surrogateescape')

err_thread.join()

self.proc.communicate()
self.proc.stdout.close()
self.proc.stderr.close()

if self.proc.returncode != 0:
    cause = "git command - %s (return code: %d)" % \
        (self.failed_message, self.proc.returncode)
    raise RepositoryError(cause=cause)
    
logging.debug("Git log fetched from %s repository (%s)",
              self.uri, self.dirpath)
```

than this:

```python
err_thread.start()
# Read stdout, yielding to the consumer of this iterator
for line in self.proc.stdout:
    yield line.decode('utf-8', errors='surrogateescape')
# No more stdout, wait for the thread reading stderr, close descriptors
err_thread.join()
# Wait until subprocess is done
self.proc.communicate()
self.proc.stdout.close()
self.proc.stderr.close()
if self.proc.returncode != 0:
    # Subprocess failed, error in failed_message, from stdout
    cause = "git command - %s (return code: %d)" % \
        (self.failed_message, self.proc.returncode)
    raise RepositoryError(cause=cause)
logging.debug("Git log fetched from %s repository (%s)",
              self.uri, self.dirpath)
```

tests/test_git.py
```diff
@@ -586,7 +584,7 @@ def test_log_empty(self):
         gitlog = repo.log(from_date=datetime.datetime(2020, 1, 1, 1, 1, 1))
         gitlog = [line for line in gitlog]
 
-        self.assertListEqual(gitlog, [''])
+        self.assertListEqual(gitlog, [])
```
sduenas: I like this. We don't win anything returning an empty line.

perceval/backends/git.py
```diff
@@ -611,6 +612,33 @@ def pull(self):
         logging.debug("Git %s repository pulled into %s",
                       self.uri, self.dirpath)
 
+    def _read_stderr (self):
```
sduenas: As a general rule, I use to put _protected_ methods after _public_ methods and sorted by when they are called.

perceval/backends/git.py
```diff
@@ -632,25 +660,46 @@ def log(self, from_date=None, encoding='utf-8'):
...
+        self.failed_message = None
+        logging.debug("Running command %s (cwd: %s, env: %s)",
+                      ' '.join(cmd_log), self.dirpath, str(env))
+        try:
+            self.proc = subprocess.Popen(cmd_log,
+                                        stdout=subprocess.PIPE,
+                                        stderr=subprocess.PIPE,
+                                        cwd=self.dirpath,
+                                        env=env)
+        except OSError as e:
```
sduenas: I think this `try..except` should enclose all the actions related to self.proc. When you are calling to `for line in self.proc.stdout`, the process can raise the same exception. You can catch the exception and terminate `err_thread` if it is still alive - I don't know if it could be possible -.

A unit test to check whether the error thread is dead would be very useful too.

sduenas:
@jgbarah thanks for the PR. Please address my comments.

jgbarah:
New version. I guess i addressed all of your comments, except for the one on the new test for detecting the end of the spawned thread. I suggest that we open a ticket for it, and I will deal with it asap. If you prefer, I can open the ticket. If you prefer, we still wait until I write the test.

jgbarah:
New version. I added a not directly related test on whether a directory intended to be used as a git repo is in fact a git repo.

PR #29, opened 2016-03-28
Update docker image name

camillem:
Updating docker image name from bitergia/perceval to grimoirelab/perceval and tag name from master to latest


sduenas:
Thanks for the patch, @camillem

PR #31, opened 2016-03-29
[Gerrit] Add --blacklist-reviews option

acs:
Add --blacklist-reviews option for not retrieving wrong reviews from the remote gerrit server.

In Eclipse gerrit for example, right now:

https://git.eclipse.org/r/67778

returns and errors, and using the ssh API the server crash with a: 

fatal: internal server error

With this option, it can be avoided to download this reviews so the rest is retrieved correctly:

```
perceval -g gerrit --user XXXX --url git.eclipse.org --from-date '2015-12-29 17:00' --blacklist-reviews '67778'
```

### DIFF
--- a/perceval/backends/gerrit.py
+++ b/perceval/backends/gerrit.py
@@ -265,7 +265,7 @@ class GerritClient():
         cmd += "limit:" + str(self.max_reviews)
 
         if self.blacklist_reviews:
-            blacklist_reviews = ' AND NOT (%s)' % (self.blacklist_reviews)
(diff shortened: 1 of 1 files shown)


sduenas:
I like the idea but giving these values as a comma separated list is error prone.

My suggestion is to pass the reviews ids to `GerritCommand` as a list of arguments (https://docs.python.org/3/library/argparse.html#nargs). These values will be stored into a list that you will have to convert to a string using `','.join(blacklist)`.

acs:
Changed according to your suggestion @sduenas

PR #34, opened 2016-04-23
Added test_github.py

zhquan:


tests/test_github.py
```diff
@@ -0,0 +1,746 @@
...
+    @httpretty.activate
+    def test_get_user(self):
+        """ Test get_user API call """
+
+        login = read_file('data/github_login')
+
+        httpretty.register_uri(httpretty.GET,
+                               GITHUB_LOGIN_URL,
+                               body=login, status=200)
+        httpretty.register_uri(httpretty.GET,
```
sduenas: You can remove this because you're only testing the result of calling to `get_user()` method. You can also remove the calls to `get_issues()` from below.

tests/test_github.py
```diff
@@ -0,0 +1,746 @@
...
+                               })
+
+        client = GitHubClient("zhquan_example", "repo", "aaa", None)
+        response = client.get_user("zhquan_example")
+        self.assertEqual(response, login)
+
+        _ = [issues for issues in client.get_issues()]
+
+        # Check requests
+        expected = {
```
sduenas: Instead of checking the parameters generated by `get_issues()` you have to check those from `get_user()`

tests/test_github.py
```diff
@@ -0,0 +1,746 @@
...
+    @httpretty.activate
+    def test_get_user_orgs(self):
+        """ Test get_user_orgs API call """
+
+        orgs = read_file('data/github_orgs')
+
+        httpretty.register_uri(httpretty.GET,
+                               GITHUB_ORGS_URL,
+                               body=orgs, status=200)
+        httpretty.register_uri(httpretty.GET,
```
sduenas: Remove this. Same comment that above.

tests/test_github.py
```diff
@@ -0,0 +1,746 @@
...
+        args = ['--owner', 'zhquan_example',
+                '--repository', 'repo',
+                '--from-date', '2016-01-03',
+                '-t', 'aaa']
+
+        cmd = GitHubCommand(*args)
+        self.assertIsInstance(cmd.parsed_args, argparse.Namespace)
+        self.assertEqual(cmd.parsed_args.owner, "zhquan_example")
+        self.assertEqual(cmd.parsed_args.repository, "repo")
+        self.assertEqual(cmd.parsed_args.backend_token, "aaa")
```
sduenas: You missed to check the value of `from_date` parameter.

tests/test_github.py
```diff
@@ -0,0 +1,746 @@
...
+    sys.path.insert(0, '..')
+
+from perceval.cache import Cache
+from perceval.errors import BackendError, CacheError, ParseError
+from perceval.backends.github import GitHub, GitHubCommand, GitHubClient
+
+
+GITHUB_API_URL = "https://api.github.com"
+GITHUB_ISSUES_URL = GITHUB_API_URL + "/repos/zhquan_example/repo/issues"
+GITHUB_LOGIN_URL = GITHUB_API_URL + "/users/zhquan_example"
```
sduenas: Change the name of this constant to `GITHUB_USER_URL`. When I read it I think that it is an URL to sign in.

tests/test_github.py
```diff
@@ -0,0 +1,746 @@
...
+    @httpretty.activate
+    def test_fetch(self):
+        """ Test whether a list of issues is returned """
+
+        command = ""
+        body = read_file('data/github_request')
+        login = read_file('data/github_login')
+        orgs = read_file('data/github_orgs')
+
+        httpretty.register_uri(httpretty.GET,
```
sduenas: What is this? I think the client does not make any request to this URL.

tests/test_github.py
```diff
@@ -0,0 +1,746 @@
...
+                "id": 1
+            },
+            "perceval_version": "0.1.0",
+            "updated_on": 1454328801,
+            "origin": "https://github.com/zhquan_example/repo",
+            "uuid": "58c073fd2a388c44043b9cc197c73c5c540270ac",
+            "backend_version": "0.1.0"
+        }
+
+        for key in expected:
```
sduenas: Instead of using this `for` loop to check the values of the issue, you can use the method self.assertDictEqual to compare the two dictionaries.

tests/test_github.py
```diff
@@ -0,0 +1,746 @@
...
+                "id": 2
+            },
+            "perceval_version": "0.1.0",
+            "updated_on": 1458054569.0,
+            "origin": "https://github.com/zhquan_example/repo",
+            "uuid": "4236619ac2073491640f1698b5c4e169895aaf69",
+            "backend_version": "0.1.0"
+        }
+
+        for key in expected:
```
sduenas: Same comment that above but you also have to check the contents of the two issues, not just only one.

tests/data/github_request
```diff
@@ -0,0 +1 @@
...wers_url":"https://api.github.com/users/zhquan_example/followers","following_url":"https://api.github.com/users/zhquan_example/following{/other_user}","gists_url":"https://api.github.com/users/zhquan_example/gists{/gist_id}","starred_url":"https://api.github.com/users/zhquan_example/starred{/owner}{/repo}","subscriptions_url":"https://api.github.com/users/zhquan_example/subscriptions","organizations_url":"https://api.github.com/users/zhquan_example/orgs","repos_url":"https://api.github.com/users/zhquan_example/repos","events_url":"https://api.github.com/users/zhquan_example/events{/privacy}","received_events_url":"https://api.github.com/users/zhquan_example/received_events","type":"User","site_admin":false},"labels":[],"state":"closed","locked":false,"assignee":null,"milestone":null,"comments":4,"created_at":"2016-02-01T07:10:24Z","updated_at":"2016-02-01T12:13:21Z","closed_at":"2016-02-01T12:13:21Z","pull_request":{"url":"https://api.github.com/repos/zhquan_example/repo/pulls/1","html_url":"https://github.com/zhquan_example/repo/pull/1","diff_url":"https://github.com/zhquan_example/repo/pull/1.diff","patch_url":"https://github.com/zhquan_example/repo/pull/1.patch"},"body":"Body"}]
```
sduenas: Please change the name to this file. Call it github_issues or similar. You should also have to add more issues to this JSON, to check that the backend works parsing more than an issue from the same JSON stream.

sduenas:
As a general comment, please format the JSON files. It is impossible to read/edit anything there. You can format the files using python. For instance:

```
$ cat github_issue_2 | python3 -m json.tool
```

sduenas:
@zhquan, I've been reviewing your last commit and it is almost OK. The problem is I was expecting to have the check for `sleep_for_rate` in a separated test. Would you mind to create a specific test for this case? There is also a comment regarding your check, please fixed it.

And you also have to remove `ParseError` from line 40. We're not using it.

tests/test_github.py
```diff
@@ -370,27 +370,35 @@ def test_get_page_issues(self):
...
-                                    'X-RateLimit-Remaining': '4985',
-                                    'X-RateLimit-Reset': '14611'
+                                    'X-RateLimit-Remaining': '20',
+                                    'X-RateLimit-Reset': '15'
                                })
 
-        client = GitHubClient("zhquan_example", "repo", "aaa", None)
+        client = GitHubClient("zhquan_example", "repo", "aaa", sleep_for_rate=True)
+
+        before = datetime.datetime.utcnow()
```
sduenas: You should use time.time(). It is faster.

sduenas:
Merged. Good work @zhquan.

PR #36, opened 2016-04-25
Support sleep mode for waiting until the rate limit is reset.

acs:
Add a new param (sleep-for-rate) to the GitHub backend so if it is defined, then when the rate limit is exhausted, wait until the rate limit is reset and the backend can continue with the issues retrieval.

If the param (sleep-for-rate) is not defined, the throw a RateLimitException which includes the seconds to wait until the rate limit is reset.

### DIFF
--- a/perceval/backends/github.py
+++ b/perceval/backends/github.py
@@ -22,11 +22,12 @@
 #     Alberto Martín <alberto.martin@bitergia.com>
 #
 
+from datetime import datetime
 import json
 import logging
 import os.path
-
 import requests
+from time import sleep
 
 from ..backend import Backend, BackendCommand, metadata
 from ..cache import Cache
@@ -225,6 +226,8 @@ class GitHubClient:
         self.repository = repository
         self.token = token
         self.base_url = base_url
+        self.rate_limit = None
+        self.rate_limit_reset = None
 
     def __get_url(self):
         github_api = GITHUB_API_URL
@@ -254,6 +257,19 @@ class GitHubClient:
             headers = {'Authorization': 'token ' + self.token}
             return headers
 
+    def __get_github(self, url, params=None, headers=None):
+        """ GET HTTP caring of rate limit """
+        if self.rate_limit and self.rate_limit <= 1:
+            reset_seconds = (self.rate_limit_reset-datetime.utcnow()).seconds+1
+            logging.info("Rate limit consumed. Waiting %i secs for rate limit reset." % (reset_seconds))
+            sleep(reset_seconds)
+        r = requests.get(url, params=params, headers=headers)
+        r.raise_for_status()
+        self.rate_limit = r.headers['X-RateLimit-Remaining']
+        self.rate_limit_reset = datetime.fromtimestamp(int(r.headers['X-RateLimit-Reset']))
+        logger.debug("Rate limit: %s" % (self.rate_limit))
+        return r
+
     def get_issues(self, start=None):
         """ Return the items from github API using links pagination """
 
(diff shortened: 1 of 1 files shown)


perceval/backends/github.py
```diff
@@ -22,15 +22,16 @@
 #     Alberto Martín <alberto.martin@bitergia.com>
 #
 
+from datetime import datetime
```
sduenas: Please move this line under the last import of packages from the standard library. You should also move the `from time import sleep` line there too.

perceval/backends/github.py
```diff
@@ -39,11 +40,24 @@
...
+class RateLimitError(BaseError):
+
+    message = "%(cause)s %(reset_seconds)s seconds for rate reset"
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self._reset_seconds = kwargs['reset_seconds']
+
+    @property
+    def reset_seconds(self):
```
sduenas: Rename this to `seconds_to_reset`. This is a property or an attribute. With the `reset_` prefix it looks like more that calling this will reset something to seconds. You should also rename the attribute in all the occurrences on this class.

perceval/backends/github.py
```diff
@@ -39,11 +40,24 @@
 
 GITHUB_URL = "https://github.com/"
 GITHUB_API_URL = "https://api.github.com"
-
+RATE_LIMIT_MIN = 10  # Min rate limit before sleeping until rate limit reset
```
sduenas: I think `MIN` should be a prefix not a suffix, as we do with `MAX_ISSUES` or similar constants.

perceval/backends/github.py
```diff
@@ -39,11 +40,24 @@
...
 
 logger = logging.getLogger(__name__)
 
 
+class RateLimitError(BaseError):
+
+    message = "%(cause)s %(reset_seconds)s seconds for rate reset"
```
sduenas: The text should be `seconds to rate` ...

perceval/backends/github.py
```diff
@@ -254,6 +276,25 @@ def __get_headers(self):
             headers = {'Authorization': 'token ' + self.token}
             return headers
 
+    def __get_github(self, url, params=None, headers=None):
+        """ GET HTTP caring of rate limit """
+        if self.rate_limit is not None and self.rate_limit <= RATE_LIMIT_MIN:
+            reset_seconds = (self.rate_limit_reset-datetime.utcnow()).seconds+1
```
sduenas: Use whitespaces between operators.

perceval/backends/github.py
```diff
@@ -254,6 +276,25 @@ def __get_headers(self):
...
+            cause = "GitHub rate limit exhausted."
+            if self.sleep_for_rate:
+                logging.info("%s Waiting %i secs for rate limit reset." % (cause, reset_seconds))
+                sleep(reset_seconds)
+            else:
+                ex = RateLimitError(cause=cause, reset_seconds=reset_seconds)
```
sduenas: I think you can raise the exception without creating a variable before.

perceval/backends/github.py
acs: ok

perceval/backends/github.py
acs: ok

perceval/backends/github.py
acs: ok

perceval/backends/github.py
acs: Yes, I have the "ex" var for testing the attribute ... but it is not needed anymore.

perceval/backends/github.py
acs: ok

perceval/backends/github.py
acs: ok

PR #38, opened 2016-05-24
Jenkins backend for Perceval

acs:
Jenkins backend which retrieves the builds from Jenkins site.

The backend includes also unit tests.


perceval/backends/jenkins.py
```diff
@@ -0,0 +1,243 @@
...
+import requests
+
+from ..backend import Backend, BackendCommand, metadata
+from ..cache import Cache
+from ..errors import CacheError
+from ..utils import (DEFAULT_DATETIME,
+                     datetime_to_utc,
+                     str_to_datetime,
+                     urljoin)
+
```
sduenas: Two white lines should be in between.

perceval/backends/jenkins.py
```diff
@@ -0,0 +1,243 @@
...
+logger = logging.getLogger(__name__)
+
```
sduenas: Same comment as above here.

perceval/backends/jenkins.py
```diff
@@ -0,0 +1,243 @@
...
+            raw_builds = self.client.get_builds(job['url'])
+            self._push_cache_queue(raw_builds)
+            self._flush_cache_queue()
+            builds = json.loads(raw_builds)
+            builds = builds['builds']
+            for build in builds:
+                yield build
+                nbuilds += 1
+
+        logging.info("Total number of jobs: %i" % (njobs))
```
sduenas: Instead of logging, use the object `logger`.

perceval/backends/jenkins.py
```diff
@@ -0,0 +1,243 @@
...
+
+    :param site: URL of jenkins node: https://build.opnfv.org/ci
+
+    :raises HTTPError: when an error occurs doing the request
+    """
+
+    def __init__(self, site):
+        self.site = site
+
+
```
sduenas: Remove this blank line.

perceval/backends/jenkins.py
```diff
@@ -0,0 +1,243 @@
...
+    :param site: Jenkins site
+    :param job_url: Jenkins job from which to get the builds
+    :param cache: cache object to store raw data
+    :param origin: identifier of the repository; when `None` or an
+        empty string are given, it will be set to `site` value
+    """
+    version = '0.1.0'
+
+    def __init__(self, site, job_url=None, cache=None, origin=None):
```
sduenas: Do we need the URL of a job? I mean, if we already have the `site` that is the URL of the Jenkins site I think it does not make sense to give it again just to fetch the data from a job. In my opinion, the job parameter should be only the name of the job. Then, the client will build the URL with that information.

sduenas: I have another comment here regarding `job_url`. I understand what you are trying to do but I think it breaks the common schema that we have on the rest of backends. We don't implement the possibility of fetching data from an item and if we allow it, the functionality should be implemented to retrieve data from a list of items, not for just one item.

sduenas: All of our backends retrieve data using the parameter `--from-date`. Can be it applied here?

perceval/backends/jenkins.py
```diff
@@ -0,0 +1,243 @@
...
+        origin = origin if origin else site
+
+        super().__init__(origin, cache=cache)
+        self.site = site
+        self.job_url = job_url
+        self.client = JenkinsClient(site)
+
+    @metadata
+    def fetch(self, from_date=DEFAULT_DATETIME):
```
sduenas: If `from_date` parameter is not supported, it must be removed from the method.

perceval/backends/jenkins.py
```diff
@@ -0,0 +1,243 @@
...
+
+        for job in jobs:
+            njobs += 1
+            logger.debug("Adding builds from %s (%i/%i)" % (job['url'], njobs, len(jobs)))
+            raw_builds = self.client.get_builds(job['url'])
+            self._push_cache_queue(raw_builds)
+            self._flush_cache_queue()
+            builds = json.loads(raw_builds)
+            builds = builds['builds']
+            for build in builds:
```
sduenas: So, the main item we return here is builds. Why builds instead of jobs? I think it would make sense to return builds if  `from_date` was used but since it isn't used maybe it is better to return a job as an item.

perceval/backends/jenkins.py
```diff
@@ -0,0 +1,243 @@
...
+    def get_jobs(self):
+        """ Retrieve all jobs
+        """
+        url_jenkins = self.site + "/view/All/api/json"
+
+        req = requests.get(url_jenkins)
+        req.raise_for_status()
+        return req.text
+
+    def get_builds(self, job_url):
```
sduenas: Same comment as above. Since you have the URL of the site, you only just need the name of the job to retrieve its data.

perceval/backends/jenkins.py
```diff
@@ -0,0 +1,243 @@
...
+    @classmethod
+    def create_argument_parser(cls):
+        """Returns the Jenkins argument parser."""
+
+        parser = super().create_argument_parser()
+
+        # Jenkins options
+        group = parser.add_argument_group('Jenkins arguments')
+
+        group.add_argument("--site", required=True,
```
sduenas: `site` parameter is mandatory so it should be a positional argument instead an optional one. `job-url` should be also positional. The call to the command then would be:

```
$ perceval jenkins http://example.com
$ perceval jenkins http://example.com my-job
```

tests/test_jenkins.py
```diff
@@ -0,0 +1,292 @@
...
+
+import httpretty
+
+if not '..' in sys.path:
+    sys.path.insert(0, '..')
+
+from perceval.cache import Cache
+from perceval.errors import CacheError
+from perceval.backends.jenkins import Jenkins, JenkinsCommand, JenkinsClient
+
```
sduenas: Put two blank lines here.

tests/test_jenkins.py
```diff
@@ -0,0 +1,292 @@
...
+JENKINS_SERVER_URL = 'http://example.com/ci'
+JENKINS_JOBS_URL = JENKINS_SERVER_URL+'/view/All/api/json'
+JENKINS_JOB_BUILDS_URL = JENKINS_SERVER_URL + '/job/job1/api/json?depth=2'
+JENKINS_JOB_BUILDS_URL_1 = JENKINS_SERVER_URL + '/job/apex-build-brahmaputra/api/json?depth=2'
+JENKINS_JOB_BUILDS_URL_2 = JENKINS_SERVER_URL + '/job/apex-build-master/api/json?depth=2'
+
```
sduenas: Two blank lines here two.

tests/test_jenkins.py
```diff
@@ -0,0 +1,292 @@
...
+def read_file(filename, mode='r'):
+    with open(filename, mode) as f:
+        content = f.read()
+    return content
+
```
sduenas: And here another two! :-D

tests/test_jenkins.py
```diff
@@ -0,0 +1,292 @@
...
+class TestJenkinsBackend(unittest.TestCase):
```
sduenas: You also have to include some tests to check that metadata parameters are right. Specifically, `uuid` and `updated_on`. There are some tests in other backends that check those parameters.

tests/test_jenkins.py
```diff
@@ -0,0 +1,292 @@
...
+        response = client.get_jobs()
+
+        self.assertEqual(response, body)
+
+    @httpretty.activate
+    def test_get_builds(self):
+        """Test get_builds API call"""
+
+        # Set up a mock HTTP server
+        body = read_file('data/jenkins_jobs.json')
```
sduenas: To avoid problems, use the json you have defined for builds instead of this one.

sduenas:
Can you add some entry to the README file about the usage of this backend and can you add it to the list of available backends in `perceval` script?

perceval/backends/jenkins.py
acs: Done!

perceval/backends/jenkins.py
acs: Done!

perceval/backends/jenkins.py
acs: Sure! Done!

perceval/backends/jenkins.py
acs: Done!

perceval/backends/jenkins.py
acs: I think we can just remove the job_url case. In the past it was useful but right now, downloading a complete Jenkins site with around 4K builds it is around 7 minutes. So following the same approach than in other backends, just get all and filter later.

About the "--from-date", in jenkins there is history, but it is limited to a number of items. So it is no useful the --from-date param, IMHO.

perceval/backends/jenkins.py
acs: I have left it in the API so the call with this param is supported (for API compatibility with the rest of backends). But it is better to just remove it and make the API use developer life a bit complex but safer.

perceval/backends/jenkins.py
acs: This method is called with the information that comes from the JSON file, the URL for the job, and this URL is complete. So it is better this API, now that it is not used for supporting job_url from command line.

tests/test_jenkins.py
acs: Done!

tests/test_jenkins.py
acs: Done :+1:

tests/test_jenkins.py
acs: Done!

tests/test_jenkins.py
acs: Hmmm, this is the list of jobs that have the builds. Then you have the builds for each job in a separate file.

tests/test_jenkins.py
acs: Done!

perceval/backends/jenkins.py
acs: site now is url, to be consistent with other backends. It is a positional argument.

And "my-job" param is not supported anymore.

Done!

perceval/backends/jenkins.py
acs: What we need to use for this backend is the builds info, which includes also the job. So IMHO it is enough to work just directly with builds.

perceval/backends/__init__.py
```diff
@@ -27,7 +27,7 @@
 from .jira import Jira, JiraCommand
 from .mbox import MBox, MBoxCommand
 from .stackexchange import StackExchange, StackExchangeCommand
-
+from .jenkins import Jenkins, JenkinsCommand
```
sduenas: Would you like to place this library ordered by name. The same with the keywords on both dictionaries below.

perceval/backends/jenkins.py
sduenas: Understand it!

perceval/backends/__init__.py
acs: Sure! Done!

PR #40, opened 2016-05-26
[tests] Jira added test_jira.py

zhquan:


tests/test_jira.py
```diff
@@ -0,0 +1,380 @@
...
+JIRA_SEARCH_URL = JIRA_SERVER_URL + '/rest/api/2/search'
+
+
+def read_file(filename, mode='r'):
+    with open(filename, mode) as f:
+        content = f.read()
+    return content
+
+
+class TestJiraBAckend(unittest.TestCase):
```
sduenas: Typo here. The A from backend is upper case.

sduenas: I miss an initialization test like we have in other tests files (i.e Bugzilla)

tests/test_jira.py
```diff
@@ -0,0 +1,380 @@
...
+        issues = Jira.parse_issues(raw_parse_json)
+
+        result = [issue for issue in issues]
+
+        parse = json.loads(parse_json)
+        self.assertDictEqual(result[0], parse[0])
+        self.assertDictEqual(result[1], parse[1])
+
+
+class TestJiraClient(unittest.TestCase):
```
sduenas: There are not any docstring on this class. Add them.

tests/test_jira.py
```diff
@@ -0,0 +1,380 @@
...
+    def test___init__(self):
```
sduenas: The name of this method should be `test_init` or `test_initialization`, as we have been doing with other tests.

tests/test_jira.py
```diff
@@ -0,0 +1,380 @@
...
+
+        self.assertEqual(requests[0].method, 'GET')
+        self.assertRegex(requests[0].path, '/rest/api/2/search')
+        self.assertDictEqual(requests[0].querystring, expected_req_0)
+
+        self.assertEqual(requests[1].method, 'GET')
+        self.assertRegex(requests[1].path, '/rest/api/2/search')
+        self.assertDictEqual(requests[1].querystring, expected_req_1)
+
+        self.assertEqual(pages[0], expected_json[0])
```
sduenas: You don't have to compare the result with other files. You need to compare it with what you sent, that it is stored in `bodies_json` list.

tests/test_jira.py
```diff
@@ -0,0 +1,380 @@
...
+        self.assertDictEqual(httpretty.last_request().querystring, expected_req)
+
+
+class TestJiraCommand(unittest.TestCase):
+
+    @httpretty.activate
+    def test_parsing_on_init(self):
+        """Test if the class is initialized"""
+
+        args = ['--project', 'None',
```
sduenas: A project named 'None' and a certificate which its value is 'None'? Please fix this to avoid mistakes.

tests/test_jira.py
```diff
@@ -0,0 +1,380 @@
...
+
+        raw_parse_json = read_file('data/jira/jira_issues_page_1.json')
+        parse_json = read_file('data/jira/jira_issues_parse_expected.json')
+
+        issues = Jira.parse_issues(raw_parse_json)
+
+        result = [issue for issue in issues]
+
+        parse = json.loads(parse_json)
+        self.assertDictEqual(result[0], parse[0])
```
sduenas: I don't like this way of testing. For this tests, I prefer hardcoded values than compare results with files. I want to see what it is going on and what we are trying to test. Please remove this from the other tests and add some better checks.

tests/test_jira.py
```diff
@@ -0,0 +1,380 @@
...
+        jira = Jira(JIRA_SERVER_URL)
+
+        with self.assertRaises(CacheError):
+            _ = cache_issues = [cache_issue for cache_issue in jira.fetch_from_cache()]
+
+class TestJiraBackendParsers(unittest.TestCase):
+    """ Jira backend parsers tests"""
+
+    def test_parse_questions(self):
+        """ Test issue parsing """
```
sduenas: Don't put a whitespace before the first word.

tests/test_jira.py
```diff
@@ -0,0 +1,380 @@
...
+    def test_fetch_from_non_set_cache(self):
+        """Test if a error is raised when the cache was not set"""
+
+        jira = Jira(JIRA_SERVER_URL)
+
+        with self.assertRaises(CacheError):
+            _ = cache_issues = [cache_issue for cache_issue in jira.fetch_from_cache()]
+
+class TestJiraBackendParsers(unittest.TestCase):
+    """ Jira backend parsers tests"""
```
sduenas: Don't put a whitespace before the first word.

tests/test_jira.py
```diff
@@ -0,0 +1,422 @@
...
+        self.assertRegex(requests[0].path, '/rest/api/2/search')
+        self.assertDictEqual(requests[0].querystring, expected_req[0])
+
+        self.assertEqual(requests[1].method, 'GET')
+        self.assertRegex(requests[1].path, '/rest/api/2/search')
+        self.assertDictEqual(requests[1].querystring, expected_req[1])
+
+        self.assertEqual(len(issues), 3)
+
+        self.assertEqual(issues[0]['perceval_version'], '0.1.0')
```
sduenas: None of these (`perceval_version`, `perceval_backend` and `backend_name`) need to be tested. These parameters are tested on `test_backend.py`. Instead, check one of the contents of the issue, for instance the id of the bug.

tests/test_jira.py
```diff
@@ -0,0 +1,422 @@
...
+
+        self.assertEqual(len(issues), 0)
+
+        request = httpretty.last_request()
+        self.assertEqual(request.method, 'GET')
+        self.assertRegex(request.path, '/rest/api/2/search')
+        self.assertDictEqual(request.querystring, expected_req)
+
+
+class TestJiraBAckendCache(unittest.TestCase):
```
sduenas: The 'A' must be lower cased.

tests/test_jira.py
sduenas: You still need to lower the A character.

tests/test_jira.py
```diff
@@ -0,0 +1,422 @@
...
+        self.tmp_path = tempfile.mkdtemp(prefix='perceval_')
+
+    def tearDown(self):
+        shutil.rmtree(self.tmp_path)
+
+    @httpretty.activate
+    def test_fetch_from_cache(self):
+        """Test whether a list of issues is returned from cache"""
+
+        bodies_json = read_file('data/jira/jira_issues_page_2.json')
```
sduenas: Why don't you use the same data sets that you use on `test_fetch()`? It's more easier to understand what we're doing with the test and includes a full case.

tests/test_jira.py
```diff
@@ -0,0 +1,422 @@
...
+
+class TestJiraCommand(unittest.TestCase):
+
+    @httpretty.activate
+    def test_parsing_on_init(self):
+        """Test if the class is initialized"""
+
+        args = ['--project', 'Perceval Jira',
+                '--verify', False,
+                '--cert', 'None',
```
sduenas: `--cert` parameter should be a string. I know that here, None is a string but to avoid mistakes put anything different.

tests/test_jira.py
```diff
@@ -0,0 +1,437 @@
...
+
+def read_file(filename, mode='r'):
+    with open(filename, mode) as f:
+        content = f.read()
+    return content
+
+
+class TestJiraBackend(unittest.TestCase):
+    """Jira backend tests"""
+
```
sduenas: The only thing I miss here is a test where you check the initialization values.

For instance: https://github.com/grimoirelab/perceval/blob/master/tests/test_bugzillarest.py#L122

PR #45, opened 2016-06-23
[Git] Adding support for fetching branches.

### LINKED ISSUE #43: [git] Option for backends.git.Git for parsing only some specific branch
In some cases, it is convenient to parse only some specific branch, instead of all of them. It would be nice having that for backends.git.Git.fetch, probably as an additional argument.

I will produce a pull request for this whenever I have some time...

jgbarah:
This patch adds support to the git backend for having a new parameter to allow for fetching commits only from a certain list of branches. Closes #43

### DIFF
--- a/perceval/__init__.py
+++ b/perceval/__init__.py
@@ -18,6 +18,7 @@
 #
 # Authors:
 #     Santiago Dueñas <sduenas@bitergia.com>
+#
 
 import logging
 
--- a/perceval/backends/git.py
+++ b/perceval/backends/git.py
@@ -55,7 +55,7 @@ class Git(Backend):
     :raises RepositoryError: raised when there was an error cloning or
         updating the repository.
     """
-    version = '0.2.0'
+    version = '0.3.0'
 
     def __init__(self, uri, gitpath, cache=None, origin=None):
         origin = origin if origin else uri
@@ -65,7 +65,7 @@ class Git(Backend):
         self.gitpath = gitpath
 
     @metadata
-    def fetch(self, from_date=DEFAULT_DATETIME):
+    def fetch(self, from_date=DEFAULT_DATETIME, branches=None):
         """Fetch commits.
 
         The method retrieves from a Git repository or a log file
@@ -75,16 +75,28 @@ class Git(Backend):
         Take into account that `from_date` is ignored when the commits
         are fetched from a Git log file.
 
+        The list of branches is a list of strings, with the names of the
+        branches to fetch. If the list of branches is empty, no commit
+        is fetched. If the list of branches is None, all commits
+        for all branches will be fetched.
+
         The class raises a `RepositoryError` exception when an error occurs
         accessing the repository.
 
         :param from_date: obtain commits newer than a specific date
             (inclusive)
+        :param branches: names of branches to fetch from (default: None)
 
         :returns: a generator of commits
         """
-        logger.info("Fetching commits: '%s' git repository from %s",
-                    self.uri, str(from_date))
+        if branches is None:
+            branches_text = "all"
+        elif len(branches) == 0:
+            branches_text = "no"
+        else:
+            branches_text = ", ".join(branches)
+        logger.info("Fetching commits: '%s' git repository from %s; %s branches",
+                    self.uri, str(from_date), branches_text)
 
         # Ignore default datetime to avoid problems with git
         # or convert to UTC
@@ -94,7 +106,7 @@ class Git(Backend):
             from_date = datetime_to_utc(from_date)
 
         ncommits = 0
-        commits = self.__fetch_and_parse_log(from_date)
+        commits = self.__fetch_and_parse_log(from_date, branches)
 
         for commit in commits:
             yield commit
@@ -103,12 +115,12 @@ class Git(Backend):
         logger.info("Fetch process completed: %s commits fetched",
                     ncommits)
 
-    def __fetch_and_parse_log(self, from_date):
+    def __fetch_and_parse_log(self, from_date, branches):
         if os.path.isfile(self.gitpath):
             return self.parse_git_log_from_file(self.gitpath)
         else:
             repo = self.__create_and_update_git_repository()
-            gitlog = repo.log(from_date)
+            gitlog = repo.log(from_date, branches)
             return self.parse_git_log_from_iter(gitlog)
 
     def __create_and_update_git_repository(self):
@@ -195,6 +207,7 @@ class GitCommand(BackendCommand):
         self.outfile = self.parsed_args.outfile
         self.from_date = str_to_datetime(self.parsed_args.from_date)
         self.origin = self.parsed_args.origin
+        self.branches = self.parsed_args.branches
 
         if self.parsed_args.git_log:
             git_path = self.parsed_args.git_log
@@ -216,7 +229,8 @@ class GitCommand(BackendCommand):
         git log. Commits are converted to JSON objects and printed to the
         defined output.
         """
-        commits = self.backend.fetch(from_date=self.from_date)
+        commits = self.backend.fetch(from_date=self.from_date,
+                                    branches=self.branches)
 
         try:
             for commit in commits:
@@ -245,6 +259,10 @@ class GitCommand(BackendCommand):
         parser.add_argument('uri',
                             help="URI of the Git log repository")
 
+        # Optional arguments
+        parser.add_argument('--branches', dest='branches',
+                            nargs='+', type=str, default=None,
+                            help='Fetch commits only from these branches')
         return parser
 
 
@@ -619,7 +637,7 @@ class GitRepository:
         logging.debug("Git %s repository pulled into %s",
                       self.uri, self.dirpath)
 
-    def log(self, from_date=None, encoding='utf-8'):
+    def log(self, from_date=None, branches=None, encoding='utf-8'):
         """Read the commit log from the repository.
 
         The method returns the Git log of the repository using the
@@ -632,8 +650,14 @@ class GitRepository:
         When `from_date` is given, it gets the commits equal or older
         than that date. This date is given in a datetime object.
 
+        The list of branches is a list of strings, with the names of the
+        branches to fetch. If the list of branches is empty, no commit
+        is fetched. If the list of branches is None, all commits
+        for all branches will be fetched.
+
         :param from_date: fetch commits newer than a specific
             date (inclusive)
+        :param branches: names of branches to fetch from (default: None)
         :param encoding: encode the log using this format
 
         :returns: a generator where each item is a line from the log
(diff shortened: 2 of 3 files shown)


sduenas:
I think we need more discussion on this. I'm not against it but before opening the door to include these filters we need some chat.

To give you some context, when @acs and I designed _Perceval_ we decided to not include any filter but `from_date`. This is the only one included just for practical reasons because it give us the **resume** (incremental) mode. The main idea in _Perceval_ is to get as much data as the source give us. The more the better. Later, the stack of _GrimoireLab_ will be in charge of deal with it, filtering the data that it is not necessary.

If we include this, we open the door to include more filters on other backends. I'm not against it and I understand it is useful for those that will only use Perceval. I just want to know your opinion and if we're in the same page.

gpoo:
Thinking aloud:

I noticed that the Git class is implemented using subprocess. I think, if it were implemented using a library (say gitpython), then it could be implemented as an iterable.

Having an iterable, it could be possible to call `filter` without adding filters in the implementation of the classes, but in the callers.

Does this make sense?

jgbarah:
To @sduenas. The main reason for this kind of filters is when you use Perceval for specific purposes, and where efficiency matters. A filter like this could be implemented by filtering the output of Perceval on the whole repository, but it would mean unnecessary complexity, if that's the only thing you need, and loss of eficiency, since you need to parse a git log which is potentially much longer than the one you really need.

I guess the final issue is how we understand Perceval. If it is (1) just for getting all the information that we can get from any repository, no filters should be needed. But then probably in that case we should remove the current `from_date` filter, which is basically a similar case (I know about its practical reasons, but those are not that different from the `branches` case, I guess).

If Perceval is (2) a general-purpose library, which can get as much information a wanted from a given repository, but it can as well interact with the underlying API or command to get some specific data, we should implement this kind of filters and "specialized flags", to control what information you want.

I guess Perceval is now (1), and I don't think we have to invest in (2), except that somebody comes and implements things they need, they are not breaking anything, and there is a use case. Otherwise, Perceval is not going to be suitable for scenarios where is could be, without losing any advantage for mining whole repositories.

jgbarah:
To @gpoo. I'm not sure I understand. It *is* an iterator. At least, `fetch` is an iterator. The subprocess thing is just a kludge for parsing and iterating over the output of `git log` as soon as possible, and at the same time having access to the result of executing the `git log` command. Let me walk you through the code if you want. Or maybe I'm missing something?

But in any case, as I commented above, being able of filtering the output of `git log` for a specific purpose is not always that reasonable. For example, in the `branches` case, if you are to parse the complete output, you have to: (1) store it (probably in memory), because you will need to go through it backwards; (2) find the commits with the relevant refs; (3) for each of those, find its parents, recursively until the beginning of the `git log`story: those commits would be the result.

Now, compare this with just using `git log branch` instead of `git log`, and nothing else.

gpoo:
@jgbarah It was a quick though. But perhaps what I had in mind was a two steps fetch.

The first step you get the list of branches, and then you fetch per branch. Both of them iterators. In that way, you do not need to add extra parameters (something that @sduenas would like to avoid) and still be nice with the resources.

In both cases, you can apply filter to the iterable (if you need to) and only retrieve and/or process the ones you are interested to. And still keeping the API simple.

gpoo:
I wonder if the output would be the same,though (getting all commits in one go and getting them per branch).

jgbarah:
@gpoo Now I understand you. But have in mind that, in git, a branch is a branch up to the first commit in the graph. That is, if you get eg. all master, and then yo get all branch1, which forks from master at some point, you still have to get the complete branch1 up to the first commit in master. Therefore, the implementation would became complex.

In any case, I guess branches are just a special case of providing specific parameters to git log. In the end, if git log can do it, at least potentially Perceval could do it. You can figure out some case where you need commits between two dates, or between two commits, or which correspond to refs/tags/*. All of that can be done with git log.

But anyway, I guess this is probably the core of the  discussion: whether Perceval should match what the underlying API can do, or not.

sduenas:
As @jgbarah says, the key of the discussion is whether or not Perceval backends should implement the API of the repository they are accessing.

I think we can follow what @jgbarah suggests. We don't have time nor resources to implement any single option or API parameter but if someone comes with it, we can discuss if we should include it or not.

Anyway, I will start reviewing your PR and if everyone thinks we can follow that approach, I will merge it.

perceval/backends/git.py
```diff
@@ -80,11 +80,19 @@ def fetch(self, from_date=DEFAULT_DATETIME):
 
         :param from_date: obtain commits newer than a specific date
             (inclusive)
+        :param branches: consider only these branches; when None, consider
+            all branches. Has an effect only if `gitpath´ is a repository.
+            Default: None.
+        :type branches: list of str
```
sduenas: We don't use this tag on Perceval. Please remove it.

perceval/backends/git.py
```diff
@@ -80,11 +80,19 @@ def fetch(self, from_date=DEFAULT_DATETIME):
 
         :param from_date: obtain commits newer than a specific date
             (inclusive)
+        :param branches: consider only these branches; when None, consider
```
sduenas: Which is the behaviour when `branches` is an empty list?

perceval/backends/git.py
```diff
@@ -634,6 +642,10 @@ def log(self, from_date=None, encoding='utf-8'):
 
         :param from_date: fetch commits newer than a specific
             date (inclusive)
+        :param branches: consider only these branches; when None, consider
```
sduenas: You should remove the section about `gitpath` because this client only works when the given directory is a git repository.

perceval/backends/git.py
```diff
@@ -642,11 +654,16 @@ def log(self, from_date=None, encoding='utf-8'):
...
         cmd_log = ['git', 'log', '--raw', '--numstat', '--pretty=fuller',
-                   '--decorate=full', '--all', '--reverse', '--topo-order',
-                   '--parents', '-M', '-C', '-c', '--remotes=origin']
+                   '--decorate=full', '--reverse', '--topo-order',
+                   '--parents', '-M', '-C', '-c']
         if from_date:
             dt = from_date.strftime("%Y-%m-%d %H:%M:%S %z")
             cmd_log.append('--since=' + dt)
+        if branches:
+            branches = ['remotes/origin/' + branch for branch in branches]
```
sduenas: As I can understand, when `branches` is an empty list or `None` the behaviour is to fetch every commit. Is this the semantic we want for an empty list?

tests/test_git.py
```diff
@@ -146,6 +146,84 @@ def test_fetch_since_date(self):
 
         shutil.rmtree(new_path)
 
+    def test_fetch_branch(self):
+        """Test whether commits are fetched from a Git repository for a given branch"""
+
+        new_path = os.path.join(self.tmp_path, 'newgit')
+
+        import logging
```
sduenas: This is not needed here.

tests/test_git.py
```diff
@@ -146,6 +146,84 @@ def test_fetch_since_date(self):
...
+                    'c6ba8f7a1058db3e6b4bc6f1090e932b107605fb',
+                    '589bb080f059834829a2a5955bebfd7c2baa110a',
+                    'ce8e0b86a1e9877f42fe9453ede418519115f367',
+                    '51a3b654f252210572297f47597b31527c475fb8',
+                    '456a68ee1407a77f3e804a30dff245bb6c6b872f']
+
+        self.assertEqual(len(commits), len(expected))
+
+        for x in range(len(commits)):
+            print (expected[x])
```
sduenas: Remove `print` statements here and on lines 195 and 218.

tests/test_git.py
```diff
@@ -146,6 +146,84 @@ def test_fetch_since_date(self):
 
         shutil.rmtree(new_path)
 
+    def test_fetch_branch(self):
```
sduenas: I miss a test case when `branches is an empty list.

perceval/backends/git.py
jgbarah: How should I specify the type of the parameter?

perceval/backends/git.py
jgbarah: Good point. Probably nonsense... I guess I should use [] instead of None, and have the current None behaviour (get all commits from all branches) implemented as [] instead. Is that ok?

perceval/backends/git.py
jgbarah: OK

perceval/backends/git.py
jgbarah: See above. I'm proposing merging both cases in just the empty list one, and yes, fetch every commit in that case.

tests/test_git.py
jgbarah: Oooops. Sorry about that.

tests/test_git.py
jgbarah: Oooops. Sorry. That's the rests of a test that I was doing to check what happened in some circunstances...

jgbarah:
New version submitted. Let me know what do you think about it, please.

perceval/__init__.py
```diff
@@ -17,7 +17,8 @@
 # Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
 #
 # Authors:
-#     Santiago Dueñas <sduenas@bitergia.com>
+#       Jesus M. Gonzalez-Barahona <jgb@bitergia.com>
```
sduenas: Why did you remove me from `__init__.py` file ???

perceval/backends/git.py
```diff
@@ -65,7 +65,7 @@ def __init__(self, uri, gitpath, cache=None, origin=None):
         self.gitpath = gitpath
 
     @metadata
-    def fetch(self, from_date=DEFAULT_DATETIME):
+    def fetch(self, from_date=DEFAULT_DATETIME, branches=[]):
```
sduenas: `branches` cannot has an empty list nor a list as default value. It is a bad practice in Python and can hide bugs.
Please read [Common Gotchas in Python](http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments) to know more about it.

sduenas: In my opinion, the default value of branches should be `None`. When `None` is given, no filters are applied and all commits are returned. When the list is empty, it means that there's a filter but not branches were given, so no commits will be returned. It could only be useful just for profiling purposes but I think it fits well with the meaning of the filter. In addition, a message can be logged to warns about that fact.

sduenas: By the way, `GitCommand` should also support this feature. You need to manage it on the class. You can do the same I did on `telegram` class with the `chats` filter.

https://github.com/grimoirelab/perceval/blob/master/perceval/backends/telegram.py#L307

sduenas:
@jgbarah please check my comments

perceval/__init__.py
jgbarah: No idea. Sorry about that. Probably I used some template of mine or something :-( I'll revert that.

perceval/backends/git.py
jgbarah: OK, I will do that as you comment.

jgbarah:
Done. New commit pushed.

sduenas:
@jgbarah, everything looks great but you have to fix one small thing. You forgot to increase the version of this backend. Go to line 58 and replace version `0.2.0` by `0.3.0`.

Regarding your commit message remember that we write in lower case the tag that is in between the square brackets. The first word after it should be a verb in its infinitive/imperative form. Finally, do not end the line with a period. With this changes, the message should be something like:

```
[git] Add support for fetching branches
```

jgbarah:
Done!

sduenas:
Sorry @jgbarah, there's another thing to fix. In `GitCommand` you missed to pass the `branches` argument to `fetch()` method (line 231). You will also need to store it on `__init__` method to use it later when `run()` is called.

jgbarah:
Done. Thanks.

sduenas:
Merged. Thanks @jgbarah .

PR #46, opened 2016-06-24
[mediawiki] Add mediawiki backend

acs:


perceval/backends/mediawiki.py
```diff
@@ -0,0 +1,475 @@
...
+        super().__init__(origin, cache=cache)
+        self.url = url
+        self.client = MediaWikiClient(url)
+
+    def __get_max_date(self, reviews):
+        """" Get the max date in unixtime format from reviews """
+        max_ts = 0
+        for review in reviews:
+            ts = str_to_datetime(review['timestamp'])
+            ts = ts.replace(tzinfo=dateutil.tz.tzutc())
```
sduenas: If you want to get a epoch timestamp you should use  `str_to_datetime` and [datetime_to_utc](https://github.com/grimoirelab/perceval/blob/master/perceval/utils.py#L68). Only after it you would be able to call to `timestamp()`.

perceval/backends/mediawiki.py
```diff
@@ -0,0 +1,475 @@
...
+
+    def __init__(self, url, cache=None, origin=None):
+        origin = origin if origin else url
+
+        super().__init__(origin, cache=cache)
+        self.url = url
+        self.client = MediaWikiClient(url)
+
+    def __get_max_date(self, reviews):
+        """" Get the max date in unixtime format from reviews """
```
sduenas: Don't forget to remove leading and trailing spaces on docstrings.

perceval/backends/mediawiki.py
```diff
@@ -0,0 +1,475 @@
...
+    version = '0.1.0'
+
+    def __init__(self, url, cache=None, origin=None):
+        origin = origin if origin else url
+
+        super().__init__(origin, cache=cache)
+        self.url = url
+        self.client = MediaWikiClient(url)
+
+    def __get_max_date(self, reviews):
```
sduenas: This is a //private// method so it should be included after the public methods. Review all the other methods.

perceval/backends/mediawiki.py
```diff
@@ -0,0 +1,475 @@
...
+
+        :returns: a generator of pages
+        """
+
+        self._purge_cache_queue()
+
+        if from_date == DEFAULT_DATETIME:
+            from_date = None
+        else:
+            from_date = datetime_to_utc(from_date).replace(tzinfo=None)
```
sduenas: Why do you need to replace the `tzinfo` value here?

perceval/backends/mediawiki.py
```diff
@@ -0,0 +1,475 @@
...
+
+        mediawiki_version = self.client.get_version()
+        logging.info("MediaWiki version: %s", mediawiki_version)
+
+        if mediawiki_version[0] >= 1 and mediawiki_version[1] >= 27:
```
sduenas: If the major version is 2, them the minor version should be equal or greater than 27 to work with this version.

perceval/backends/mediawiki.py
```diff
@@ -0,0 +1,475 @@
...
+
+        self._purge_cache_queue()
+
+        if from_date == DEFAULT_DATETIME:
+            from_date = None
+        else:
+            from_date = datetime_to_utc(from_date).replace(tzinfo=None)
+
+        mediawiki_version = self.client.get_version()
+        logging.info("MediaWiki version: %s", mediawiki_version)
```
sduenas: Use `logger` instead of `logging`.

perceval/backends/mediawiki.py
```diff
@@ -0,0 +1,475 @@
...
+
+        logger.info("Looking for pages at url '%s'", self.url)
+
+        self._purge_cache_queue()
+        npages = 0  # number of pages processed
+
+        # The pages are organized in content namespaces
+        raw_namespaces = self.client.get_namespaces()
+        namespaces = json.loads(raw_namespaces)["query"]["namespaces"]
+        namespaces_contents = [ns for ns in namespaces if 'content' in namespaces[ns].keys()]
```
sduenas: I think you should explain why you do manage the namespaces this way and why some are filtered and some others not.

perceval/backends/mediawiki.py
```diff
@@ -0,0 +1,475 @@
...
+                cause = "Can't get incremental pages older than %i days." % MAX_RECENT_DAYS
+                cause += " Do a complete analysis without from_date for older changes."
+                raise BackendError(cause=cause)
+
+        self._purge_cache_queue()
+        nns = 0 # number of namespaces
+        npages = 0  # number of pages processed
+
+        # The pages are organized in content namespaces
+        raw_namespaces = self.client.get_namespaces()
```
sduenas: This section of code is pretty similar (or equal) to the one that starts on line 121. I think you can move this to a common method.

perceval/backends/mediawiki.py
```diff
@@ -0,0 +1,475 @@
...
+            for ns in namespaces_contents:
+                nns += 1
+                apcontinue = '' # pagination for getting pages
+                logger.debug("Getting pages for namespace: %s", ns)
+                while apcontinue is not None:
+                    raw_pages = self.client.get_pages(ns, apcontinue)
+                    self._push_cache_queue(raw_pages)
+                    self._flush_cache_queue()
+                    data_json = json.loads(raw_pages)
+                    if 'continue' in data_json:  # 1.27
```
sduenas: If this works for versions prior to 1.27, why do support 1.27 here?

perceval/backends/mediawiki.py
```diff
@@ -0,0 +1,475 @@
...
+            for pages_dict in pages_dicts:
+                if pages_dict in data_json['query']:
+                    pages_json = data_json['query'][pages_dict]
+                    for page in pages_json:
+                        page_reviews = self.__build_page_reviews(page, json.loads(next(cache_items)))
+                        yield page_reviews
+
+    @staticmethod
+    def metadata_id(item):
+        """Extracts the identifier from a Build item."""
```
sduenas: Wrong comment

perceval/backends/mediawiki.py
```diff
@@ -0,0 +1,475 @@
...
+        return str(item['pageid'])
+
+    @staticmethod
+    def metadata_updated_on(item):
+        """Extracts the update time from a MediaWiki item.
+
+        The timestamp is extracted from 'timestamp' field.
```
sduenas: Wrong comment. It should be `update` field instead.

perceval/backends/mediawiki.py
```diff
@@ -0,0 +1,475 @@
...
+    projects in a MediaWiki node.
+
+    :param url: URL of jenkins node: https://build.opnfv.org/ci
+
+    :raises HTTPError: when an error occurs doing the request
+    """
+
+    def __init__(self, url):
+        self.url = url
+        self.api_url = self.url + "/" + "api.php"
```
sduenas: Would you mind to use `urljoin` method from `utils` module?

perceval/backends/mediawiki.py
```diff
@@ -0,0 +1,475 @@
...
+
+    def get_namespaces(self):
+        """ Retrieve all contents namespaces
+        """
+        ns_query = "action=query&meta=siteinfo&siprop=namespaces"
```
sduenas: The parameters attributes should be build using `requests` package. See how we do it on [Bugzilla client](https://github.com/grimoirelab/perceval/blob/master/perceval/backends/bugzilla.py#L550) and on its [call method](https://github.com/grimoirelab/perceval/blob/master/perceval/backends/bugzilla.py#L588).

You should change this in the rest of the code.

perceval/backends/mediawiki.py
```diff
@@ -0,0 +1,475 @@
...
+            raise RuntimeError(str(e))
+
+    @classmethod
+    def create_argument_parser(cls):
+        """Returns the MediaWiki argument parser."""
+
+        parser = super().create_argument_parser()
+
+        # MediaWiki options
+        group = parser.add_argument_group('MediaWiki arguments')
```
sduenas: You can remove this because it is not used. There aren't any optional parameter related to this backend.

perceval/backends/mediawiki.py
```diff
@@ -0,0 +1,475 @@
...
+                self._flush_cache_queue()
+                page_reviews = self.__build_page_reviews(page, json.loads(revisions_raw))
+                yield page_reviews
+                npages += 1
+
+        logger.info("Total number of pages: %i", npages)
+
+
+
+    def __fetch_pre1_27(self, from_date=None):
```
sduenas: This method is too long. I think you can split it in several methods for a better understanding.

sduenas:
@acs can you proceed to fix my comments?

perceval/backends/mediawiki.py
acs: Ok, analyzing the difference between what it is done in datetime_to_utc and the current approach. In this code it is missing to add to the date with the timezone before removing it (converting it to utc timezone). Changed!

perceval/backends/mediawiki.py
acs: Fixed and reviewed in the complete mediawiki.py file.

perceval/backends/mediawiki.py
acs: Done! Reviewed in all the mediawiki.py file.

perceval/backends/mediawiki.py
acs: Fixed!

perceval/backends/mediawiki.py
acs: No, of course. Done!

perceval/backends/mediawiki.py
acs: Done!

acs:
@sduenas working in the comments ...

perceval/backends/mediawiki.py
acs: I am using it later the current time without timezone in to compare dates, I need that from_date does not have timezone (it is an utc datetime in anycase).

https://github.com/acs/perceval/blob/mediawiki/perceval/backends/mediawiki.py#L221

But we can change it.

perceval/backends/mediawiki.py
acs: Fixed!

perceval/backends/mediawiki.py
acs: Fixed!

perceval/backends/mediawiki.py
acs: Done!

perceval/backends/mediawiki.py
acs: I have changed the code and I can find the similar code. I have thought during code of sharing more code between implementations but there are several things that make implementations different. But we can try to refactor it more. Let's review it after this first iteration.

perceval/backends/mediawiki.py
acs: Right! Old code ... removed!

perceval/backends/mediawiki.py
acs: Fixed!

perceval/backends/mediawiki.py
acs: Done!

perceval/backends/mediawiki.py
acs: Ok, I will review it with the above comment of sharing code between fetch implementations.

acs:
@sduenas all comments addressed. Could you check? Thanks!

perceval/backends/mediawiki.py
sduenas: `datetime_to_utc` adds a timezone value when it is called. When `from_date` does not have timezone it adds UTC by default. Later dates must be compared using a timezone too.

perceval/backends/mediawiki.py
```diff
@@ -0,0 +1,517 @@
...
+        """
+        if not self.cache:
+            raise CacheError(cause="cache instance was not provided")
+
+        cache_items = self.cache.retrieve()
+
+        pages_dicts = ['allrevisions','allpages','recentchanges']
+
+        # Namespaces -> pages -> revisions
+        next(cache_items) # First namespaces, we don't need them
```
sduenas: If we don't need them, why then should we store them?

perceval/backends/mediawiki.py
```diff
@@ -0,0 +1,517 @@
...
+
+        The timestamp is extracted from 'timestamp' field.
+        This date is a UNIX timestamp but needs to be converted to
+        a float value.
+
+        :param item: item generated by the backend
+
+        :returns: a UNIX timestamp
+        """
+        updated = None
```
sduenas: I don't see this. Does it means there are items that don't have a `updated` value???

perceval/backends/mediawiki.py
```diff
@@ -0,0 +1,517 @@
...
+    @staticmethod
+    def metadata_id(item):
+        """Extracts the identifier from a wiki page."""
+        return str(item['pageid'])
+
+    @staticmethod
+    def metadata_updated_on(item):
+        """Extracts the update field from a MediaWiki item.
+
+        The timestamp is extracted from 'timestamp' field.
```
sduenas: Fix this comment, please.

perceval/backends/mediawiki.py
acs: So the idea is to have all dates, that are in UTC, with timezones (00:00). Ok, I will change the code with that in mind.

perceval/backends/mediawiki.py
```diff
@@ -0,0 +1,517 @@
...
+        namespaces = json.loads(raw_namespaces)["query"]["namespaces"]
+        namespaces_contents = [ns for ns in namespaces if 'content' in namespaces[ns].keys()]
+        self._push_cache_queue(raw_namespaces)
+        self._flush_cache_queue()
+
+        arvcontinue = '' # pagination for getting revisions and their pages
+        while arvcontinue is not None:
+            raw_pages = self.client.get_pages_from_allrevisions(namespaces_contents, from_date, arvcontinue)
+            self._push_cache_queue(raw_pages)
+            self._flush_cache_queue()
```
sduenas: We use  `flush_cache_queue()` method when we know that everything was parsed and yield. I think it is better to put this after the `for page in pages_json` block (line 193) and remove nested calls to this method on places like `__get_page_reviews` and so on.

perceval/backends/mediawiki.py
```diff
@@ -0,0 +1,517 @@
...
+        :returns: a generator of pages
+        """
+
+        logger.info("Looking for pages at url '%s'", self.url)
+
+        self._purge_cache_queue()
+        npages = 0  # number of pages processed
+
+        # The pages are organized in content namespaces
+        raw_namespaces = self.client.get_namespaces()
```
sduenas: This code is the same that the one you use on line 280. This should be moved to a method.

perceval/backends/mediawiki.py
acs: Good point. I thought the same but in order to understand how the pages were retrieved it could be useful (for debugging the cache for example). But we can remove it, yes.

So, do you prefer removing them? I think it is the best option.

perceval/backends/mediawiki.py
acs: Yes.

perceval/backends/mediawiki.py
sduenas: Yes, take into account that `from_date` can include a timezone, so we use `datetime_to_utc` for that reason. It converts `from_date` to UTC (using timezone info) and the rest of the dates should be used in the same way.

perceval/backends/mediawiki.py
sduenas: Yes, please remove it.

perceval/backends/mediawiki.py
sduenas: Which are those cases? The first edition of a page?

perceval/backends/mediawiki.py
acs: Ok, my thought was that it the date is converted to UTC, the timezone is always 00:00 so we don't need to include it. But it is a good idea to have always timezones in the dates. Changing it!

sduenas: That's true but in Python there are to types of dates, naive and aware. Naive does not have information about timezones but aware does. Both types of dates can be compared between them. The interpreter raises an exception when it happens. For our sake, it is better to use always aware dates.

perceval/backends/mediawiki.py
acs: Done!

perceval/backends/mediawiki.py
acs: Ok, maybe it is good to flush the cache before all is parsed and yield so if a problem occurs, we have a cache of the already processed items. What do you think? In any case, I will implement now your suggestion. Done!

perceval/backends/mediawiki.py
acs: Done!

perceval/backends/mediawiki.py
acs: No, it was a corner case. I will try to reproduce it and report here the results!

PR #47, opened 2016-07-04
Mozilla reps backend

acs:


perceval/backends/mozilla_reps.py
```diff
@@ -0,0 +1,263 @@
...
+
+from ..backend import Backend, BackendCommand, metadata
+from ..cache import Cache
+from ..errors import BackendError, CacheError, ParseError
+
+from ..utils import (DEFAULT_DATETIME,
+                     datetime_to_utc,
+                     str_to_datetime,
+                     urljoin)
+
```
sduenas: 2 blank lines go here!!!

perceval/backends/mozilla_reps.py
```diff
@@ -0,0 +1,263 @@
...
+logger = logging.getLogger(__name__)
+
```
sduenas: And 2 blank lines here too.

perceval/backends/mozilla_reps.py
```diff
@@ -0,0 +1,263 @@
...
+    :param url: MozillaReps url
+    :param cache: cache object to store raw data
+    :param origin: identifier of the repository; when `None` or an
+        empty string are given, it will be set to `url` value
+    """
+    version = '0.1.0'
+
+    def __init__(self, url=None, cache=None, origin=None):
+        if not url:
+            url = "https://reps.mozilla.orghttps://reps.mozilla.org "
```
sduenas: This should be a constant outside this class. In fact, there's a bug on the URL.

URL field should be a mandatory argument, I think.

perceval/backends/mozilla_reps.py
```diff
@@ -0,0 +1,263 @@
...
+
+        :returns: a generator of events
+        """
+
+        logger.info("Looking for events at url '%s'", self.url)
+
+        self._purge_cache_queue()
+
+        nevents = 0  # number of events processed
+        nevents_total = 0  # number of events from API data
```
sduenas: Rename it to tevents

perceval/backends/mozilla_reps.py
```diff
@@ -0,0 +1,263 @@
...
+            events_data = json.loads(raw_events)
+            try:
+                logger.debug("Reps: %i/%i", nevents, nevents_total)
+                nevents_total = events_data['meta']['total_count']
+                events_next_uri = events_data['meta']['next']
+                if not events_next_uri:
+                    more_events = False
+                events = events_data['objects']
+            except:
+                cause = ("Bad JSON format for mozilla_events: %s" % (events_data))
```
sduenas: These parenthesis are not needed.

perceval/backends/mozilla_reps.py
```diff
@@ -0,0 +1,263 @@
...
+                cache.clean()
+            else:
+                cache.backup()
+        else:
+            cache = None
+
+        self.backend = MozillaReps(self.url, cache=cache, origin=self.origin)
+
+    def run(self):
+        """Fetch and print the Builds.
```
sduenas: Check this docstring. There are many references to **builds**.

perceval/backends/mozilla_reps.py
```diff
@@ -0,0 +1,263 @@
...
+        Builds are converted to JSON objects and printed to the
+        defined output.
+        """
+        if self.parsed_args.fetch_cache:
+            builds = self.backend.fetch_from_cache()
+        else:
+            builds = self.backend.fetch()
+
+        try:
+            for build in builds:
```
sduenas: These aren't builds.

perceval/backends/mozilla_reps.py
```diff
@@ -0,0 +1,263 @@
...
+                raise ParseError(cause=cause)
+            for rep in events:
+                yield rep
+                nevents += 1
+            self._flush_cache_queue()
+
+        logger.info("Total number of events: %i (%i expected)", nevents, nevents_total)
+
+    @metadata
+    def fetch_from_cache(self):
```
sduenas: These aren't builds, again.

sduenas: Some logging messages would be useful too.

perceval/backends/mozilla_reps.py
```diff
@@ -0,0 +1,263 @@
...
+        logger.info("Looking for events at url '%s'", self.url)
+
+        self._purge_cache_queue()
+
+        nevents = 0  # number of events processed
+        nevents_total = 0  # number of events from API data
+        more_events = True # There are more events to be processed
+        events_next_uri = None # URI for the next events query
+
+        while more_events:
```
sduenas: Please add some blank lines in this while loop for a better understanding of what the loop does. For instance:

```
        while more_events:
            raw_events = self.client.get_events(events_next_uri)
            self._push_cache_queue(raw_events)
            events_data = json.loads(raw_events)

            try:
                logger.debug("Reps: %i/%i", nevents, nevents_total)
                
                nevents_total = events_data['meta']['total_count']
                events_next_uri = events_data['meta']['next']

                if not events_next_uri:
                    more_events = False

                events = events_data['objects']
            except:
                cause = "Bad JSON format for mozilla_events: %s" % (events_data)
                raise ParseError(cause=cause)

            for rep in events:
                yield rep
                nevents += 1

            self._flush_cache_queue()

File: perceval/backends/mozilla_reps.py [hunk 10]
```diff
@@ -0,0 +1,263 @@
...
+        This date is a UNIX timestamp but needs to be converted to
+        a float value.
+
+        :param item: item generated by the backend
+
+        :returns: a UNIX timestamp
+        """
+        return MozillaReps.__metadata_updated_on_event(item)
+
+    @staticmethod
```
sduenas: Why do we need this two methods here? Why don't you access the data directly on `metadata_id` and `metadta_updated_on`?

perceval/backends/mozilla_reps.py
```diff
@@ -0,0 +1,263 @@
...
+
+    This class implements a simple client to retrieve builds from
+    projects in a MozillaReps node.
+
+    :param url: URL of MozillaReps url https://reps.mozilla.org
+
+    :raises HTTPError: when an error occurs doing the request
+    """
+
+    MAX_ITEMS_EV = "40"  # max number of events per iteration
```
sduenas: Usually, this is outside of the client because we think that the number of items requested are configurable. I don't know if this is the case or not.

perceval/backends/mozilla_reps.py
```diff
@@ -0,0 +1,263 @@
...
+        """
+        logger.debug("MozillaReps client calls API: %s params: %s",
+                     self.api_events_url, str(params))
+
+        req = requests.get(self.api_events_url, params=params)
+        req.raise_for_status()
+
+        return req.text
+
+    def get_events(self,  next_url=None):
```
sduenas: `next_url` is never used, Why do we need then it?

tests/test_mozilla_reps.py
```diff
@@ -0,0 +1,250 @@
...
+            return (200, headers, body)
+
+        httpretty.register_uri(httpretty.GET,
+                               MOZILLA_REPS_API_EVENTS,
+                               responses=[
+                                    httpretty.Response(body=request_callback)
+                               ])
+
+    @classmethod
+    def check_events_contents(cls, testObj, events):
```
sduenas: This should be outside this in a base class for testing or similar. It is not a good practice to pass the same class you're testing to get access to `unittest` methods.

tests/test_mozilla_reps.py
```diff
@@ -0,0 +1,250 @@
...
+        if len(events) > 1:
+            testObj.assertEqual(events[1]['data']['estimated_attendance'], 50)
+            testObj.assertEqual(events[1]['data']['local_start'], '2012-06-09T09:00:00')
+            testObj.assertEqual(events[1]['data']['start'], '2012-06-09T07:00:00')
+            testObj.assertEqual(events[1]['origin'], MOZILLA_REPS_SERVER_URL)
+            testObj.assertEqual(events[1]['uuid'], 'e83e53b1f1039f62c66d2a977e23df3adf5bdc87')
+            testObj.assertEqual(events[1]['updated_on'], 1339344000.0)
+
+
+class TestMozillaRepsBackend(unittest.TestCase):
```
sduenas: `ParseError` exception is not tested.

tests/test_mozilla_reps.py
```diff
@@ -0,0 +1,250 @@
...
+    def test_fetch_empty(self):
+        """Test whether it works when no jobs are fetched"""
+
+        HTTPServer.routes(empty=True)
+
+        mozilla_reps = MozillaReps(MOZILLA_REPS_SERVER_URL)
+        events = [event for event in mozilla_reps.fetch()]
+
+        self.assertEqual(len(events), 0)
+
```
sduenas: 2 blank lines!!!!

tests/test_mozilla_reps.py
```diff
@@ -0,0 +1,250 @@
...
+    def tearDown(self):
+        shutil.rmtree(self.tmp_path)
+
+    @httpretty.activate
+    def test_fetch_from_cache(self):
+        """Test whether the cache works"""
+
+        HTTPServer.routes()
+
+        # First, we fetch the builds from the server, storing them
```
sduenas: These aren't builds! Please check your comments and docstrings and fix them in the whole PR.

tests/test_mozilla_reps.py
```diff
@@ -0,0 +1,250 @@
...
+    """MozillaReps backend tests using a cache"""
+
+    def setUp(self):
+        self.tmp_path = tempfile.mkdtemp(prefix='perceval_')
+
+    def tearDown(self):
+        shutil.rmtree(self.tmp_path)
+
+    @httpretty.activate
+    def test_fetch_from_cache(self):
```
sduenas: You also have to check the number of requests send to the server to know whether it really retrieved the data from the cache or not.

sduenas:
@acs please check my comments.

As a general comment, I think we should change the name of this backend to `remo` because it is the tool that is used to manage the events of Mozilla community. Anyone can use this tool too to schudule her own events.

perceval/backends/mozilla_reps.py
acs: ok!

perceval/backends/mozilla_reps.py
acs: ok!

perceval/backends/mozilla_reps.py
acs: I am having issues with C&P in my touchpad, but I have tested all before sending the PR. Strange. In any case, fixed and moved to a constant.

About having a default URL, probably the only site using this software is the Mozilla Reps one, so it is sensible default IMHO. But if you are uncomfortable with it, I change it.

perceval/backends/mozilla_reps.py
acs: I like the "n" before the name to reflect it is a number, but ok, renamed!

perceval/backends/mozilla_reps.py
acs: Arrgg, I think I have removed all references to builds (from jenkins backend) but it seems I have not done it totally. Reviewing ... done! The case sensitive replacement was the problem.

perceval/backends/mozilla_reps.py
acs: Fixed!

perceval/backends/mozilla_reps.py
acs: Done both!

perceval/backends/mozilla_reps.py
acs: Done!

perceval/backends/mozilla_reps.py
acs: Legacy code from and older version that supports events and representatives. Fixed.

perceval/backends/mozilla_reps.py
acs: It is a max fixed value in the server.

perceval/backends/mozilla_reps.py
acs: Yes, it is used but in the last changes, I have not reflect the next_url case in the param logic!

Fixed!

tests/test_mozilla_reps.py
acs: It is not raise in the backend. The data received is plain JSON directly, so the parser is the json parser directly, we don't have any parsers to be tested in our code. Are you thinking in receiving bad JSON from the server for a network problem or something like that?

tests/test_mozilla_reps.py
acs: Done!

tests/test_mozilla_reps.py
acs: Fixed in the whole PR. Sorry about that!

tests/test_mozilla_reps.py
acs: Sure! Done!

tests/test_mozilla_reps.py
acs: Moved to a private method in TestMozillaRepsBackend. Only used there. Fixed!

acs:
@sduenas all comments addressed and renamed MozillaReps to ReMo. Please review it :)

perceval/backends/remo.py
```diff
@@ -0,0 +1,275 @@
...
+
+        while more_events:
+            raw_events = self.client.get_events(events_next_uri)
+            self._push_cache_queue(raw_events)
+            events_data = json.loads(raw_events)
+
+            try:
+                logger.debug("Reps: %i/%i", nevents, tevents)
+
+                tevents = events_data['meta']['total_count']
```
sduenas: I think here goes a `+=` operator and not a `=`

perceval/backends/remo.py
```diff
@@ -0,0 +1,275 @@
...
+                events_next_uri = events_data['meta']['next']
+
+                if not events_next_uri:
+                    more_events = False
+
+                events = events_data['objects']
+            except:
+                cause = ("Bad JSON format for mozilla_events: %s" % (events_data))
```
sduenas: Remove these parenthesis.

perceval/backends/remo.py
```diff
@@ -0,0 +1,275 @@
...
+
+        nevents = 0
+
+        for items in cache_items:
+            events = json.loads(items)['objects']
+            for event in events:
+                nevents += 1
+                yield event
+
+        logger.info("Retrieval process completed: %s bugs retrieved from cache",
```
sduenas: Wrong message!

perceval/backends/remo.py
```diff
@@ -0,0 +1,275 @@
...
+    @staticmethod
+    def metadata_id(item):
+        """Extracts the identifier from an event item."""
+        return str(item['event_url'])
+
+    @staticmethod
+    def metadata_updated_on(item):
+        """Extracts the update time from a ReMo item.
+
+        The timestamp is extracted from 'timestamp' field.
```
sduenas: Wrong message!

perceval/backends/remo.py
```diff
@@ -0,0 +1,275 @@
...
+        This date is a UNIX timestamp but needs to be converted to
+        a float value.
+
+        :param item: item generated by the backend
+
+        :returns: a UNIX timestamp
+        """
+        return float(parser.parse(item['end']).timestamp())
```
sduenas: Why are you calling here to `parser`? I think we have enough functions in `utils` module to deal with these cases.

perceval/backends/remo.py
```diff
@@ -0,0 +1,275 @@
...
+    MAX_ITEMS_EV = "40"  # max number of events per iteration
+
+    def __init__(self, url):
+        self.url = url
+        self.api_events_url = urljoin(self.url, '/api/v1/event/')
+        self.api_events_url += '/'  # API needs a final /
+
+    def call(self, params):
+        """Run an API command.
+        :param cgi: cgi command to run on the server
```
sduenas: Is this a CGI?

perceval/backends/remo.py
```diff
@@ -0,0 +1,275 @@
...
+        req.raise_for_status()
+
+        return req.text
+
+    def get_events(self,  next_url=None):
+        """Retrieve all events"""
+        offset = "0"
+        if next_url:
+            # /api/v1/event/?limit=40&offset=40
+            offset = next_url.split("offset=")[1]
```
sduenas: If what we need is to pass the offset, why don't we pass it directly? Anyone can use this without having an URL for that.

tests/test_remo.py
```diff
@@ -0,0 +1,260 @@
...
+
+        if len(events) > 1:
+            self.assertEqual(events[1]['data']['estimated_attendance'], 50)
+            self.assertEqual(events[1]['data']['local_start'], '2012-06-09T09:00:00')
+            self.assertEqual(events[1]['data']['start'], '2012-06-09T07:00:00')
+            self.assertEqual(events[1]['origin'], MOZILLA_REPS_SERVER_URL)
+            self.assertEqual(events[1]['uuid'], 'e83e53b1f1039f62c66d2a977e23df3adf5bdc87')
+            self.assertEqual(events[1]['updated_on'], 1339344000.0)
+
+
```
sduenas: I'm gonna kill you. This is not so difficult. Put only ONE blank line between methods of the same class!!!

tests/test_remo.py
```diff
@@ -0,0 +1,260 @@
...
+class TestReMoBackend(unittest.TestCase):
+    """ReMo backend tests"""
+
+    def test_initialization(self):
+        """Test whether attributes are initializated"""
+
+        remo = ReMo(MOZILLA_REPS_SERVER_URL, origin='test')
+
+        self.assertEqual(remo.url, MOZILLA_REPS_SERVER_URL)
```
sduenas: It would be useful to check when the `url` is `None` it is changed to the default value as the documentation says.

tests/test_remo.py
```diff
@@ -0,0 +1,260 @@
...
+        HTTPServer.routes()
+
+        # Test fetch events with their reviews
+        remo = ReMo(MOZILLA_REPS_SERVER_URL)
+
+        events = [page for page in remo.fetch()]
+
+        self.assertEqual(len(events), 4)
+
+        self.__check_events_contents(events)
```
sduenas: As the backend needs to do the pagination by itself, it will be useful to check that HTTP requests were done as you expected.

perceval/backends/remo.py
acs: Because this is specific logic for this API REST, so I find it more natural to have it in the client, and don't put the parsing logic in the user of the client.

perceval/backends/remo.py
acs: No, in this field it is included the total_count always.

perceval/backends/remo.py
acs: Yes, done!

perceval/backends/remo.py
acs: I fixed it in a commit just after you review it. Sorry about that.

perceval/backends/remo.py
acs: Fixed!

perceval/backends/remo.py
acs: Yes, and all of them use this parser.parse as basics. But yes, I can add convert to that if you are more comfortable with it. Done!

perceval/backends/remo.py
acs: Removed the doc about this non existent param.

tests/test_remo.py
acs: Ok, added the test. Good point.

tests/test_remo.py
acs: Yes, as in the rest of 537-1 lines done correctly. Fixed. Probably we should be more strict using PEP-8 so the editor is more useful with its warnings about this issues. Or define some way to automatic check the style of code files.

tests/test_remo.py
acs: Not sure about that. Checking the number of calls done? The params in each call?

tests/test_remo.py
acs: Done!

tests/test_remo.py
acs: Fixed

perceval/backends/remo.py
acs: Removed the doc for this non existent param.

perceval/backends/remo.py
acs: Fixed

perceval/backends/remo.py
acs: Fixed

perceval/backends/remo.py
acs: Done

tests/test_remo.py
acs: Are you thinking in the number of HTTP calls? The params in each call?

acs:
@sduenas completed last review iteration.

tests/test_remo.py
sduenas: You need to add a test here to check the base case when no URL is given to the backend.

tests/test_remo.py
sduenas: Yes, I meant the parameters sent on each call.

acs: Ok, done!

tests/test_remo.py
acs: Hmmm, it is just below:

    remo = ReMo(MOZILLA_REPS_SERVER_URL)

or am I missing something?

acs:
@sduenas last iteration completed!

tests/test_remo.py
sduenas: The call should be `ReMo()` and the code should assign to `url` the value of `MOZILLA_REPS_SERVER_URL`.

sduenas: Forget about it. I've just seen you added it to the bottom of this test.

PR #49, opened 2016-07-12
[perceval] Add Kitsune backend

acs:


perceval/backends/kitsune.py
```diff
@@ -0,0 +1,326 @@
...
+
+
+class Kitsune(Backend):
+    """Kitsune backend for Perceval.
+
+    This class retrieves the questions and answers from a
+    Kitsune url. To initialize this class a
+    url could be provided. If not, https://support.mozilla.org will be used.
+
+    questions and answers are returned from newer to older.
```
sduenas: Start with an uppercase letter.

perceval/backends/kitsune.py
```diff
@@ -0,0 +1,326 @@
...
+
+        for raw_questions in self.client.get_questions():
+            if not more_updated:
+                break
+            self._push_cache_queue(raw_questions)
+            try:
+                questions_data = json.loads(raw_questions)
+                tquestions = questions_data['count']
+                questions = questions_data['results']
+            except:
```
sduenas: This except block should check a exception. This can hide other critical exceptions.

perceval/backends/kitsune.py
```diff
@@ -0,0 +1,326 @@
...
+        logger.info("Looking for questions at url '%s' updated after %s",
+                    self.url, str(from_date))
+
+        self._purge_cache_queue()
+
+        nquestions = 0  # number of questions processed
+        tquestions = 0  # number of questions from API data
+        more_updated = True # more questions updated after from_date
+
+        for raw_questions in self.client.get_questions():
```
sduenas: I think you should use a `while` loop here instead of a `for`.  You use `mode_updated` to  exit from the loop twice. Why don't you use this value as a sentinel value in a `while`loop?

perceval/backends/kitsune.py
```diff
@@ -0,0 +1,326 @@
...
+            for question in questions:
+                if str_to_datetime(question['updated']) <= from_date:
+                    logging.info("No more questions updated after: %s (last %s)", from_date, question['updated'])
+                    more_updated = False
+                    break
+                question['answers_data'] = []
+                for raw_answers in self.client.get_question_answers(question['id']):
+                    self._push_cache_queue(raw_answers)
+                    answers = json.loads(raw_answers)['results']
+                    question['answers_data'] += answers
```
sduenas: Why `answers_data` and not just `answers`?

perceval/backends/kitsune.py
```diff
@@ -0,0 +1,326 @@
...
+
+        questions_raw = next(cache_items)
+        while questions_raw:
+            if not json.loads(questions_raw):
+                # Last item is the empty dict
+                questions_raw = next(cache_items)
+            questions = json.loads(questions_raw)['results']
+            for question in questions:
+                question['answers_data'] = []
+                for answers_raw in cache_items:
```
sduenas: As you are using `next()` call on other places, I think you should be stuck to it and use it here too.

perceval/backends/kitsune.py
```diff
@@ -0,0 +1,326 @@
...
+        return req.text
+
+    def get_questions(self):
+        """Retrieve questions from newer to older (-updated)"""
+
+        page = KitsuneClient.FIRST_PAGE
+        next_uri = None # URI for the next questions query
+
+        while True:
+            api_questions_url = urljoin(self.api_url, '/question')+'/'
```
sduenas: Operations must be separated by whitespaces.

perceval/backends/kitsune.py
```diff
@@ -0,0 +1,326 @@
...
+                break
+            page += 1
+
+    def get_question_answers(self, question_id):
+        """Retrieve all answers for a question from newer to older (-updated)"""
+
+        page = KitsuneClient.FIRST_PAGE
+
+        while True:
+            api_answers_url = urljoin(self.api_url, '/answer')+'/'
```
sduenas: Operations must be separated by whitespaces.

perceval/backends/kitsune.py
acs: Done!

perceval/backends/kitsune.py
acs: Ok, any exception in this code must stop the process and raise the Parse exception, but yes, it is cleaner to control the two cases: bad JSON syntax, and bad JSON fields. 

Fixed!

perceval/backends/kitsune.py
acs: If we use a while, we need to control to exit conditions:

* more_updated
* end of self.client.get_questions()

and the syntax won't be more readable than now? But I don't like the current version. How would you do it with a while?

perceval/backends/kitsune.py
acs: Because there is a answers field already with the total number of answers.

perceval/backends/kitsune.py
acs: The for does automatically the next. Are you thinking in changing it to a while and use next?

perceval/backends/kitsune.py
acs: Fixed!

perceval/backends/kitsune.py
acs: Fixed!

perceval/backends/kitsune.py
sduenas: I supposed that. :+1:

perceval/backends/kitsune.py
```diff
@@ -91,7 +89,8 @@ def fetch(self, from_date=DEFAULT_DATETIME):
                 questions_data = json.loads(raw_questions)
                 tquestions = questions_data['count']
                 questions = questions_data['results']
-            except:
+            except (ValueError, KeyError) as ex:
+                logging.error(ex)
```
sduenas: Do not use `logging`. Use `logger` object instead.

sduenas:
I don't see this whole thing about filtering by `pages`. It is so client specific and adds another type of filter. I don't like offset parameters either but we should go with it.
By the way, please check your PR, the tests failed on Travis with your last commit.

perceval/backends/kitsune.py
```diff
@@ -0,0 +1,358 @@
...
+        drop_questions = from_offset - page_offset
+
+        for raw_questions in self.client.get_questions(from_offset):
+            self._push_cache_queue(raw_questions)
+            try:
+                questions_data = json.loads(raw_questions)
+                tquestions = questions_data['count']
+                questions = questions_data['results']
+            except (ValueError, KeyError) as ex:
+                logging.error(ex)
```
sduenas: Change this. You have to use `logger` in the whole module.

perceval/backends/kitsune.py
```diff
@@ -0,0 +1,358 @@
...
+            url = KITSUNE_URL
+        origin = origin if origin else url
+
+        super().__init__(origin, cache=cache)
+        self.url = url
+        self.client = KitsuneClient(url)
+
+    @kitsune_metadata
+    @metadata
+    def fetch(self, from_offset=DEFAULT_OFFSET, from_date=DEFAULT_DATETIME):
```
sduenas: Rename `from_offset` to `offset` because it the name we are using in other backends.

perceval/backends/kitsune.py
```diff
@@ -0,0 +1,358 @@
...
+                nquestions += 1
+                self._push_cache_queue('{}') # Mark with empty dict end of question
+
+            logger.debug("Questions: %i/%i", nquestions, tquestions)
+
+            self._flush_cache_queue()
+
+        logger.info("Total number of questions: %i (%i total)", nquestions, tquestions)
+
+    @metadata
```
sduenas: I miss `kitsune_metadata` decorator here.

perceval/backends/kitsune.py
```diff
@@ -0,0 +1,383 @@
...
+        :offset: obtain questions after offset
+        :returns: a generator of questions
+        """
+        logger.info("Looking for questions at url '%s' using offset %s",
+                    self.url, str(offset))
+
+        self._purge_cache_queue()
+        # Add to the cache the offset so it can be used to recover from cache
+        self._push_cache_queue(offset)
+        self._flush_cache_queue()
```
sduenas: I think this call shoud be removed. It is only useful if the process was successful. If not, it will corrupt the cache.

acs: Removed! I have in mind that the flush should be done when a semantic unit is completed.

tests/test_kitsune.py
```diff
@@ -0,0 +1,337 @@
...
+        # Test fetch questions with their reviews
+        kitsune = Kitsune(KITSUNE_SERVER_URL)
+
+        # Get all questions
+        offset = 0
+        questions = [question for question in kitsune.fetch(offset=offset)]
+        self.assertEqual(len(questions), 4)
+        self.__check_questions_contents(questions)
+
+        # 4 total questions minus 2, 2 questions returned
```
sduenas: I would like to check on these cases that the `offset` assigned to each item is right one.

perceval/backends/kitsune.py
```diff
@@ -0,0 +1,383 @@
...
+
+            logger.debug("Questions: %i/%i", nquestions, tquestions)
+
+            self._flush_cache_queue()
+
+        logger.info("Total number of questions: %i (%i total)", nquestions, tquestions)
+
+    @kitsune_metadata
+    @metadata
+    def fetch_from_cache(self):
```
sduenas: This method is too long. I think you can split it in several private methods to make it more readable.

perceval/backends/kitsune.py
```diff
@@ -0,0 +1,383 @@
...
+
+
+class Kitsune(Backend):
+    """Kitsune backend for Perceval.
+
+    This class retrieves the questions and answers from a
+    Kitsune url. To initialize this class a
+    url could be provided. If not, https://support.mozilla.org will be used.
+
+    Questions and answers are returned from newer to older.
```
sduenas: Is this comment updated?

perceval/backends/kitsune.py
```diff
@@ -0,0 +1,383 @@
...
+
+    def __init__(self, url=None, cache=None, origin=None):
+        if not url:
+            url = KITSUNE_URL
+        origin = origin if origin else url
+
+        super().__init__(origin, cache=cache)
+        self.url = url
+        self.client = KitsuneClient(url)
+        self.cache_offset = 0
```
sduenas: I think you don't need this attribute anymore.

perceval/backends/kitsune.py
```diff
@@ -0,0 +1,383 @@
...
+                    self.url, str(offset))
+
+        self._purge_cache_queue()
+        # Add to the cache the offset so it can be used to recover from cache
+        self._push_cache_queue(offset)
+        self._flush_cache_queue()
+
+        nquestions = 0  # number of questions processed
+        tquestions = 0  # number of questions from API data
+        drop_questions = 0 # drop questions from page before the offset
```
sduenas: You don't need to initialize this variable here.

acs: Yep. Removed!

perceval/backends/kitsune.py
acs: No, good catch. Fixed!

perceval/backends/kitsune.py
acs: No! Removed!

perceval/backends/kitsune.py
acs: It is as long as the fetch method. Yes, three level for is a bad practice in a method.

But I would prefer to close a first version of this backend now, we are a bit late.

What do you think? If we refactor this fetch method, I would refactor also the main fetch.

tests/test_kitsune.py
acs: Good point. Adding it ...done!

tests/test_kitsune.py
```diff
@@ -0,0 +1,339 @@
...
+        offset = 2
+        questions = [question for question in kitsune.fetch(offset=offset)]
+        self.assertEqual(len(questions), 2)
+        self.assertEqual(questions[0]['offset'],2)
```
sduenas: Add a space between parameters :-)

acs: Done! Sorry!

PR #51, opened 2016-07-15
[gerrit] Fix #50

### LINKED ISSUE #50: [gerrit] Bad query when filtering several reviews using blacklist option
When executing gerrit backend with --blacklist option to filter reviews, these reviews are added to the ssh call as: 

    blacklist_reviews = " AND NOT (%s)" % (','.join(self.blacklist_reviews))

and it must be

    blacklist_reviews = " AND NOT (%s)" % (' OR '.join(self.blacklist_reviews))

so the result is

    ssh -p 29418 XXXXX@git.eclipse.org gerrit  query limit:1 '(status:open OR status:closed) AND NOT (75970 OR 67778 OR 74821)'

acs:


sduenas:
The commit is OK but I need a commit message explaining what the error was and why this fix the bug.

You can read this awesome post to get ideas: http://chris.beams.io/posts/git-commit/

PR #52, opened 2016-07-15
[jira] Add extra information in custom fields

albertinisg:
This PR adds the field `name` and `id` to every custom field, keeping its value.


perceval/backends/jira.py
```diff
@@ -163,6 +167,34 @@ def parse_issues(raw_page):
         for issue in issues:
             yield issue
 
+    @staticmethod
```
sduenas: I don't see these two methods as methods of this class. They are really specific to how fields are parsed. As we don't have a parser here I think it should be better to put them outside of this class, keeping them as module functions.
What do you think?

perceval/backends/jira.py
```diff
@@ -274,6 +311,14 @@ def get_issues(self, from_date):
                 issues = req.text
                 self.__log_status(start_at, tissues)
 
+    def get_fields(self):
+        """Retrieve all the fields available.  """
+        s = self.__init_session()
```
sduenas: I should rename `s` to `session`. In Python, is `s` is the common name used to store a string.

tests/test_jira.py
```diff
@@ -199,6 +273,24 @@ def test_fetch_empty(self):
...
+    @httpretty.activate
+    def test_filter_custom_fields(self):
+        """Test that all the fields returned are just custom"""
+        body = read_file('data/jira/jira_fields.json')
+
+        httpretty.register_uri(httpretty.GET,
+                               JIRA_FIELDS_URL,
+                               body=body, status=200)
+
+        jira = Jira(JIRA_SERVER_URL)
```
sduenas: You're testing a static method. You don't need to instantiate the class. You just need to call the method this way:
```
Jira.filter_custom_fields(json)
```

perceval/backends/jira.py
albertinisg: It's ok for me, but also all those kind of functionalities in other backends are still in the `Backend` class (maybe not as functions, but logic is there) , so I'm afraid we are breaking the structure followed until now.

perceval/backends/jira.py
albertinisg: Done in b2963caa0ea3d0de920b84c42b94b86825f67d60

tests/test_jira.py
albertinisg: Done in 51eaecf9348a9288a0a2a547a578868e042a51cd

perceval/backends/jira.py
sduenas: Not really. The methods we have in other classes are private methods that parse the input. Those methods are usually static methods so everyone can use them only to parse a stream of data. Specific methods that parse some parts of the streams are private and not static.
If you think the methods you have defined will be used by someone at some point, they should be public and static but as they are parsing specific (and this class doesn't have any parsing method) I think they should be out of this class.

albertinisg: Understood! :relaxed: One last thing, as I'm moving this functions outside the `Backend`, I guess I must add a test for `map_custom_fields` (as it won't be private), but this testing is already implicit in `test_fetch`. Does it makes sense to add another test? (in terms of coverage).

sduenas: Yes, you guessed it right. You need to add some tests for this function.
BTW, for this function, why don't you use this with [map()](https://docs.python.org/3/library/functions.html#map)?

albertinisg: I've been reading about the  [map()](https://docs.python.org/3/library/functions.html#map) function and seems it's exactly the same as:
```python
[f(x) for x in iterable]
```
So I've decided to go for that choice, also to get used with the list comprehensions. I've re-written the functions according to that approach:
```python
def map_custom_fields(custom_fields, issue):

    def complete_field(custom_fields, element):
        value = issue['fields'][element]
        issue['fields'][element] = {}
        issue['fields'][element]['value'] = value
        issue['fields'][element]['name'] = custom_fields[element]['name']
        issue['fields'][element]['id'] = custom_fields[element]['id']

    for element in issue['fields'].keys():
        [complete_field(custom_fields, element) for custom_field in custom_fields.keys() if custom_field == element]

    return issue
```

and 

```python
def filter_custom_fields(fields):

    custom_fields = {}

    _ = [field for field in fields if field['custom'] is True]

    for custom_field in _:
        custom_fields[custom_field['id']] = custom_field

    return custom_fields
```
What do you think?

sduenas: In general, I don't like the idea of changing the status of an object inside of a function. It can drive to future problems. I think more in something like:

```python

def map_custom_field(custom_fields, fields):
    build_cf = lambda cf, v: {'id': cf['id'], 'name': cf['name'], 'value' : v}

    return  {k : build_cf(custom_fields[k], v) \
             for k, v in fields.items() \
             if k in custom_fields}

mapping = map_custom_field(custom_fields, issue['fields'])
for k, v in mapping.items():
    issue['fields'][k] = v
```

Regarding the other function, don't use the underscore variable, this usually means you aren't going to use it.

albertinisg: Thanks!! I''ll try it out

albertinisg: Done in 328c547efb1a09060241c3cf3244083560283bb8

PR #55, opened 2016-09-02
[mediawiki] Use generic numeric format in datetime object

### LINKED ISSUE #54: [Mediawiki] Using --from-date on recent mediawiki versions results in a exception (UTC Format related)
Hey there,

I want to report a bug:

Was testing perceval on our wiki project which uses mediawiki 1.27. 

Passing the --from-data params with a timestamp crashes the script.

For example:

```
[2016-09-01 18:26:41,164 - requests.packages.urllib3.connectionpool - DEBUG] - "GET /api.php?list=allrevisions&format=json&arvnamespace=0&action=query&arvstart=2015-06-01T00%3A00%3A00%2B00%3A00&arvprop=ids&arvlimit=max&arvdir=newer HTTP/1.1" 200 None
  File "/home/crisbal/WikiToLearn/perceval/venv/lib/python3.5/site-packages/perceval/backends/mediawiki.py", line 494, in run
    for build in pages:
  File "/home/crisbal/WikiToLearn/perceval/venv/lib/python3.5/site-packages/perceval/backend.py", line 171, in decorator
    for data in func(self, *args, **kwargs):
  File "/home/crisbal/WikiToLearn/perceval/venv/lib/python3.5/site-packages/perceval/backends/mediawiki.py", line 110, in fetch
    for page_reviews in fetcher:
  File "/home/crisbal/WikiToLearn/perceval/venv/lib/python3.5/site-packages/perceval/backends/mediawiki.py", line 203, in __fetch_1_27
    pages_json = data_json['query']['allrevisions']
KeyError: 'query'

  File "/home/crisbal/WikiToLearn/perceval/venv/bin/perceval", line 176, in <module>
    main()
  File "/home/crisbal/WikiToLearn/perceval/venv/bin/perceval", line 97, in main
    cmd.run()
  File "/home/crisbal/WikiToLearn/perceval/venv/lib/python3.5/site-packages/perceval/backends/mediawiki.py", line 505, in run
    raise RuntimeError(str(e))
RuntimeError: 'query'
```

The problem is that newer mediawiki versions refuse to work with the formatted timestamp produced by python (https://phabricator.wikimedia.org/T144482)

It's an issue that mediawiki devs have to fix but we could hack comething to make it work.

albertinisg:
This PR fixes #54  

As @crisbal reported, isoformat() method from datetime library is not supported by Mediawiki API (http://meta.wikitolearn.org/api.php).

I've changed it to a numeric generic format and it worked as expected.

### DIFF
--- a/perceval/backends/mediawiki.py
+++ b/perceval/backends/mediawiki.py
@@ -428,7 +428,7 @@ class MediaWikiClient:
     def get_pages_from_allrevisions(self, namespaces, from_date=None, arvcontinue=None):
 
         if from_date:
-            from_date_str = from_date.isoformat()
+            from_date_str = from_date.strftime("%Y-%m-%d %H:%M:%S")
 
         params = {
             "action":"query",
--- a/perceval/backends/mediawiki.py
+++ b/perceval/backends/mediawiki.py
@@ -428,7 +428,10 @@ class MediaWikiClient:
     def get_pages_from_allrevisions(self, namespaces, from_date=None, arvcontinue=None):
 
         if from_date:
-            from_date_str = from_date.strftime("%Y-%m-%d %H:%M:%S")
+            if from_date.tzinfo != dateutil.tz.tzutc():
+                raise ValueError("Datetime is not in UTC timezone")
+
+            from_date_str = from_date.strftime("%Y-%m-%dT%H:%M:%SZ")
 
         params = {
             "action":"query",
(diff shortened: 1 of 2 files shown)


sduenas:
The solution of this is not that easy. On line 99, the backend converts the given datetime to UTC because most of our backends use UTC times to make the requests. If version 1.27 doesn't support timezones, the date shouldn't be converted.

You also need to be sure on which version this error happens. Does it happen in versions prior to 1.27? I guess not.

sduenas:
Reading Phabricator ticket again, we've just noticed that Mediawiki supports the ZULU time (a UTC timestamp with a Z at the end). So, the proposed fix needs to add this Z character at the end of the string and raise an error when the given time isn't in UTC.

albertinisg:
@sduenas tests completed!

PR #56, opened 2016-09-03
[mediawiki] Remove timezone in from_date

acs:
Remove timezone in from_date because MediaWiki >= 1.27 does not support it.


acs:
Already covered in #55

PR #61, opened 2016-09-18
Add support for creating Pypi packages

jgbarah:
Changes to setup.py for creating packages suitable to be uploaded to Pypi.
Version bumped to 0.3.0a1 (Alpha 1 to 0.3.0) to have a base version number compatible with PEP 440 to start with. (then bumped to Alpha 4, after some tests on the test Pypi archive).
Documentation to create the pacakges in docs/howto_pkg.rst.
Fix minor issue in docs/howto_doc.rst.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,18 +21,52 @@
 #     Santiago Dueñas <sduenas@bitergia.com>
 #
 
+# To use a consistent encoding
+import codecs
+from os import path
 import re
 
-from distutils.core import setup
+# Always prefer setuptools over distutils
+from setuptools import setup
 
+here = path.abspath(path.dirname(__file__))
+readme_md = path.join(here, 'README.md')
+version_py = path.join(here, 'perceval', '_version.py')
 
-with open('perceval/_version.py', 'r', encoding='utf-8') as fd:
+# Pypi wants the description to be in reStrcuturedText, but
+# we have it in Markdown. So, let's convert formats.
+# Set up thinkgs so that if pypandoc is not installed, it
+# just issues a warning.
+try:
+    import pypandoc
+    long_description = pypandoc.convert(readme_md, 'rst')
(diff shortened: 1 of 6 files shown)


setup.py
```diff
@@ -21,18 +21,51 @@
 #     Santiago Dueñas <sduenas@bitergia.com>
 #
 
+# Always prefer setuptools over distutils
```
sduenas: The import section should have the packages ordered as is described in [PEP8 document](https://www.python.org/dev/peps/pep-0008/#id23).

setup.py
```diff
@@ -21,18 +21,51 @@
...
+    with codecs.open(readme_md, encoding='utf-8') as f:
+        long_description = f.read()
 
-with open('perceval/_version.py', 'r', encoding='utf-8') as fd:
+with codecs.open(version_py, 'r', encoding='utf-8') as fd:
     version = re.search(r'^__version__\s*=\s*[\'"]([^\'"]*)[\'"]',
                         fd.read(), re.MULTILINE).group(1)
 
 setup(name="perceval",
+      description="Get data from software repositories",
```
sduenas: I suggest to change this by **"Fetch data from software repositories"**

setup.py
```diff
@@ -21,18 +21,51 @@
...
+      long_description=long_description,
+      url="https://github.com/GrimoireLab/perceval",
```
sduenas: The URL is https://github.com/grimoirelab/perceval

jgbarah:
I've pushed a new version of the commit, I guess that with all your concerns addressed.

WRT the doc about how to create Pypi packages, it tries to be about how to create the specific Pypi packages for Perceval, as a way of letting anyone understand how those are produced (the same way that we have doc about how to produce the documentation).

However, if you prefer to skip that, I can remove it.

PR #65, opened 2016-09-22
[mediawiki] Use pages API by default for MediaWiki >= 1.27

acs:
The use of the allrevisions API can be force using the new param --reviews-api when MediaWiki >= 1.27.

Updated tests for this new scenario.


sduenas:
Merged.

PR #66, opened 2016-09-26
Replace wrong backend name in usage msg: bugillzarest with bugzilla rest

sanacl:


sanacl:
I've rebased my second commit and squash-ed it to have a single commit.

PR #67, opened 2016-09-27
To work with Windows 10

shongoku:
I tried to do a Jenkins test but didn't work because the code made a path with a name with a bad carácter for Windows.
I've change this and now this work for Linux/Unix and Windows systems.
Besides, it must change the line to execute a Jenkins test for Windows:

python perceval jenkins http://jenkins.cyanogenmod.com/

without single quotation marks

Regards


shongoku:
About the last point, I've only changed on jenkins backend because I change a little text for other but at first I can do the change for all backends.
I'm goint to descard this pull request and then I'll correct all points.

PR #69, opened 2016-09-29
Remo backend using the API v2

acs:
ReMo has a new API more complete and that will be the final one. This backend uses it to gather events, activities/reports (new in v2) and users.

https://reps.mozilla.org/api/beta/

### DIFF
--- a/perceval/backends/remo2.py
+++ b/perceval/backends/remo2.py
@@ -20,17 +20,17 @@
 #     Alvaro del Castillo <acs@bitergia.com>
 #
 
+import functools
 import json
 import logging
 import os.path
 
 import requests
 
-from dateutil import parser
 
 from ..backend import Backend, BackendCommand, metadata
 from ..cache import Cache
-from ..errors import BackendError, CacheError, ParseError
+from ..errors import CacheError
 
 from ..utils import (DEFAULT_DATETIME,
                      datetime_to_utc,
@@ -41,6 +41,21 @@ from ..utils import (DEFAULT_DATETIME,
 logger = logging.getLogger(__name__)
 
 MOZILLA_REPS_URL = "https://reps.mozilla.org"
+REMO_DEFAULT_OFFSET = 0
+
+def remo_metadata(func):
+    """ReMo metadata decorator.
+
+    This decorator takes items overrides `metadata` decorator to add extra
+    information related to Kitsune (offset of the item).
+    """
+    @functools.wraps(func)
+    def decorator(self, *args, **kwargs):
+        for item in func(self, *args, **kwargs):
+            item['offset'] = item['data'].pop('offset')
+            yield item
+    return decorator
+
 
 class ReMo(Backend):
     """ReMo backend for Perceval.
@@ -68,13 +83,17 @@ class ReMo(Backend):
         self.client = ReMoClient(url)
         self.__users = {}  # internal users cache
 
+    @remo_metadata
     @metadata
-    def fetch(self, kind='events'):
+    def fetch(self, offset=REMO_DEFAULT_OFFSET, kind='events'):
         """Fetch events from the ReMo url.
 
         The method retrieves, from a ReMo url, the
         events.
 
+
+        :offset: obtain items after offset
+        :kind: kind of items to retrieve
         :returns: a generator of events
         """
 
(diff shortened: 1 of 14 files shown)


perceval/backends/remo2.py
```diff
@@ -0,0 +1,318 @@
...
+    @metadata
+    def fetch(self, kind='events'):
+        """Fetch events from the ReMo url.
+
+        The method retrieves, from a ReMo url, the
+        events.
+
+        :returns: a generator of events
+        """
+
```
sduenas: Remove this empty line

perceval/backends/remo2.py
```diff
@@ -0,0 +1,318 @@
...
+        supported_kinds = ['activities', 'events', 'users']
+
+        if kind not in supported_kinds:
+            raise RuntimeError('ReMo perceval backend does not support %s', kind)
```
sduenas: This statement should raise a `ValueError` exception instead of a `RuntimeError`. Moreover, I think you forgot to concatenate `kind` with the previous string.

perceval/backends/remo2.py
```diff
@@ -0,0 +1,318 @@
...
+            url = MOZILLA_REPS_URL
+        origin = origin if origin else url
+
+        super().__init__(origin, cache=cache)
+        self.url = url
+        self.client = ReMoClient(url)
+        self.__users = {}  # internal users cache
+
+    @metadata
+    def fetch(self, kind='events'):
```
sduenas: I'm not a big fan of `kind` name for this attribute. I think we should find a better name.

perceval/backends/remo2.py
```diff
@@ -0,0 +1,318 @@
...
+    @metadata
+    def fetch_from_cache(self):
+        """Fetch the items from the cache.
+
+        :returns: a generator of items
+
+        :raises CacheError: raised when an error occurs accessing the
+            cache
+        """
+
```
sduenas: Remove this empty line

perceval/backends/remo2.py
```diff
@@ -0,0 +1,318 @@
...
+        logger.info("Retrieving cached ReMo items: '%s'", self.url)
+
+        if not self.cache:
+            raise CacheError(cause="cache instance was not provided")
+
+        cache_items = self.cache.retrieve()
+
+        nitems = 0
+
+        while True:
```
sduenas: This infinite loop will make `next(cache_items)` to raise an exception when the iterator is exhausted.

perceval/backends/remo2.py
```diff
@@ -0,0 +1,318 @@
...
+        """Extracts the update time from a ReMo item.
+
+        The timestamp is extracted from 'end' field.
+        This date is converted to a perceval format using a float value.
+
+        :param item: item generated by the backend
+
+        :returns: a UNIX timestamp
+        """
+
```
sduenas: Remove this line

perceval/backends/remo2.py
```diff
@@ -0,0 +1,318 @@
...
+            # events updated field
+            updated = item['end']
+        elif 'date_joined_program' in item:
+            # users updated field that always appear
+            updated = item['date_joined_program']
+        elif 'report_date' in item:
+            # activities updated field
+            updated = item['report_date']
+        else:
+            raise RuntimeError("Can't find updated field for item %s", item)
```
sduenas: Don't use `RuntimeError` exception, use `ValueError` instead

perceval/backends/remo2.py
```diff
@@ -0,0 +1,318 @@
...
+
+    This class implements a simple client to retrieve events from
+    projects in a ReMo site.
+
+    :param url: URL of ReMo (sample https://reps.mozilla.org)
+
+    :raises HTTPError: when an error occurs doing the request
+    """
+
+    def __init__(self, url, limit=40):
```
sduenas: Define the value of `40` in a constant, do not hardcode it here.

perceval/backends/remo2.py
```diff
@@ -0,0 +1,318 @@
...
+        # 40 it the max limit for events.
+        self.url = url
+        self.limit = limit
+        self.api_activities_url = urljoin(self.url, '/api/beta/activities/')
```
sduenas: I think it would be better to have all these path in some constants.

perceval/backends/remo2.py
sduenas: I miss there the parameter `from_date`. It would break the API, I think. In fact, you should add it to make it possible to get items from a certain point in time.

perceval/backends/remo2.py
acs: Right! And the concatenate is because I was using the string inside logger. Fixed also!

perceval/backends/remo2.py
acs: Done!

perceval/backends/remo2.py
acs: I don't like kind neither. Instead of from_date. we must use offset. Added in the last version.

Alternatives to kind are type and class, but both are reserved keywords and we should use type_ or something like that, or class_ or klass. With this scenario, kind is not so bad. But I am open to alternatives!

perceval/backends/remo2.py
acs: done!

perceval/backends/remo2.py
acs: Fixed with:

        for item in cache_items:

A bit strange the issue is not detected with the tests ... but it is cleaner the for logic.

perceval/backends/remo2.py
acs: Done!

perceval/backends/remo2.py
acs: Done, and fixed also how the exception string is built.

perceval/backends/remo2.py
acs: This value is not needed anymore and I have removed if from code. Fixed!

perceval/backends/remo2.py
acs: Refactored as:

`self.api_activities_url = urljoin(self.url, ReMoClient.API_PATH+'/activities/')`

ReMoClient.API_PATH is probably the string that is going to change. Are you comfortable with that?

acs:
All review comments addressed and the new offset logic added with tests. Do you need something more for merging the PR? Thanks!!!!

sduenas:
Merged.

PR #70, opened 2016-10-03
[remo2] Change ReMo API URL for v2 to its final location: /api/remo/v1/

acs:
Probably the API is not called anymore as v2 so as soon as we discuss and fix all issues between the old v1 and v2 (new v1), remo2 will be renamed to remo and the old remo will be removed.


acs:
Upstream change:
https://github.com/mozilla/remo/commit/60d494225ff22c17eea9229fdf868874cfdb9302

acs:
Tests fixed!

PR #75, opened 2016-10-10
Added the tool setuptools

shongoku:
Added the package setuptools to install at beginning of installation.


sduenas:
I've been thinking on this a little more and now I think we shouldn't include this package in the requirements file. This file is also used by `pip` to install the dependencies of `perceval` package which does not require of `setuptools`.
Including information in the `README` file about how it is installed using either `pip` or `setup.py` script should be enough.

PR #78, opened 2016-10-18
[kitsune] Add error management when the kitsune sever fails

acs:
Add error management when the kitsune sever fails providing questions in a page

### DIFF
--- a/perceval/backends/kitsune.py
+++ b/perceval/backends/kitsune.py
@@ -98,6 +98,7 @@ class Kitsune(Backend):
 
         nquestions = 0  # number of questions processed
         tquestions = 0  # number of questions from API data
+        equestions = 0  # number of questions dropped by errors
 
         # Always get complete pages so the first item is always
         # the first one in the page
@@ -107,8 +108,30 @@ class Kitsune(Backend):
         drop_questions = offset - page_offset
         current_offset = offset
 
-        for raw_questions in self.client.get_questions(offset):
+        questions_page = self.client.get_questions(offset)
+
+        while True:
+            try:
+                raw_questions = next(questions_page)
+            except StopIteration:
+                break
+            except requests.exceptions.HTTPError as e:
+                # Continue with the next page if it is a 500 error
+                if e.response.status_code == 500:
+                    logger.exception(e)
+                    logger.error("Problem getting Kitsune questions. " +
+                                 "Loosing %i questions. Going to the next page.",
+                                 KitsuneClient.ITEMS_PER_PAGE)
+                    equestions += KitsuneClient.ITEMS_PER_PAGE
+                    current_offset += KitsuneClient.ITEMS_PER_PAGE
+                    questions_page = self.client.get_questions(current_offset)
+                    continue
+                else:
+                    # If it is another error just propagate the exception
+                    raise e
+
             self._push_cache_queue(raw_questions)
+
             try:
                 questions_data = json.loads(raw_questions)
                 tquestions = questions_data['count']
@@ -134,11 +157,13 @@ class Kitsune(Backend):
                 nquestions += 1
                 self._push_cache_queue('{}') # Mark with empty dict end of question
 
-            logger.debug("Questions: %i/%i", nquestions, tquestions)
+            logger.debug("Questions: %i/%i", nquestions + offset, tquestions)
 
             self._flush_cache_queue()
 
         logger.info("Total number of questions: %i (%i total)", nquestions, tquestions)
+        logger.info("Questions with errors dropped: %i", equestions)
+
 
     @kitsune_metadata
     @metadata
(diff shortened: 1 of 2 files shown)


perceval/backends/kitsune.py
```diff
@@ -107,8 +109,28 @@ def fetch(self, offset=DEFAULT_OFFSET):
...
+
+        while True:
+            try:
+                raw_questions = next(questions_page)
+            except StopIteration:
+                break
+            except requests.exceptions.HTTPError as e:
+                # Continue with the next page if it is a 500 error
+                if e.response.status_code == 500:
+                    traceback.print_exc()
```
sduenas: This should be printed to the log not to the error output.

perceval/backends/kitsune.py
```diff
@@ -107,8 +109,28 @@ def fetch(self, offset=DEFAULT_OFFSET):
...
+                    logging.error("Problem getting Kitsune questions. Loosing %i questions. Going to the next page.", KitsuneClient.ITEMS_PER_PAGE)
```
sduenas: Use `logger` object, not `logging`. You also have to split the line. It is too long.

tests/test_kitsune.py
```diff
@@ -211,6 +216,18 @@ def test_fetch_empty(self):
 
         self.assertEqual(len(questions), 0)
 
+    @httpretty.activate
```
sduenas: I don't understand this test. I know that it is going to fail when an `HTTPError` exception is raised unless it is a 500 error, that would continue without failing. Your tests should take into account that behaviour.

tests/test_kitsune.py
```diff
@@ -211,6 +216,18 @@ def test_fetch_empty(self):
...
+    def test_fetch_server_error(self):
+        """Test whether it works when the server fails"""
+
+        HTTPServer.routes(empty=True)
+
+        kitsune = Kitsune(KITSUNE_SERVER_URL)
+        offset = (KITSUNE_SERVER_FAIL_PAGE-1)*KITSUNE_ITEMS_PER_PAGE
```
sduenas: Add separation between operators.

tests/test_kitsune.py
acs: In this tests we access the page (69) which generates the 500 error that it is managed inside the backend, and no questions are returned. Then, the next page (70) is retrieved which does not exists and generates a 404 (not found) error which is captured by the assert. Then, the total number of pages is checked, which must be 0.

perceval/backends/kitsune.py
acs: Fixed!

tests/test_kitsune.py
acs: Fixed!

perceval/backends/kitsune.py
```diff
@@ -134,11 +157,13 @@ def fetch(self, offset=DEFAULT_OFFSET):
                 nquestions += 1
                 self._push_cache_queue('{}') # Mark with empty dict end of question
 
-            logger.debug("Questions: %i/%i", nquestions, tquestions)
+            logger.debug("Questions: %i/%i", nquestions+offset, tquestions)
```
sduenas: Put spaces between operators.

tests/test_kitsune.py
```diff
@@ -87,6 +89,12 @@ def request_callback(method, uri, headers):
                     body = mozilla_questions_2
                 else:
                     body = mozilla_question_answers_1
+            elif page == str(KITSUNE_SERVER_FAIL_PAGE):
+                # To tests for Internal Server Error
+                return (500, headers, '')
+            elif page == str(KITSUNE_SERVER_FAIL_PAGE+1):
```
sduenas: Spaces between operators.

tests/test_kitsune.py
```diff
@@ -211,6 +219,18 @@ def test_fetch_empty(self):
...
         self.assertEqual(len(questions), 0)
 
+    @httpretty.activate
+    def test_fetch_server_error(self):
+        """Test whether it works when the server fails"""
+
+        HTTPServer.routes(empty=True)
+
+        kitsune = Kitsune(KITSUNE_SERVER_URL)
+        offset = (KITSUNE_SERVER_FAIL_PAGE-1) * KITSUNE_ITEMS_PER_PAGE
```
sduenas: Same comment as above here. Put spaces between operators.

tests/test_kitsune.py
sduenas: This test is much better. The server will fail but the backend will continue fetching items.

perceval/backends/kitsune.py
acs: Fixed! Ups!

tests/test_kitsune.py
acs: Great!

sduenas:
Merged!

PR #79, opened 2016-10-18
[jenkins] Add blacklist support to jenkins backend

acs:


tests/test_jenkins.py
```diff
@@ -161,6 +162,24 @@ def test_fetch_empty(self):
 
         self.assertEqual(len(builds), 0)
 
+    @httpretty.activate
```
sduenas: For this test you can also add a check to test if that item was not retrieved. For instance, you can use `assertNotIn()` method with `builds` list.

tests/test_jenkins.py
```diff
@@ -161,6 +162,24 @@ def test_fetch_empty(self):
...
+        blacklist = [JENKINS_JOB_BUILDS_1]
+
+        self.__configure_http_server()
+
+        jenkins = Jenkins(JENKINS_SERVER_URL, blacklist_jobs=blacklist)
+        nrequests = len(requests_http)
+        builds = [build for build in jenkins.fetch()]
+        # No HTTP calls at all must be done for JENKINS_JOB_BUILDS_1
+        # Just the first call for all jobs and a second call for JENKINS_JOB_BUILDS_2
+        self.assertEqual(len(requests_http)-nrequests, 2)
```
sduenas: Let whitespaces between operators.

perceval/backends/jenkins.py
```diff
@@ -48,15 +48,17 @@ class Jenkins(Backend):
     :param url: Jenkins url
     :param tag: label used to mark the data
     :param cache: cache object to store raw data
+    :param blacklist_jobs: exclude the jobs of this list while fetching
     """
-    version = '0.3.0'
+    version = '0.3.1'
```
sduenas: The version should be `0.4.0` because you're adding a new feature and not fixing bugs.

tests/test_jenkins.py
acs: Fixed

perceval/backends/jenkins.py
acs: Fixed!

tests/test_jenkins.py
acs: But we avoid to get the job, not builds. Checking if we can do something like that because it is a nice improvement ... there is no job field in the builds, so we can not do this checking.

sduenas: Ok. I get it.

sduenas:
Merged!

PR #82, opened 2016-10-20
[remo2] Add remo2 to the list of supported backends in help and README.

acs:
Change API v2 to API v1 final as the name of the API used.


sduenas:
This PR is not needed anymore since we have moved the development of Mozilla backends to: https://github.com/grimoirelab/perceval-mozilla/

PR #85, opened 2016-10-24
Adapted to work with microsoft windows.

shongoku:
Some little changes in several files to adapt the path. With this path they work with Microsoft Windows.


sduenas:
The tests are failing. You have to rebase your commit because you are using an older version of the repository.

PR #86, opened 2016-10-25
Windows

shongoku:
Some little changes in several files to adapt the path. With this path they work with Microsoft Windows.


sduenas:
This PR has conflicts and contains a commit that it is not needed.

PR #87, opened 2016-10-26
Adapted the paths to work with Microsoft Windows too.

shongoku:
Before the files used the symbol "\" and other that this don't work with Microsoft. I've deleted this part to work with Posix and Windows systems.

### DIFF
--- a/perceval/backends/__init__.py
+++ b/perceval/backends/__init__.py
@@ -20,55 +20,66 @@
 #     Santiago Dueñas <sduenas@bitergia.com>
 #
 
-import importlib
-import os
-
-from ..backend import Backend, BackendCommand
-
-
-def register_backends(dirpath):
-    filenames = [fn for fn in os.listdir(dirpath)]
-
-    # Find candidate modules and remove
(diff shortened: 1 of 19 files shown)


sduenas:
I still think we need to find a better way to include this feature. I was thinking if we merge this patch we will have a different behaviour in Windows with the directories where the cache is stored. For instance, in Windows, a URI like `https://example.org/project/repository` will be stored in a directory of name `example.org/project/repository` while in Unix, the cache will be stored in the subdirectory `repository`, under `project` and `example.org`.
I think we should have the same behaviour for both platforms.

shongoku:
The question is if you have two operating system, why do you have  the same URI? The main is the result and both are OK.

jgbarah:
We now have a lot of conflicts here, and I'm not sure if you @sduenas came with a better solution for this... Any idea, anyone? Having this working in Windows would be rather interesting, from my point of view.

sduenas:
I'm going to close this PR because unfortunately, for technical reasons, GrimoireLab doesn't support Windows.

PR #101, opened 2016-12-14
[askbot] Fetch full list of comments

albertinisg:
Before, by parsing the HTML we were unable to retrieve the full list of
comments inside each element. Now we do a different request to fetch the
comments by a given object_id. The content displayed in the item will be
exactly the same as the returned one.

Backend version updated to 0.1.2


perceval/backends/core/askbot.py
```diff
@@ -44,7 +44,7 @@ class Askbot(Backend):
     :param url: Askbot site URL
     :param tag: label used to mark the data
     """
-    version = '0.1.1'
+    version = '0.1.2'
```
albertinisg: Sure, I'll update the commit

PR #106, opened 2017-01-09
fixed FSF address

celan:
Fixed the address of the Free Software Foundation in all occurrences. See here for the right address:
http://www.fsf.org/about/contact/


sduenas:
Thanks @celan. I didn't know how the old FSF address ended up in the license.

PR #117, opened 2017-01-27
[gerrit] Use 'number' instead 'id' as unique identifier for a review

acs:
Fix https://github.com/grimoirelab/perceval/issues/116


acs:
Increased version number @sduenas

sduenas:
Merged on 877be83

PR #120, opened 2017-02-23
[jenkins] Change the URL from which jobs are collected to retrieve all jobs

acs:
Right now we are collecting the jobs from the URL:

<jenkins_url>/view/All/api/json

but this view does not always exists, and it could not contain all the jobs.

Using directly:

<jenkins_url>/api/json

we get all the jobs from the jenkins server.

### DIFF
--- a/perceval/backends/core/jenkins.py
+++ b/perceval/backends/core/jenkins.py
@@ -51,7 +51,7 @@ class Jenkins(Backend):
     :param cache: cache object to store raw data
     :param blacklist_jobs: exclude the jobs of this list while fetching
     """
-    version = '0.5.1'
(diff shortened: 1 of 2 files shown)


perceval/backends/core/jenkins.py
```diff
@@ -199,7 +199,7 @@ def __init__(self, url, blacklist_jobs=None):
     def get_jobs(self):
         """ Retrieve all jobs
         """
-        url_jenkins = self.url + "/view/All/api/json"
+        url_jenkins = self.url + "/api/json"
```
sduenas: Use `urljoin` function from `perceval.utils` to merge the URL and the path.

PR #121, opened 2017-02-28
[github] Manage 404 not found error when getting the public github orgs for a login

acs:
Sometimes GitHub API when calling  https://api.github.com/users/joltup/orgs return
404 Client Error: Not Found for url: https://api.github.com/users/joltup/orgs. But
if you try to call manually this url, it works and returns an empty array in
the cases detected.

This is an error that must be managed inside the backend because the user of the
perceval backend can not manage it and continue with the rest of the logic for
extracting a raw github item.

This orgs information is not critical and it is filled with an empty array if the
404 not found exception is raised, the normal response in these cases from the
GitHub API when it does not fail.

### DIFF
--- a/perceval/backends/core/github.py
+++ b/perceval/backends/core/github.py
@@ -14,7 +14,7 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program; if not, write to the Free Software
-# Foundation, 51 Franklin Street, Fifth Floor, Boston, MA 02110-1335, USA. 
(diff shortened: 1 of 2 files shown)


acs:
https://github.com/grimoirelab/perceval/blob/master/perceval/backends/core/redmine.py#L171

We need to just cover the 404 case. I will upload the PR @sduenas . Thanks for the review!

acs:
@sduenas updated the PR and this time it also includes testing.

PR #127, opened 2017-03-16
 Add retry for ssh commands to support temporal gerrit server errors

acs:
In gerrit servers when there a huge number of reviews (>300K) the ssh collect process is too demanding for the remote server which can not cope with the petitions and the remove server fails.

With this commit, if the remote server fails, the backend waits for 60s and retries the query. And the 120s and 180s. If there are 3 errors, the backend raises an exception.

So the backend now is more robust against server failures.


acs:
@sduenas updated the PR with a clean version with just one commit and no merges from master!

PR #134, opened 2017-04-03
[gerrit] Add option to avoid the remote ssh host check in gerrit commands based on ssh

acs:
If --ssh-host-check-no is provided the to every ssh command the option
-o StrictHostKeyChecking=no is added to there is no need to have already
stored in the ssh local config the identity of the remote server.


acs:
Good suggestion @sduenas . Changed the name of the param to disable-host-key-check

PR #137, opened 2017-04-21
[logs] Add new log messages

mghfdez:
After a first revision of the Git backend module, a few new log messages has been added:

- It may be useful to distinguish between parsing a given git log file or it has to obtain that log file from a repository.
`
logger.info("Fetching commits from the given log file %s", self.gitpath)
logger.info("Log file not specified, retrieving from repository.")
`

- As proposed, now a message informs if the repository is found locally or it has to be cloned first.
`
logger.info("Git repository not found in %s", self.gitpath)
logger.info("Found Git repository in %s", self.gitpath)
`

- Message    `logger.debug("Git %s repository cloned into %s",...` has been changed into 'info' type, as this information may help to discriminate if the program has stuck in the cloning phase.

### DIFF
--- a/perceval/backends/core/git.py
+++ b/perceval/backends/core/git.py
@@ -125,18 +125,21 @@ class Git(Backend):
 
     def __fetch_and_parse_log(self, from_date, branches):
         if os.path.isfile(self.gitpath):
+            logger.info("Fetching commits from the given log file %s", self.gitpath)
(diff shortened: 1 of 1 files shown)


jgbarah:
I'm sorry this patch is now conflicting. I'm sorry this was sitting here for so long, but if you still think it is relevant, would you mind updtaing it to the latest version in the master branch?

PR #138, opened 2017-04-29
Add vBulletin support

jeremy-lq:
While not OSS itself, many open source communities including LQ, Fedora, and Ubuntu currently use vBulletin. This PR adds initial support and has been tested with versions 3 and 4.


acs:
HI @jeremy-lq 

Thank you very much for your contribution. Give us this week to evaluate it and provide feedback but it seems to be pretty cool. An interesting data sources and including tests in the implementation. Stay tuned!

sduenas:
@jeremy-lq I couldn't make it work with Ferdora's forum nor with vBulletin
```
$ perceval vbulletin http://forums.fedoraforum.org
RuntimeError: 404 Client Error: Not Found for url: http://forums.fedoraforum.org/api.php?api_m=api_init&clientversion=0.1.0&clientname=Perceval+vBulletin+client&uniqueid=7789472b-9059-41a6-9f15-b7e1896e8b80&platformname=Linux&platformversion=4.4.0-70-generic
```

Could you provide an example of usage?

jeremy-lq:
The mobile API must be installed for this to work correctly. You'll also need to pass in a valid API key with the -t flag.

sduenas:
I couldn't find any place where I cat get a token to test this (I've tried with Ubuntu forums and with the vBulletin forum itself). Any idea?

jeremy-lq:
The API key isn't something publicly available. The quickest way to test would likely be to have a local copy of vbulletin installed with the mobile API configured.

sduenas:
Sorry but I'm afraid we can't merge this on master branch. I have no time to install a vBulletin instance to test this PR. I close it but I will try to reopen it as soon as I have some free time.

Anyway, thanks for your contribution.

PR #145, opened 2017-07-04
Add snap functionality

gsilvapt:
Added snap functionality for this package, published here: https://dashboard.snapcraft.io/dev/snaps/7712/

Snaps aim to have apps working in any Linux distribution, anywhere, from desktop devices to IoT. They are quick to install, safe to run and you can run updates immediately. Snaps are also self-contained as they are read-only images and they bundle everything it needs to run the app. There's also a possibility to run beta features with the channel and developer mode flags. These do not affect the rest of the system either.
You can find everything about snaps in https://snapcraft.io/

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,22 @@
+name: perceval-snap 
+version: '0.8.0' 
+summary: Retrieve and gather data from software repositories, 
+description: |
+  Package that gathers and retrieves data from known software repositories.
+  Repository sources includes askbot, bugzilla, bugzillarest, confluence, 
+  discourse, dockerhub, gerrit, git, github, gname, hyperkitty, jenkins, jira, 
+  mbox, mediawiki, meetup, nntp, phabricator, pipermail, redmine, rss, slack, 
+  stackexchange, supybot, telegram.
+grade: stable 
+confinement: strict 
+
+parts:
+  perceval:
+    source: https://github.com/grimoirelab/perceval.git
+    source-tag: 0.8.0
+    plugin: python
+    requirements: requirements.txt
+apps:
+  perceval:
+    command: perceval
+    plugs: [network]
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -12,7 +12,7 @@ confinement: strict
 
 parts:
   perceval:
-    source: https://github.com/grimoirelab/perceval.git
+    source: .
     plugin: python
     requirements: requirements.txt
 apps:
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,4 +1,4 @@
-name: perceval-snap 
+name: perceval
 version: '0.8.0' 
 summary: Retrieve and gather data from software repositories, 
 description: |
@@ -13,7 +13,6 @@ confinement: strict
 parts:
   perceval:
     source: https://github.com/grimoirelab/perceval.git
-    source-tag: 0.8.0
     plugin: python
     requirements: requirements.txt
 apps:


snap/snapcraft.yaml
```diff
@@ -0,0 +1,22 @@
...
+  discourse, dockerhub, gerrit, git, github, gname, hyperkitty, jenkins, jira, 
+  mbox, mediawiki, meetup, nntp, phabricator, pipermail, redmine, rss, slack, 
+  stackexchange, supybot, telegram.
+grade: stable 
+confinement: strict 
+
+parts:
+  perceval:
+    source: https://github.com/grimoirelab/perceval.git
+    source-tag: 0.8.0
```
elopio: Change this to be `source: .` and remove the tag. That way it's easier to enable continuous delivery.

snap/snapcraft.yaml
```diff
@@ -0,0 +1,22 @@
+name: perceval-snap
```
elopio: rename this to just `perceval`.

snap/snapcraft.yaml
```diff
@@ -0,0 +1,21 @@
...
+  Repository sources includes askbot, bugzilla, bugzillarest, confluence, 
+  discourse, dockerhub, gerrit, git, github, gname, hyperkitty, jenkins, jira, 
+  mbox, mediawiki, meetup, nntp, phabricator, pipermail, redmine, rss, slack, 
+  stackexchange, supybot, telegram.
+grade: stable 
+confinement: strict 
+
+parts:
+  perceval:
+    source: https://github.com/grimoirelab/perceval.git
```
elopio: When the snapcraft.yaml is upstream, there is no need to specify the full path to the repo, because you are already in a directory with the source downloaded. Please replace it with:

`source: .`

jgbarah:
I'm not familiar with snaps, but by looking at the YAML file, it seems to me:

* It is using the old requirements.txt file. Now, dependencies are in `setup.py`. Not sure how to get them from there...
* In general, the `setup.py` file has all the information to build the package, including the `perceval` command. Wouldn't it better to use it for the snap? Or that is not possible?

gsilvapt:
I have used the Python plugin to snap this application. It is required to tell in the yaml where the requirements are as it is the most used method.

@elopio Is there a way to get the requirements from the setup.py into the yaml file?

PS: When I tested the final snap, it seemed to work fine.

jgbarah:
Thanks for the answer. I guess maybe @sduenas is maintaining proper dependencies in `requirements.txt`. I would be more comfortable using dependencies in `setup.py` (and using it for installation, as well), since it is the one we maintain for pip packages, but I'm not sure that's possible.

elopio:
If the dependencies are all in setup.py, you can just drop the requirements keyword from the yaml.

About the script declared in setup.py, you still have to declare it in the snapcraft.yaml because snaps require extra metadata that python doesn't have. In this case, `plugs`, which are the permissions that the snap require outside of its confined space. network means that it can make web requests.

jgbarah:
Thanks! @gsilvapt would mind then producing a new version of the pr not using requirements, as suggested? In addition, how could I better test it in a Debian box?

gsilvapt:
Is it okay to show the previous commits? Some projects don't mind, others care... So I'm asking because I would need help taking care of the previous commis.

Regarding your question, I think [this link](https://snapcraft.io/docs/build-snaps/build-on-lxd-docker) might help you on that topic. That is from our official documentation and it is about how to build snaps on containers like LXD and Docker. Hope it helps!

elopio:
@jgbarah you can also just go to https://build.snapcraft.io and enable the automatic continuous delivery from there. Every change on master will result in a new snap in the edge channel. The users will be able to find searching the store only snaps that are in the stable channel. So edge is perfect for experimenting, gathering feedback, checking if the snap will actually provide good value to your users.

abitrolly:
I'd like to use `perceval` from snap. What is the progress on this?

gsilvapt:
hi!

The snap was published here at the time: https://dashboard.snapcraft.io/snaps/perceval/.

I'm not sure if it works though, and it seems it doesn't. This PR was not accepted at the time and the snap funcionality is thus not available. Perhaps we can both work to put the snap back together a re-ping this PR? At the time, @elopio also made some suggestions to the yaml file I believe I didn't complete back then.

abitrolly:
I don't mind helping with this PR. If you add me as collaborator I can at least address @elopio comments and push some releases for testing.

gsilvapt:
Whoa, whoa, I 'm not a collaborator of this project :P What you can do though is to fork the repository and create the snap. After committing && pushing that, you can submit a new PR to replace this one. 

The comments of Leo I meant are in this thread (for guidance).

sduenas:
This PR is old and we aren't interested in support snap right now.

PR #151, opened 2017-09-05
[discourse] Do not process topics with last_posted_at = None

acs:
This error is described at 

https://github.com/grimoirelab/perceval/issues/150

and it can be reproduced with:

`perceval discourse https://discourse.mozilla.org/`

This PR fixes it.

### DIFF
--- a/perceval/backends/core/discourse.py
+++ b/perceval/backends/core/discourse.py
@@ -234,6 +234,9 @@ class Discourse(Backend):
 
         for topic in topics_page['topic_list']['topics']:
             topic_id = topic['id']
+            if topic['last_posted_at'] is None:
+                logger.warning("Topic %s with last_posted_at null. Ignoring it.", topic['title'])
+                continue
             updated_at = str_to_datetime(topic['last_posted_at'])
             pinned = topic['pinned']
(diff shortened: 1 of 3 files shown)


perceval/backends/core/discourse.py
```diff
@@ -234,6 +234,9 @@ def __parse_topics_page(self, raw_json):
 
         for topic in topics_page['topic_list']['topics']:
             topic_id = topic['id']
+            # a topic without posts? - acs
```
sduenas: What does this comment mean?

sduenas: Before merging this, we need some unit tests.

acs: About the comment, I am not totally sure why the topic does not have a last updated date, so my guess is that it has no posts yet. This is why I have left this comment.

I will work in the unit tests, yes!

acs:
@sduenas I have removed the comment because there are posts in this topic with the null value for last_updated_at. So I have decided to log a warning about this fact.

acs:
This issue is waiting for this fix:

https://github.com/mozilla/participation-metrics-org/issues/123

Issue #53, opened 2016-08-18
[github] Get information extra information about issues and pull requests


jsmanrique:
It would be interesting having the following information from GitHub issues and pull requests:
- Comments (who, what, how many, etc.)
- Labels

jsmanrique:
Any insights about how this can be done?

sduenas:
@acs any idea about when we can include this feature?

acs:
@jsmanrique @sduenas we just need to prioritize it. It should not be that hard to implement.

jsmanrique:
ASAP?

acs:
Ok, in the kitchen now waiting for its turn. Stay tuned.

Labels are already included now:

https://github.com/grimoirelab/GrimoireELK/blob/master/tests/data/github.json#L184

jsmanrique:
@acs, when labels info is added?

Still missing comments authors info I think. Correct me if I am wrong

acs:
@jsmanrique comments are not included yet in the enriched index. 

About the labels, taking a look to the github enricher:

https://github.com/grimoirelab/GrimoireELK/blob/master/grimoire_elk/elk/github.py#L323

The labels are included in this field.

jsmanrique:
AFAIK the number of 'comments' are part of raw data. I suppose each time you ask for the raw data, you get the latest number of comments for each issue.

I think we are mixing raw data extraction and data enrichment.

acs:
I am always talking about the enriched data. In the raw data I am not sure if the comments for each issue/pr are included, or we need to add an extra step to collect this information and include it in the raw data.

jsmanrique:
Yes, that's something to discuss, but it might be interesting, at least, to know:
* people opening issues/pr (already included)
* people commenting issues/pr (not done yet)
* number of comments by each person in each issue/pr

valeriocos:
Hi @acs @jsmanrique @sduenas 
GitHub includes different kinds of comment somehow related to an issue. The way to fetch them (all together or separately) is something that should be discussed.

For instance if you look at #120, you will see that:
issue comments are at https://api.github.com/repos/grimoirelab/perceval/issues/120/comments
review comments are at https://api.github.com/repos/grimoirelab/perceval/pulls/120/reviews
commit comments are at https://api.github.com/repos/grimoirelab/perceval/pulls/120/comments

valeriocos:
The github backend is now able to fetch issue comment data (see PR #152)

[...]

PR #152, opened 2017-09-07
[github] Get GitHub issue comments

valeriocos:
This pull request contains code to fetch issue comments (see #53).


perceval/backends/core/github.py
```diff
@@ -85,6 +86,31 @@ def __init__(self, owner=None, repository=None,
...
+            self._push_cache_queue('{COMMENTS}')
+            self._push_cache_queue(raw_comments)
+            self._flush_cache_queue()
+            for comment in json.loads(raw_comments):
+                comment["user_data"] = self.__get_user(comment['user']['login'])
+                comments.append(comment)
+
+        return comments
+
+    def __get_issue_assignees(self, raw_assignees):
```
sduenas: This should be removed and included in another PR. It's a new feature not related with this PR.

perceval/backends/core/github.py
```diff
@@ -85,6 +86,31 @@ def __init__(self, owner=None, repository=None,
                                    sleep_for_rate, min_rate_to_sleep)
         self._users = {}  # internal users cache
 
+    def __get_issue_comments(self, issue_number):
+        """ Get issue comments """
+        comments = []
```
sduenas: Leave a blank line between the docstring and the code.

perceval/backends/core/github.py
```diff
@@ -120,83 +147,112 @@ def fetch(self, from_date=DEFAULT_DATETIME):
...
 
         cache_items = self.cache.retrieve()
-
+        current_issue = None
         issues = None
 
         while True:
             try:
                 raw_item = next(cache_items)
+
```
sduenas: I think you should split this work in several and small methods. It would be easy to read and to understand. One method for processing users, one for assignees and so on.

perceval/backends/core/github.py
```diff
@@ -318,17 +378,16 @@ def __send_request(self, url, params=None, headers=None):
         logger.debug("Rate limit: %s" % (self.rate_limit))
         return r
 
-    def get_issues(self, start=None):
+    def get_paginated_items(self, path, payload_type, start=None):
```
sduenas: I like the idea of this method but in my opinion the method for fetching issues should not be removed. We can have one private method called `_fetch_items` and several methods like `issues` or `comments` that calling to `fetch_items` will download the data. For instance, see how it's implemented in `meetup` backend: https://github.com/grimoirelab/perceval/blob/master/perceval/backends/core/meetup.py#L375

sduenas: In any case, this functionality should be included in a separated commit.

perceval/backends/core/github.py
```diff
@@ -85,6 +86,32 @@ def __init__(self, owner=None, repository=None,
...
+            self._push_cache_queue('{COMMENTS}')
+            self._push_cache_queue(raw_comments)
+            self._flush_cache_queue()
+            for comment in json.loads(raw_comments):
+                comment["user_data"] = self.__get_user(comment['user']['login'])
+                comments.append(comment)
+
+        return comments
+
+    def __get_issue_assignee(self, raw_assignee):
```
sduenas: As I said before, you should remove this feature.

perceval/backends/core/github.py
```diff
@@ -120,83 +148,133 @@ def fetch(self, from_date=DEFAULT_DATETIME):
...
             for issue in issues:
-                for field in ['user', 'assignee']:
+                for field in TARGET_ISSUE_FIELDS:
+                    issue[field + "_data"] = {}
+
                     if issue[field]:
-                        issue[field + "_data"] = self.__get_user(issue[field]['login'])
-                    else:
-                        issue[field + "_data"] = {}
+                        if field == "user":
```
sduenas: For this types of strings, please use single quotes. We usually use them. You should change them in other places of this patch.

perceval/backends/core/github.py
```diff
@@ -120,83 +148,133 @@ def fetch(self, from_date=DEFAULT_DATETIME):
...
+                            issue[field + "_data"] = self.__get_user(issue[field]['login'])
+                        elif field == "assignee":
+                            issue[field + "_data"] = self.__get_issue_assignee(issue[field])
+                        elif field == "comments":
+                            issue[field + "_data"] = self.__get_issue_comments(issue["number"])
+                self._push_cache_queue('{}')
+                self._flush_cache_queue()
                 yield issue
 
+    def __fetch_issues_from_cache(self, cache_items):
```
sduenas: Private methods should be after public methods.

perceval/backends/core/github.py
```diff
@@ -120,83 +148,133 @@ def fetch(self, from_date=DEFAULT_DATETIME):
...
+        """Fetch issues from cache"""
+
+        raw_content = next(cache_items)
+        issues = (i for i in json.loads(raw_content))
```
sduenas: Why a tuple and not a list?

perceval/backends/core/github.py
```diff
@@ -120,83 +148,133 @@ def fetch(self, from_date=DEFAULT_DATETIME):
...
             try:
                 raw_item = next(cache_items)
+
+                if raw_item == '{ISSUES}':
+                    issues = self.__fetch_issues_from_cache(cache_items)
+
+                    if issues:
+                        current_issue = next(issues)
+                        self.__init_extra_issue_fields(current_issue)
+
```
sduenas: Remove blank lines between `if/elif/else` statements. It improves the reading of the code.

perceval/backends/core/github.py
```diff
@@ -120,83 +148,133 @@ def fetch(self, from_date=DEFAULT_DATETIME):
...
+                elif raw_item == '{}':
+                    yield current_issue
+                    current_issue = next(issues)
+                    self.__init_extra_issue_fields(current_issue)
+
             except StopIteration:
                 if issues:
-                    for issue in self.__build_issues(issues):
-                        yield issue
+                    (i for i in issues)
```
sduenas: What is the rationale behind this?

perceval/backends/core/github.py
valeriocos: the code in the master in branch already dealt with issue assignee. I removed the feature issue_assignees

perceval/backends/core/github.py
valeriocos:
```python
(i for i in json.loads(raw_content))
``` 
should be the pythonic way to return a generator

perceval/backends/core/github.py
valeriocos: My mistake, this part can be replaced by 
````python
except StopIteration:
    break
````

File: perceval/backends/core/github.py (hunk 5 above)
Comment (sduenas) [reply perceval/backends/core/github.py]: Ok! I forgot about that.

File: perceval/backends/core/github.py [hunk 11]
```diff
@@ -120,83 +102,174 @@ def fetch(self, from_date=DEFAULT_DATETIME):
...
+        """ Get issue assignee """
+
+        self._push_cache_queue('{ASSIGNEE}')
+        self._push_cache_queue(raw_assignee)
+        self._flush_cache_queue()
+        assignee = self.__get_user(raw_assignee['login'])
+
+        return assignee
+
+    def __get_user(self, login):
```
sduenas: Has this method change? There is a cache in memory to avoid many trips to the server when requesting for users. You can find the dict in `self._users`. This method should check that dict first.

valeriocos: No it hasn't. This method calls
 ```python 
self.client.get_user(login)
```
 which takes care of caching the users

sduenas: Ok!

valeriocos:
:tada: :tada: :tada:

PR #170, opened 2017-09-28
[github] Minor fixes backend

valeriocos:
This pull request fixes:
- The initialization of issue data attributes. Before all of them were set to empty dicts, now comments, mentions and assignees data attributes are set to empty lists
- Improve the tag naming for caching
- Move the public methods has_resuming, has_caching, metadata_id, metadata_updated_on and metadata_category before the private methods


perceval/backends/core/github.py
```diff
@@ -388,9 +435,9 @@ def __init_extra_issue_fields(self, issue):
 
         issue['user_data'] = {}
         issue['assignee_data'] = {}
-        issue['assignees_data'] = {}
-        issue['comments_data'] = {}
-        issue['reactions_data'] = {}
+        issue['assignees_data'] = []
```
sduenas: Not sure whether to initialize all these keys is better to use `setdefaults` rather than do it manually. What do you think?

valeriocos: Probably *setdefaults* is a cleaner solution

perceval/backends/core/github.py
```diff
@@ -70,7 +70,7 @@ class GitHub(Backend):
     :param min_rate_to_sleep: minimun rate needed to sleep until
          it will be reset
     """
-    version = '0.10.2'
+    version = '0.10.3'
```
sduenas: The version should be updated to 0.11. The API has changed.

PR #171, opened 2017-10-06
__version__ is not part of the module perceval

iht:
The perceval bin script is failing on my system, because it cannot find __version__ in the perceval module. This change fixes it, perceval can be used as a command line utility again

### DIFF
--- a/bin/perceval
+++ b/bin/perceval
@@ -79,7 +79,7 @@ PERCEVAL_EPILOG_MSG = \
 """Run '%(prog)s <backend> --help' to get information about a specific backend."""
 
 PERCEVAL_VERSION_MSG = \
-"""%(prog)s """  + perceval.__version__
+"""%(prog)s """  + perceval._version.__version__
 
 
 # Logging formats


sduenas:
You have something wrong in your installation. This is not required. Perceval script imports perceval package. This package has access to version because it imports `__version__` symbol using `__all__` variable.

https://github.com/grimoirelab/perceval/blob/master/perceval/__init__.py#L28

jgbarah:
I can confirm this happens to me as well, and I have no evidence of my installation being wrong... I'm reopening, just in case, while we check what's happening exactly. In fact, in the same installation, perceval-0.9.2 shows this behavior, while perceval-0.9.0 works like a charm... :-(

jgbarah:
A more detailed description of what happens:

```
(ivenv) $ perceval --help
  File "/tmp/ivenv/bin/perceval", line 83, in <module>
    """%(prog)s """  + perceval.__version__
AttributeError: module 'perceval' has no attribute '__version__'
(ivenv) $ python
Python 3.5.4 (default, Aug 12 2017, 14:08:14) 
[GCC 7.1.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import perceval
>>> perceval.__version__
  File "<stdin>", line 1, in <module>
AttributeError: module 'perceval' has no attribute '__version__'
>>> 

```

jgbarah:
OK, I've tracked what happens. It does not have to do with perceval itself, but with the interaction between the perceval and perceval-opnfv (and maybe other perceval-*) pip packages.

If I pip install perceval 0.9.2 in a fresh environment, it works perfectly. If I look in its installation, I see the correct perceval/__init__.py file. But if after that I pip install perceval-opnfv, that file gets changed to an empty one. That's the reason why __version__ can no longer be tracked.

I'm opening an issue at perceval-opnfv and closing this one. I think the problem lies in how the packaging is done.

jgbarah:
Please, check new versions of perceval-opnfv and perceval-mozilla, which will be produced likely tomorrow, after grimoirelab/perceval-opnfv#6 and grimoirelab/perceval-mozilla#4 are merged.

Reopening just in case you can confirm (or not) that this fix works.

sduenas:
This was already fixed.

PR #172, opened 2017-10-06
[github] Add  pull request data

valeriocos:
This patch enables the GitHub backend to fetch pull request data. Pull requests can be fetched setting in the fetch method the category 'pull request'. The pull request data consists of:
    - review comments
    - requested reviewers
    - reactions on review comments
    - data about the user merging the pull request
    - data about the user submitting the pull request.


sduenas:
We're not going to add this until we define how to manage in perceval issues and PR categories. I'm closing it but it will be reopened as soon as possible.

sduenas:
Since we have `categories` in Perceval, I think we should try to add PRs support using a category for them. What do you think @valeriocos? This means to addapt this PR to the new implementation of `github` backend.

valeriocos:
That's indeed a good idea!

PR #190, opened 2017-10-25
Fix Perceval version

nlamirault:
I install Perceval from pip. Then : 

```
$ perceval --version
  File "/home/nlamirault/Apps/grimoirelab/bin/perceval", line 83, in <module>
    """%(prog)s """  + perceval.__version__
AttributeError: module 'perceval' has no attribute '__version__'
```

With this patch :

```
$ perceval --version
perceval 0.9.2
```

### DIFF
--- a/bin/perceval
+++ b/bin/perceval
@@ -29,7 +29,7 @@ import sys
 
 import perceval
 import perceval.backends
-
+from perceval import _version
 
 PERCEVAL_USAGE_MSG = \
 """%(prog)s [-c <file>] [-g] <backend> [<args>] | --help | --version"""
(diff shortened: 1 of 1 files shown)


jgbarah:
This happened when you installed perceval from pip, or when you installed something from pip that included perceval as a dependency?

I'm asking because we have another report, #171, quite similar. We have been working with it, and it seems that it happens when, in addition to perceval, perceval-opnfv and/or perceval-mozilla were installed. Buth are pulled in if you install, for example, grimoire-mordred or grimoire-elk. In this case, mordred or p2o.py work, because perceval as a module is working, but the `perceval` command does not, failing as you mention.

If this were the case, we think we're fixing it with new versions of perceval-opnfv and perceval-mozilla, which will be produced likely tomorrow, after grimoirelab/perceval-opnfv#6 and grimoirelab/perceval-mozilla#4 are merged.

If this were not the case, we would need to revisit this. So, leaving the pr open until you can confirm it is what I explain above, and if possible that you confirm the above fix fixes the problem for you (or not, and we need to drill deeper).

PR #192, opened 2017-10-31
[github] Add GitHub Enterprise support

dpose:
So far, GitHub backend didn't support custom API url for GitHub Enterprise.
Now, the API url can be set as a parameter (--enterprise-api_url) when
the backend is run.


sduenas:
What's the purpose of this PR? Why don't you use `enterprise-url` parameter?

PR #193, opened 2017-11-03
Fix GitHub Enterprise support

dpose:
The new structure for the API url is 'https://git.example.com/api/v3'


sduenas:
Thanks! Merged!

PR #194, opened 2017-11-03
[archive] Add archive manager to simplify retrieval of cached data

valeriocos:
This PR simplifies the retrieval of cached data by delegating its management to an archive manager.

In a nutshell, the data returned during fetch executions is stored in different archives (one per execution) handled by an archive manager. Such a manager allows to transparently play back previous fetch executions, enabling the recovery of the stored data as it was originally collected from the data source.


valeriocos:
Another PR targets this new feature (https://github.com/grimoirelab/perceval/pull/253)

PR #208, opened 2017-11-15
[client] Add generic client

valeriocos:
This PR proposes a generic client to deal with common problems (server errors, rate limit bans, connection losts) when fetching data from APIs.
The generic client has been successfully applied to the GitHub backend

PR #212, opened 2017-11-22
[client] Add generic client (v2)

valeriocos:
This PR proposes a generic client (and replaces PR #208)  to deal with common problems (server errors, rate limit bans, connection losts) when fetching data from APIs.
The generic client has been successfully applied to the GitHub backend.

### DIFF
--- a/perceval/client.py
+++ b/perceval/client.py
@@ -1,0 +1,236 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2015-2017 Bitergia
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, 51 Franklin Street, Fifth Floor, Boston, MA 02110-1335, USA.
+#
+# Authors:
+#     Valerio Cosentino <valcos@bitergia.com>
+#
+
+import logging
+import time
+
+import requests
+from requests.adapters import HTTPAdapter
+from urllib3.util import Retry
+
+from .errors import RateLimitError
+
+logger = logging.getLogger(__name__)
+
+
+class HttpClient:
+    """Abstract class for HTTP clients.
+
+    Base class to query data sources taking care of retrying requests in case connection issues. If the data source does
+    not send back a response after retrying a request, a RetryError exception is thrown.
+
+    Sub-classes can use the methods fetch and _fetch to obtain data from the data source.
+
+    To track which version of the client was used during the fetching
+    process, this class provides a `version` attribute that each client
+    may override.
+
+    :param base_url: base URL of the data source
+    :param max_retries: number of max retries to a data source before raising a RetryError exception
+    :param max_retries_on_connect: max number of retries on connection error
+    :param max_retries_on_read: max number of retries on read errors
+    :param max_retries_on_redirect: max number of retries on redirects
+    :param max_retries_on_status: max number of retries on selected error status codes
+    :param status_forcelist: list of status codes where the retry attempts happen
+    :param method_whitelist: list of uppercased HTTP method verbs where the retry happen
+    :param raise_on_redirect: try retries on redirect status codes
+    :param raise_on_status: try retries on selected status codes
+    :param respect_retry_after_header: try retries for status code with retry-after headers
+    :param default_sleep_time: default time to sleep in case of connection problems
+    :param headers: list of session headers
+    """
+    version = '0.1'
+
+    DEFAULT_SLEEP_TIME = 1
+
+    MAX_RETRIES = 5
+    MAX_RETRIES_ON_CONNECT = 5
+    MAX_RETRIES_ON_READ = 5
+    MAX_RETRIES_ON_REDIRECT = 5
+    MAX_RETRIES_ON_STATUS = 5
+
+    DEFAULT_METHOD_WHITELIST = False
+    DEFAULT_RAISE_ON_REDIRECT = True
+    DEFAULT_RAISE_ON_STATUS = True
+    DEFAULT_RESPECT_RETRY_AFTER_HEADER = True
+
+    DEFAULT_RETRY_AFTER_STATUS_CODES = [413, 429, 503]
+    DEFAULT_STATUS_FORCE_LIST = [408, 423, 504]
+
+    GET = "GET"
+    POST = "POST"
+
+    def __init__(self, base_url, max_retries=MAX_RETRIES, max_retries_on_connect=MAX_RETRIES_ON_CONNECT,
+                 max_retries_on_read=MAX_RETRIES_ON_READ, max_retries_on_redirect=MAX_RETRIES_ON_REDIRECT,
+                 max_retries_on_status=MAX_RETRIES_ON_STATUS, status_forcelist=DEFAULT_STATUS_FORCE_LIST,
+                 method_whitelist=DEFAULT_METHOD_WHITELIST, raise_on_redirect=DEFAULT_RAISE_ON_REDIRECT,
+                 raise_on_status=DEFAULT_RAISE_ON_STATUS, respect_retry_after_header=DEFAULT_RESPECT_RETRY_AFTER_HEADER,
+                 default_sleep_time=DEFAULT_SLEEP_TIME, headers=None):
+        self.base_url = base_url
+
+        self.max_retries = max_retries
+        self.max_retries_on_connect = max_retries_on_connect
+        self.max_retries_on_read = max_retries_on_read
+        self.max_retries_on_redirect = max_retries_on_redirect
+        self.max_retries_on_status = max_retries_on_status
+        self.status_forcelist = status_forcelist
+        self.method_whitelist = method_whitelist
+        self.raise_on_redirect = raise_on_redirect
+        self.raise_on_status = raise_on_status
+        self.respect_retry_after_header = respect_retry_after_header
+        self.default_sleep_time = default_sleep_time
+
+        self._create_http_session(headers)
+
+    def __del__(self):
+        self._close_http_session()
+
+    def _create_http_session(self, headers=None):
+        """
+        Create a http session and initialize the retry object
+
+        :param headers: list of session headers
+        """
+        self.session = requests.Session()
+
+        if headers:
+            self.session.headers.update(headers)
+
+        retries = Retry(total=self.max_retries,
+                        connect=self.max_retries_on_connect,
+                        read=self.max_retries_on_read,
+                        redirect=self.max_retries_on_redirect,
+                        status=self.max_retries_on_status,
+                        method_whitelist=self.method_whitelist,
+                        status_forcelist=self.status_forcelist,
+                        backoff_factor=self.default_sleep_time,
+                        raise_on_redirect=self.raise_on_redirect,
+                        raise_on_status=self.raise_on_status,
+                        respect_retry_after_header=self.respect_retry_after_header)
+
+        self.session.mount('http://', HTTPAdapter(max_retries=retries))
+        self.session.mount('https://', HTTPAdapter(max_retries=retries))
+
+    def _close_http_session(self):
+        """Close the http session"""
+
+        self.session.close()
+
+    def fetch(self, url, payload=None, headers=None, method=GET, stream=False):
+        """
+        Fetch the data from a given URL
+
+        :param url: link to the resource
+        :param payload: payload of the request
+        :param headers: headers of the request
+        :param method: type of request call (GET or POST)
+        :param stream: defer downloading the response body until the response content is available
+        """
+        return self.__send_request(url, payload, headers, method, stream)
+
+    def __send_request(self, url, payload=None, headers=None, method=GET, stream=False):
+        """Send a HTTP request to a URL"""
+
+        if method == self.GET:
+            response = self.session.get(url, params=payload, headers=headers, stream=stream)
+        else:
+            response = self.session.post(url, data=payload, headers=headers, stream=stream)
+
+        response.raise_for_status()
+
+        return response
+
+
+class RateLimitHandler:
+    """Class to handle rate limit for HTTP clients.
+
+        :param sleep_for_rate: sleep until rate limit is reset
+        :param min_rate_to_sleep: minimun rate needed to sleep until it will be rese
+        :param rate_limit_header: header to know the current rate limit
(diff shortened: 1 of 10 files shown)


perceval/client.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def _fetch(self, url, payload=None, headers=None, method=GET,
+               stream=False,
+               retry_on_status_codes=None,
+               retry_after_status_codes=None):
+
+        self.set_retry_after_status_codes(retry_after_status_codes)
+        self.sleep_for_rate_limit()
+
+        session = self.session or requests.Session()
```
sduenas: Why doesn't the client use always a session?

I would have seen this code in a different method, maybe as part of the initialization of the instance, to call it just once.

perceval/client.py
```diff
@@ -0,0 +1,185 @@
...
+            logger.debug("Rate limit: %s", self.rate_limit)
+        else:
+            self.rate_limit = None
+            self.rate_limit_reset_ts = None
+
+    def __send_request(self, session, url, payload=None, headers=None, method=GET, stream=False):
+        if method == self.GET:
+            response = session.get(url, params=payload, headers=headers, stream=stream)
+        else:
+            response = self.session.post(url, data=payload, headers=headers, stream=stream)
```
sduenas: I think this should be either `session` or `self.session`, but the same on the two calls. Moreover, the code should be something like this:

```
if method == self.GET:
    freq = session.get
else:
    freq = session.post

response = freq(url,  data=payload, headers=headers, stream=stream)
```

perceval/client.py
```diff
@@ -0,0 +1,185 @@
...
+    def create_http_session(self, headers=None):
+        self.session = requests.Session()
+
+        if headers:
+            self.session.headers.update(headers)
+
+    def close_http_session(self, session):
+        session.close()
+
+    def set_retry_after_status_codes(self, status_codes=None):
```
sduenas: Why is this needed?

perceval/client.py
```diff
@@ -0,0 +1,185 @@
...
+    def __init__(self, base_url, sleep_for_rate=False, min_rate_to_sleep=MIN_RATE_LIMIT,
+                 max_retries=MAX_RETRIES, default_sleep_time=DEFAULT_SLEEP_TIME):
+        self.base_url = base_url
+        self.rate_limit = None
+        self.rate_limit_reset_ts = None
+        self.session = None
+
+        self.sleep_for_rate = sleep_for_rate
+
+        if min_rate_to_sleep > self.MAX_RATE_LIMIT:
```
sduenas: I'm still not sure what to do with rate limit handlers. Not sure if we should move this behaviour to a specific class and use it as a decorator or to have a mixin. For instance:
```
class GitHubClient(RateLimitHandler, HttpClient):
    [...]
```

valeriocos: We can discuss about it. In your example, the responsabilities of the the HttpClient would be just about connection issues?

perceval/client.py
valeriocos: set_retry_after_status_codes is needed to modify the list of status codes were the retry-after header may appear. By default, those codes are 503, 413, 429, however in same cases (GitHub), the retry-after can appear also with other status codes (403)

perceval/client.py
valeriocos: From the [doc](http://docs.python-requests.org/en/latest/user/advanced/#session-objects), I read that "Session object allows you to persist certain parameters across requests". I guessed that we didn't need it and it was fine to create a session per get request.

perceval/client.py
valeriocos: get methods should pass the payload by the argument **params**, while post methods by the argument **data**. Btw, I guess the behavior is the same, maybe a couple of tests will need to be changed (If I remember well, Bugzilla or Bugzillarest).

perceval/client.py
sduenas: But, why do we need to do it this way? Why don't you need to modify Retry object and not the instance of this object?

perceval/client.py
sduenas: I see, but I think the docs make reference to parameters needed by the server that won't be harmful, like in Bugzilla with the session cookies. The server may define other cookies but in my opinion, it won't be a problem. And it will keep the configuration for retrying without having to define it each time a requests is sent.

perceval/client.py
sduenas: Yep, you're right. Btw, what's `stream` parameter and why do we need it?

perceval/client.py
sduenas: That's the idea. To create a session, to define which parameters are needed and fetch the items while connection or defined HTTP errors are handled to retry.

When at first I thought of handling rate limit in a generic client, it looked a good idea to me. But now I see it on the code, I feel it a bit weird. So, having a specific class which handles only those cases it's a better idea to me.

What do you think?

perceval/client.py
valeriocos: The stream parameter is used in **pipermail** and **hyperkitty** clients. It defers downloading the response body until you access the **Response.content** attribute (the default behavior is to immediately download the body of the response).

perceval/client.py
valeriocos: Let's try with a specific class

perceval/client.py
valeriocos: You're right, better to modify the instance of the object

perceval/client.py
sduenas: Great. I forgot about it!

perceval/client.py
valeriocos: I tried to modify the instance of the object, but the new error status codes were not taken into account (the test **test_abuse_rate_limit** was failing). So I left the code as it was

perceval/client.py
valeriocos: I added the class **RateLimitHandler** in the client.py. 
The github client remains a subclass of HttpClient -> **GitHubClient(HttpClient)**, while the HttpClient can accept a rate-limit-handler as parameter, which allows to deal with rate limit issues.

This is a bit different from your proposal (**GitHubClient(RateLimitHandler, HttpClient)**), but I'd would say that the generic handler should be coupled with the generic client. In case a given client wants to use the handler, it will have to set up the handler and pass it to the generic client, like this:

```python
rate_limit_handler = RateLimitHandler(sleep_for_rate, min_rate_to_sleep)
super().__init__(self.api_url, default_sleep_time=default_sleep_time, max_retries=max_retries,
                 rate_limit_handler=rate_limit_handler)
```

your thoughts?

perceval/client.py
```diff
@@ -0,0 +1,213 @@
...
+        yield self._fetch(url, payload, headers, method, stream,
+                          retry_on_status_codes,
+                          retry_after_status_code)
+
+    def _fetch(self, url, payload=None, headers=None, method=GET,
+               stream=False,
+               retry_on_status_codes=None,
+               retry_after_status_codes=None):
+
+        self.sleep_for_rate_limit()
```
sduenas: Instead og having this code here, I would do the next, taking advantage of multiple inheritance:
Let's take `GitHub` class the base case:

```
class GitHub(HttpClient, RateLimitHandler):

    def fetch(self, url, payload=None, headers=None, method=GET, retry_on_status_codes=None, retry_after_status_code=None):
        self.sleep_for_rate_limit()
        yield super().fetch(url, payload, headers, method, stream, retry_on_status_codes, retry_after_status_code)
        self.update_rate_limit(response)
```

Thus,  `sleep_for_rate_limit` and `update_rate_limit` can be safely removed from `HttpClient`. In addition, GitHub client `__init__` will need to call the `__init__` method of both parets.

https://stackoverflow.com/questions/26927571/multiple-inheritance-in-python3-with-different-signatures

Maybe we can do the same with pagination? What do you think?

perceval/client.py
```diff
@@ -0,0 +1,196 @@
...
+        yield self._fetch(url, payload, headers, method, stream,
+                          retry_on_status_codes,
+                          retry_after_status_code)
+
+    def _fetch(self, url, payload=None, headers=None, method=GET,
+               stream=False,
+               retry_on_status_codes=None,
+               retry_after_status_codes=None):
+
+        self.set_retry_after_status_codes(retry_after_status_codes)
```
sduenas: I would move all this retrying setup to `create_http_session()`.

perceval/client.py
```diff
@@ -0,0 +1,196 @@
...
+
+        if headers:
+            session.headers.update(headers)
+
+        return session
+
+    def close_http_session(self):
+        self.session.close()
+
+    def set_retry_after_status_codes(self, status_codes=None):
```
sduenas: I still don't see why this is needed. I mean, why can't these parameters being set during Retry initialization in line 107?

perceval/client.py
valeriocos: if we move it to _create_http_session()_, we won't be able to control every call to __fetch_. Imagine that for a given call we want to set the retry after for a 403, but in another call we don't want.

Btw, I admit I haven't found any example in the backends where this happens

perceval/client.py
valeriocos: I don't see any parameter to set the retry-after-status-codes during the initialization of the object. 
I have tried to set them up just after the initialization (line 118) but I didn't succeed.

I'm looking at [urllib3](http://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html)

perceval/client.py
```diff
@@ -0,0 +1,236 @@
...
+#
+# Authors:
+#     Valerio Cosentino <valcos@bitergia.com>
+#
+
+import logging
+import time
+
+import requests
+from requests.adapters import HTTPAdapter
```
sduenas: Use only `import requests` and `import urllib3.util`. And use them when calling to `HTTPAdapter` or `Retry`. It's clearer.

perceval/client.py
```diff
@@ -0,0 +1,236 @@
...
+
+from .errors import RateLimitError
+
+logger = logging.getLogger(__name__)
+
+
+class HttpClient:
+    """Abstract class for HTTP clients.
+
+    Base class to query data sources taking care of retrying requests in case connection issues. If the data source does
```
sduenas: Make these lines shorter and update the documentation. There are references to methods like `_fetch` which are no longer in use.

perceval/client.py
```diff
@@ -0,0 +1,236 @@
...
+    DEFAULT_RAISE_ON_STATUS = True
+    DEFAULT_RESPECT_RETRY_AFTER_HEADER = True
+
+    DEFAULT_RETRY_AFTER_STATUS_CODES = [413, 429, 503]
+    DEFAULT_STATUS_FORCE_LIST = [408, 423, 504]
+
+    GET = "GET"
+    POST = "POST"
+
+    def __init__(self, base_url, max_retries=MAX_RETRIES, max_retries_on_connect=MAX_RETRIES_ON_CONNECT,
```
sduenas: Do we really need all these configuration parameters? I think we should stick to only those we will use for sure. If we need more, we will add them in future versions.

perceval/client.py
```diff
@@ -0,0 +1,236 @@
...
+        self.raise_on_status = raise_on_status
+        self.respect_retry_after_header = respect_retry_after_header
+        self.default_sleep_time = default_sleep_time
+
+        self._create_http_session(headers)
+
+    def __del__(self):
+        self._close_http_session()
+
+    def _create_http_session(self, headers=None):
```
sduenas: Private methods go after public methods.

perceval/client.py
```diff
@@ -0,0 +1,236 @@
...
+        """
+        Create a http session and initialize the retry object
```
sduenas: Please update the docstrings to the style we have in other parts of the code.

perceval/client.py
```diff
@@ -0,0 +1,236 @@
...
+
+        self.session.mount('http://', HTTPAdapter(max_retries=retries))
+        self.session.mount('https://', HTTPAdapter(max_retries=retries))
+
+    def _close_http_session(self):
+        """Close the http session"""
+
+        self.session.close()
+
+    def fetch(self, url, payload=None, headers=None, method=GET, stream=False):
```
sduenas: Same comment as above with the documentation.

perceval/client.py
```diff
@@ -0,0 +1,236 @@
...
+        """
+        Fetch the data from a given URL
+
+        :param url: link to the resource
+        :param payload: payload of the request
+        :param headers: headers of the request
+        :param method: type of request call (GET or POST)
+        :param stream: defer downloading the response body until the response content is available
+        """
+        return self.__send_request(url, payload, headers, method, stream)
```
sduenas: As this method doesn't do anything else than calling `_send_request`, they should be merged into one.

perceval/client.py
```diff
@@ -0,0 +1,236 @@
...
+        else:
+            response = self.session.post(url, data=payload, headers=headers, stream=stream)
+
+        response.raise_for_status()
+
+        return response
+
+
+class RateLimitHandler:
+    """Class to handle rate limit for HTTP clients.
```
sduenas: Check documentation style all along this class.

tests/test_client.py
```diff
@@ -0,0 +1,237 @@
...
+    def __init__(self, base_url, default_sleep_time=0.1, max_retries=1):
+
+        super().__init__(base_url, default_sleep_time=default_sleep_time, max_retries=max_retries)
+        super().setup_rate_limit_handler()
+
+
+class TestHttpClient(unittest.TestCase):
+    """Http client tests """
+
+    def test_initialization(self):
```
sduenas: We need another test where we check that the parameters passed to the class are well set. I mean, there aren't assignment errors to variables.

tests/test_client.py
```diff
@@ -0,0 +1,237 @@
...
+                               CLIENT_SPIDERMAN_URL,
+                               body=output,
+                               status=200)
+
+        client = MockedClient(CLIENT_API_URL)
+        response = client.fetch(CLIENT_SPIDERMAN_URL)
+        self.assertEqual(response.headers['connection'], 'close')
+
+    @httpretty.activate
+    def test_fetch(self):
```
sduenas: Slipt these checks into three independent tests. Therefore, a failure in one of the tests doesn't stop for testing the others.

tests/test_client.py
```diff
@@ -0,0 +1,237 @@
...
+        client = MockedClient(CLIENT_API_URL)
+
+        urls = [CLIENT_SPIDERMAN_URL, CLIENT_SUPERMAN_URL, CLIENT_BATMAN_URL]
+
+        for url in urls:
+            with self.assertRaises(requests.exceptions.RetryError):
+                _ = client.fetch(url)
+
+    @httpretty.activate
+    def test_init_rate_limit(self):
```
sduenas: Rate limit these tests should go in a different test class. I also miss some tests here.

sduenas:
A newer version of this PR has been merged. Closing it.

zhuqianbin:
Hi, How to let perceval support cookie? Since some web site need to establish a connection with cookie.

jgbarah:
@zhuqianbin y saw your comment by chance. Please, open an issue if you have a question like this (not a comment in a pull request). If you do, please clarify what do you mean. Which perceval backend (or data source) you intend to use, and which kind of cookie you are considering (because as far as I know, none of the data sources supported by Perceval use cookies).

PR #214, opened 2017-11-27
[github] Handle init rate limit for GitHub entreprise

valeriocos:
This patch allows to disable the initialization of the rate limit for GitHub entreprise repositories

### DIFF
--- a/perceval/backends/core/github.py
+++ b/perceval/backends/core/github.py
@@ -69,7 +69,7 @@ class GitHub(Backend):
     :param min_rate_to_sleep: minimun rate needed to sleep until
          it will be reset
     """
-    version = '0.11.3'
+    version = '0.11.4'
 
     def __init__(self, owner=None, repository=None,
                  api_token=None, base_url=None,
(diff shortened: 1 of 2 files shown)


perceval/backends/core/github.py
```diff
@@ -484,17 +484,17 @@ def __init__(self, owner, repository, token, base_url=None,
...
-
-    def init_rate_limit(self, response):
-        """initialize rate_limit and rate_limit_reset_ts """
+        try:
+            response = requests.get(url, headers=self.__build_headers())
+            response.raise_for_status()
+        except requests.exceptions.HTTPError:
+            self.rate_limit = None
+            self.rate_limit_reset_ts = None
+            return
```
sduenas: Instead of calling return here, it's better to remove it and put the next two lines in an `else` block, like this:

```
self.rate_limit = None
self.rate_limit_reset_ts = None

try:
     [...]
except requests.exceptions.HTTPError as e:
    # When rate limit is not available in GitHub Enterprise instances,
    # a HTTP 404 error is received
    if e.status_code != 404:
        raise e
else:
     self.rate_limit = int(response.headers[self.RATE_LIMIT_HEADER])
     self.rate_limit_reset_ts = int(response.headers[self.RATE_LIMIT_RESET_HEADER])
```

sduenas:
@valeriocos is this integrated in the new GitHubClient? If it is, we can close this PR. Otherwise, please integrate the needed changes.

valeriocos:
yes @sduenas  the requested changes are already integrated in the new GitHubClient

sduenas:
Ok. Let's close this PR then.

PR #219, opened 2017-11-28
[http-client] Generic client for backends querying APIs

valeriocos:
This PR proposes a generic client to deal with common problems (server errors, rate limit bans, connection losts) when fetching data from APIs.


sduenas:
Merged!

PR #228, opened 2017-11-30
[hyperkitty] Apply generic HTTP client to HyperKitty

valeriocos:
The generic client is applied to the HyperKitty backend. Now connection problems are transparently handled by the generic client.

### DIFF
--- a/perceval/backends/core/hyperkitty.py
+++ b/perceval/backends/core/hyperkitty.py
@@ -27,7 +27,6 @@ import os
 import dateutil.parser
 import dateutil.relativedelta
 import dateutil.tz
-import requests
 
 from grimoirelab.toolkit.datetime import datetime_to_utc, datetime_utcnow
 from grimoirelab.toolkit.uris import urijoin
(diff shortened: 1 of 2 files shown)


perceval/backends/core/hyperkitty.py
```diff
@@ -237,6 +230,26 @@ def _download_archive(self, url, params, filepath):
         return True
 
 
+class HyperKittyClient(HttpClient):
```
sduenas: This class doesn't do anything. Just use the `HttpClient` class in the backend.

perceval/backends/core/hyperkitty.py
```diff
@@ -221,13 +221,10 @@ def _parse_date_from_filepath(self, filepath):
 
         return dt
 
-    def _download_archive(self, url, params, filepath):
-        r = requests.get(url, params=params, stream=True)
-        r.raise_for_status()
-
+    def _read_archive(self, raw_content, url, filepath):
```
sduenas: Why don't you keep it as before? Just `_download archive` (downloading the content inside the method). I don't see the reason for changing it.

PR #237, opened 2017-12-01
[redmine] Apply generic HTTP client to Redmine

valeriocos:
The generic client is applied to the Redmine backend. Now connection problems are transparently handled by the generic client. Note that this PR shares with PR #236 some code (which extends the fetch method of the generic client).


sduenas:
I can't merge this because there are some conflicts.

valeriocos:
The conflicts have been fixed, however the tests fail due to an error solved in PR #228

PR #245, opened 2017-12-05
[http-client] Change format user-agent header

valeriocos:
This patch fixes the user-agent header according to the format for automated agents, which is:
**tool-name/version (+link-to-web-site)**


valeriocos:
The error is related to some *Accept* headers overwritten in other backends. Thus, this PR is rejected and the fix will be provided in another PR.

Issue #244, opened 2017-12-05
[meetup] Negative sleep length


albertinisg:
With the latest version, perceval complains about negative sleep length value. Traceback:
```
2017-12-05 09:03:08,707 - grimoire_elk.arthur - ERROR - Error feeding ocean from meetup (https://meetup.com/): sleep length must be non-negative
Traceback (most recent call last):
  File "./grimoire_elk/arthur.py", line 134, in feed_backend
    ocean_backend.feed()
  File "./grimoire_elk/ocean/elastic.py", line 203, in feed
    for item in items:
  File "/usr/local/lib/python3.4/dist-packages/perceval-0.9.7-py3.4.egg/perceval/backend.py", line 360, in decorator
    for data in func(self, *args, **kwargs):
  File "/usr/local/lib/python3.4/dist-packages/perceval-0.9.7-py3.4.egg/perceval/backends/core/meetup.py", line 122, in fetch
    event['comments'] = self.__fetch_and_parse_comments(event_id)
  File "/usr/local/lib/python3.4/dist-packages/perceval-0.9.7-py3.4.egg/perceval/backends/core/meetup.py", line 210, in __fetch_and_parse_comments
    for raw_page in raw_pages:
  File "/usr/local/lib/python3.4/dist-packages/perceval-0.9.7-py3.4.egg/perceval/backends/core/meetup.py", line 414, in comments
    for page in self._fetch(resource, params):
  File "/usr/local/lib/python3.4/dist-packages/perceval-0.9.7-py3.4.egg/perceval/backends/core/meetup.py", line 457, in _fetch
    self.sleep_for_rate_limit()
  File "/usr/local/lib/python3.4/dist-packages/perceval-0.9.7-py3.4.egg/perceval/client.py", line 204, in sleep_for_rate_limit
    time.sleep(seconds_to_reset)
ValueError: sleep length must be non-negative
```

PR #248, opened 2017-12-05
[meetup] Fix meetup negative sleep

valeriocos:
This PR fixes the issue #244 by adding a new method in the generic client which calculates the time to reset needed to regenerate the token. Due to this modification, the tests of the GitHub backend have changed accordingly.

### DIFF
--- a/perceval/client.py
+++ b/perceval/client.py
@@ -197,13 +197,18 @@ class RateLimitHandler:
            raises a RateLimitError exception if sleep_for_rate flag is disabled.
         """
         if self.rate_limit is not None and self.rate_limit <= self.min_rate_to_sleep:
-            seconds_to_reset = self.rate_limit_reset_ts - int(time.time()) + 1
+            seconds_to_reset = self.calculate_time_to_reset()
             cause = "Rate limit exhausted."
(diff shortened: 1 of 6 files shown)


perceval/client.py
```diff
@@ -197,14 +197,27 @@ def sleep_for_rate_limit(self):
...
-            seconds_to_reset = self.rate_limit_reset_ts - int(time.time()) + 1
+            seconds_to_reset = self.__calculate_seconds_to_reset()
             cause = "Rate limit exhausted."
             if self.sleep_for_rate:
                 logger.info("%s Waiting %i secs for rate limit reset.", cause, seconds_to_reset)
                 time.sleep(seconds_to_reset)
             else:
                 raise RateLimitError(cause=cause, seconds_to_reset=seconds_to_reset)
 
+    def __calculate_seconds_to_reset(self):
```
sduenas: I suggest to define this method as abstract that should be overwritten by each specific `RateLimitHandler`. Therefore, Meetup and GitHub will have their way to manage the time to reset. The method would be renamed to `calculate_time_to_reset` to make it more generic.

Issue #247, opened 2017-12-05
[meetup] Too many 429 responses


albertinisg:
Using the latest version, seems we are analyzing too many groups for Meetup, and perceval can't handle all the error responses returned. See the following traceback:
```
2017-12-05 10:09:52,658 - grimoire_elk.arthur - ERROR - Error feeding ocean from meetup (https://meetup.com/): HTTPSConnectionPool(host='api.meetup.com', port=443): Max retries exceeded with url: /Meetup-Rust-Lyon/events?fields=event_hosts,featured,group_topics,plain_text_description,rsvpable,series&status=cancelled,upcoming,past,proposed,suggested,draft&page=200&key=XXXXXXXXXXXXXXXXX&sign=true&scroll=since%3A1970-01-01T00%3A00%3A00.000Z&order=updated (Caused by ResponseError('too many 429 error responses',))
Traceback (most recent call last):
  File "/usr/local/lib/python3.4/dist-packages/requests/adapters.py", line 440, in send
    timeout=timeout
  File "/usr/local/lib/python3.4/dist-packages/urllib3-1.22-py3.4.egg/urllib3/connectionpool.py", line 732, in urlopen
    body_pos=body_pos, **response_kw)
  File "/usr/local/lib/python3.4/dist-packages/urllib3-1.22-py3.4.egg/urllib3/connectionpool.py", line 732, in urlopen
    body_pos=body_pos, **response_kw)
  File "/usr/local/lib/python3.4/dist-packages/urllib3-1.22-py3.4.egg/urllib3/connectionpool.py", line 732, in urlopen
    body_pos=body_pos, **response_kw)
  File "/usr/local/lib/python3.4/dist-packages/urllib3-1.22-py3.4.egg/urllib3/connectionpool.py", line 732, in urlopen
    body_pos=body_pos, **response_kw)
  File "/usr/local/lib/python3.4/dist-packages/urllib3-1.22-py3.4.egg/urllib3/connectionpool.py", line 732, in urlopen
    body_pos=body_pos, **response_kw)
  File "/usr/local/lib/python3.4/dist-packages/urllib3-1.22-py3.4.egg/urllib3/connectionpool.py", line 712, in urlopen
    retries = retries.increment(method, url, response=response, _pool=self)
  File "/usr/local/lib/python3.4/dist-packages/urllib3-1.22-py3.4.egg/urllib3/util/retry.py", line 388, in increment
[...]

PR #249, opened 2017-12-07
[meetup] Handle 429 HTTP error

valeriocos:
This patch handles 429 HTTP errors, which appear when issuing to many requests. Once a retry error is thrown, the backend sleeps for the remaining of the current hour and tries again later.


valeriocos:
This issue should not be addressed by Perceval, but by other components (Arthur, Mordred, etc.). See comments on issue #247

PR #250, opened 2017-12-08
[docker] Add build-essential/gcc for dulwich

vwangler:
The dulwich build needs GCC and the docker build has been failing since it was added as a requirement.


docker/images/master/Dockerfile
```diff
@@ -8,7 +8,7 @@ ENV GIT_REV_PERCEVAL master
 # install dependencies
 
 RUN apt-get update && \
-    apt-get install -y git --no-install-recommends && \
+    apt-get install -y build-essential git --no-install-recommends && \
```
albertinisg: The set of packages increases the image size more than 100 MB, and those packages are not needed for perceval to run. After cleaning the cached packages, I would purge them:
```
apt-get purge build-essential -y
```

albertinisg:
LGTM

sduenas:
Merged. Thanks @vwangler

PR #251, opened 2017-12-11
[http-client] Fix class attributes

valeriocos:
This patch fixes the initialization of headers and status codes (forcelist and retry after status). Before they were initialized by references, causing side effects in several tests (launchpad, github, etc.). Now, they are initialized by value.
Some backends have been modified according to the new http-client parameters.

### DIFF
--- a/perceval/client.py
+++ b/perceval/client.py
@@ -50,13 +50,10 @@ class HttpClient:
     :param base_url: base URL of the data source
     :param max_retries: number of max retries to a data source
         before raising a RetryError exception
-    :param status_forcelist: list of status codes where the retry
-        attempts occur
     :param default_sleep_time: default time to sleep in case
         of connection problems
-    :param headers: list of session headers
     """
-    version = '0.1'
+    version = '0.1.1'
 
     DEFAULT_SLEEP_TIME = 1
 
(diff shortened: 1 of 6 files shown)


perceval/client.py
```diff
@@ -79,18 +76,29 @@ class HttpClient:
     GET = "GET"
     POST = "POST"
 
-    def __init__(self, base_url, max_retries=MAX_RETRIES, status_forcelist=DEFAULT_STATUS_FORCE_LIST,
-                 default_sleep_time=DEFAULT_SLEEP_TIME, headers=DEFAULT_HEADERS):
+    def __init__(self, base_url, max_retries=MAX_RETRIES, default_sleep_time=DEFAULT_SLEEP_TIME,
+                 extra_headers={}, extra_status_forcelist=[], extra_retry_after_status=[]):
```
sduenas: This is an error. We can't initialize default values that way because they are mutable.

See http://docs.python-guide.org/en/latest/writing/gotchas/

tests/test_client.py
```diff
@@ -48,14 +48,15 @@
 class MockedClient(HttpClient, RateLimitHandler):
 
     def __init__(self, base_url, default_sleep_time=0.1, max_retries=1,
-                 status_force_list=HttpClient.DEFAULT_STATUS_FORCE_LIST,
-                 headers=HttpClient.DEFAULT_HEADERS, sleep_for_rate=False,
+                 extra_status_forcelist=[],
```
sduenas: Add some values to `extra_status_forcelist` to check those values are added to the list. You should check the values of the list to. The same with `extra_headers`. Update the dict with some other headers.

PR #253, opened 2017-12-14
[archive] Add archive

valeriocos:
This PR simplifies the retrieval of cached data.

In a nutshell, the data returned during fetch executions is stored in different archives (one per execution) which allow to transparently play back previous fetch executions, enabling the recovery of the stored data as it was originally collected from the data source.

This PR replaces [PR 194](https://github.com/grimoirelab/perceval/pull/194)

### DIFF
--- a/perceval/archive.py
+++ b/perceval/archive.py
@@ -1,0 +1,344 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2015-2017 Bitergia
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, 51 Franklin Street, Fifth Floor, Boston, MA 02110-1335, USA.
+#
+# Authors:
+#     Valerio Cosentino <valcos@bitergia.com>
+#     Santiago Dueñas <sduenas@bitergia.com>
+#
+
+import hashlib
+import json
+import logging
+import os
+import pickle
+import sqlite3
+
+from grimoirelab.toolkit.datetime import (datetime_utcnow,
+                                          datetime_to_utc,
+                                          str_to_datetime)
+
+from .errors import ArchiveError
+
+
+logger = logging.getLogger(__name__)
+
+
+class Archive:
+    """Basic class for archiving raw items fetched by Perceval.
+
+    This class allows to archive raw items - usually HTML pages or
+    JSON documents - for a further recovery. These raw items will
+    be fetched, stored and retrieved back by a backend.
+
+    Each stored item will have a hash code used as unique identifier.
+    Hash codes are generated using URIs and other parameters needed
+    to fetch raw items.
+
+    When an instance of `Archive` is initialized it will expect
+    to access an existing archive file. To create a new and empty
+    archive used `create` class method instead. Metadata must be
+    initialized calling to `init_metadata` method after creating
+    a new archive.
+
+    :param archive_path: path where this archive is stored
+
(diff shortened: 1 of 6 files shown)


perceval/archive.py
```diff
@@ -0,0 +1,312 @@
...
+class Archive:
+    """Basic storage for Perceval.
+
+    This class allows to archive items for a
+    further recovery. Items are stored in a Sqlite db.
+
+    :param archive_path: path where to load/retrieve the archive
+    """
+
+    def __init__(self, archive_path):
```
sduenas: This method should check whether the archive exists and it's a valid database.
It should also load metadata information from the database.

perceval/archive.py
```diff
@@ -0,0 +1,312 @@
...
+
+        cursor = conn.cursor()
+        cursor.execute("DROP TABLE IF EXISTS archive")
+        cursor.execute("DROP TABLE IF EXISTS metadata")
+        conn.commit()
+
+        metadata_stmt = "CREATE TABLE metadata ( " \
+                        "origin TEXT, " \
+                        "backend_name TEXT, " \
+                        "backend_version TEXT, " \
```
sduenas: I think we need to include the category of the items fetched during the process.
I'm not sure if we should include `from_date` field this way. There are other possible filters like `offset` that are used in some  backends.

perceval/archive.py
```diff
@@ -0,0 +1,312 @@
...
+                       "stream TEXT, " \
+                       "verify TEXT, " \
+                       "response BLOB)"
+
+        cursor.execute(metadata_stmt)
+        cursor.execute(archive_stmt)
+        conn.commit()
+
+        cursor.close()
+        conn.close()
```
sduenas: This method must return an instance of the created class.

perceval/archive.py
```diff
@@ -0,0 +1,312 @@
...
+
+    def from_date(self):
```
sduenas: As `from_date` is a metadata field, I thought this field and others would be attributes of the instance. This attributes would be loaded when the class is initialized or when they are set using a method (for instance, a method called `init`)

perceval/archive.py
```diff
@@ -0,0 +1,312 @@
...
+        cursor.close()
+
+        if row:
+            found = pickle.loads(row[0])
+        else:
+            raise ArchiveError(cause="No metadata found")
+
+        return found
+
+    def metadata(self, origin, backend_name, backend_version, from_date):
```
sduenas: As I said above, I would name this method `init` or `initialize`, that will initialize metadata. This will be only called when the backend is fetching from a remote repository (so far, a HTTP or a Gerrit server).

Maybe we can have a property method named `metadata` to get these values from the database.

perceval/archive.py
```diff
@@ -0,0 +1,312 @@
...
+    return now.strftime("%Y-%m-%d %H:%M:%S")
+
+
+def dict_to_string(json_data):
+    """convert a dict to str"""
+
+    return json.dumps(json_data, sort_keys=True)
+
+
+def make_digest(url, payload, headers, method, stream, verify):
```
sduenas: Not sure if this is the best name for this function. As this is part of `Archive` class and it's only used there I think this should be a static method of that class.

perceval/backend.py
```diff
@@ -147,16 +169,18 @@ class BackendCommandArgumentParser:
...
     :param aliases: define aliases for parsed arguments
+    :param archive: set archiving arguments
 
     :raises AttributeArror: when both `from_date` and `offset` are set
         to `True`
     """
     def __init__(self, from_date=False, to_date=False, offset=False,
                  basic_auth=False, token_auth=False,
-                 cache=False, aliases=None):
+                 cache=False, aliases=None, archive=None):
```
sduenas: The value of `archive` here should be `False`.

perceval/archive.py
valeriocos: ok, what do you think about _make_hash_ ?

perceval/archive.py
valeriocos: why you are not sure? 
I would add the other parameters (offset, etc.) to the table metadata

sduenas: I forgot to mention it. I'm not sure how to store them. I mean, should we store parameters used by `fech` one field each? Or in other hand, one field that stores all parameters in a dict or a JSON structure?

valeriocos: I would go for one field each, if there are not many parameters to store

PR #256, opened 2017-12-15
[launchpad] Handle 404 HTTP error

valeriocos:
This patch allows to handle 404 http errors, which may occur when requesting data of deleted user.

### DIFF
--- a/perceval/backends/core/launchpad.py
+++ b/perceval/backends/core/launchpad.py
@@ -440,8 +440,8 @@ class LaunchpadClient(HttpClient):
             raw_user = self.__send_request(url_user, headers=self.__get_headers())
             user = raw_user
         except requests.exceptions.HTTPError as e:
-            if e.response.status_code == 410:
-                logger.warning("Data is not available due to HTTP 410 Gone - %s", url_user)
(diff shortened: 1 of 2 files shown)


valeriocos:
You're right! I'm going to upload a unit test for this.
I'm not totally sure if we need to check the 404 status code for __fetch_items. I would add it because the API seems to behave the same when fetching data that is not available anymore.
What do you think?

sduenas:
My concerns about handling these kind of errors are that they can hide bugs in our code. I'm won't be sure if this error is raised because the URL is wrong or because the content doesn't longer exist. 

Witn `users` we discovered that the URL is well formed but when the request is send to a deleted users, Launchpad returns a 404 or a 410. That's totally fine and that's why we handle the error. In the other hand, I'm not sure if that's the case with other items like attachments, comments, etc.

I vote for removing it and wait for the errors. Once we know for sure that Launchpad returns those errors on removed contents, we could handle them that way. What do you think?

acs:
@sduenas I will try to do the testing for 404 or a 410 responses in the dev team to avoid the time needed to test it in operation, detect the errors and return back to dev.

PR #259, opened 2017-12-21
[gerrit] Add support so --ssh-port could has no params

acs:
If you don't want to use a specific port for communicating with
the remote gerrit ssh server, just add --ssh-port with no params.


acs:
We have found an alternative solution for this issue. Closing this PR because it is not needed anymore.

Issue #262, opened 2017-12-26
Local repositories cloned into a wrong directory


sduenas:
When a local Git repository is analized by Perceval, the directory where is cloned is created inside the local repository. 

In the next example, the repository is wrongly cloned in `/tmp/no_refs_bare/-git`.

```
$ perceval -g git /tmp/no_refs_bare/ --latest-items
[2017-12-26 12:33:37,916 - root - INFO] - Sir Perceval is on his quest.
[2017-12-26 12:33:37,918 - perceval.backends.core.git - INFO] - Fetching latest commits: '/tmp/no_refs_bare/' git repository
[2017-12-26 12:33:37,919 - perceval.backends.core.git - DEBUG] - Running command git count-objects -v (cwd: /tmp/no_refs_bare/-git, env: {'PAGER': '', 'HOME': '/home/sduenas', 'LANG': 'C'})
[2017-12-26 12:33:37,921 - perceval.backends.core.git - DEBUG] - 
[2017-12-26 12:33:37,921 - perceval.backends.core.git - DEBUG] - Git /tmp/no_refs_bare/ repository has 6 objects
[2017-12-26 12:33:37,921 - perceval.backends.core.git - DEBUG] - Running command git show-ref --heads --tags (cwd: /tmp/no_refs_bare/-git, env: {'PAGER': '', 'HOME': '/home/sduenas', 'LANG': 'C'})
[2017-12-26 12:33:37,924 - perceval.backends.core.git - DEBUG] - 
[2017-12-26 12:33:37,924 - perceval.backends.core.git - DEBUG] - Running command git ls-remote -h -t --exit-code origin (cwd: /tmp/no_refs_bare/-git, env: {'PAGER': '', 'HOME': '/home/sduenas', 'LANG': 'C'})
[2017-12-26 12:33:37,928 - perceval.backends.core.git - DEBUG] - 
[2017-12-26 12:33:37,936 - perceval.backends.core.git - DEBUG] - Running command git verify-pack -v objects/pack/pack-da39a3ee5e6b4b0d3255bfef95601890afd80709 (cwd: /tmp/no_refs_bare/-git, env: {'PAGER': '', 'HOME': '/home/sduenas', 'LANG': 'C'})
[2017-12-26 12:33:37,939 - perceval.backends.core.git - DEBUG] - 
[...]

PR #269, opened 2018-01-02
[git] Fix cloned directory of local repository

valeriocos:
This patch fixes the issue #262. The git_path is derived by joining the default Perceval location plus the path of the local Git repository to analyse, stripped of the initial slash. 
The initial slash in a component of a os.path.join causes all previous components to be thrown away and the joining continues from the absolute path component.

### DIFF
--- a/perceval/backends/core/git.py
+++ b/perceval/backends/core/git.py
@@ -288,7 +288,8 @@ class GitCommand(BackendCommand):
             git_path = self.parsed_args.git_log
         elif not self.parsed_args.git_path:
             base_path = os.path.expanduser('~/.perceval/repositories/')
-            git_path = os.path.join(base_path, self.parsed_args.uri) + '-git'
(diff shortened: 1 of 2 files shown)


perceval/backends/core/git.py
```diff
@@ -288,12 +288,23 @@ def _pre_init(self):
...
             base_path = os.path.expanduser('~/.perceval/repositories/')
-            git_path = os.path.join(base_path, self.parsed_args.uri) + '-git'
+            processed_uri = self.parsed_args.uri.lstrip('/')
+            git_path = os.path.join(base_path, processed_uri) + '-git'
         else:
             git_path = self.parsed_args.git_path
 
         setattr(self.parsed_args, 'gitpath', git_path)
 
+    # base_path = os.path.expanduser('~/.perceval/repositories/')
```
sduenas: Please remove this commented code

PR #274, opened 2018-01-16
[archive-manager] Add archive manager

valeriocos:
This code adds a manager to handle archives. It allows to create, load, delete and list archives.


valeriocos:
This PR is replaced by PR [317](https://github.com/grimoirelab/perceval/pull/317)

PR #275, opened 2018-01-16
[github] Apply archive to GitHub backend

valeriocos:
This PR allows the GitHub backend to store fetched data to archives. Thus it allows to transparently play back previous fetch executions, enabling the recovery of the stored data as it was originally collected from the data source.


valeriocos:
the modifications on the GitHub backend do not seem to be integrated to the codebase

PR #278, opened 2018-01-17
[askbot] Apply archive to Askbot backend

valeriocos:
This PR allows the Askbot backend to store fetched data to archives. Thus it allows to transparently play back previous fetch executions, enabling the recovery of the stored data as it was originally collected from the data source.


perceval/backends/core/askbot.py
```diff
@@ -73,11 +72,16 @@ def fetch(self, from_date=DEFAULT_DATETIME):
 
         :returns: a generator of items
         """
-
         if not from_date:
             from_date = DEFAULT_DATETIME
 
-        from_date = datetime_to_utc(from_date).timestamp()
+        kwargs = {'from_date': from_date}
+        return super().fetch("question", **kwargs)
```
sduenas: This is a matter of style but I'd rather return a variable instead the call to a method like this. It's better for debugging. Although, this is not the case because the returning value is an iterator. In any case, I think we should be consistent and return here a value.

```
questions = super().fetch("question", **kwargs)
	
return questions
```

PR #283, opened 2018-01-17
[dockerhub] Apply archive to Dockerhub backend

valeriocos:
This PR allows the Dockerhub backend to store fetched data to archives. Thus it allows to transparently play back previous fetch executions, enabling the recovery of the stored data as it was originally collected from the data source.

### DIFF
--- a/tests/test_dockerhub.py
+++ b/tests/test_dockerhub.py
@@ -24,7 +24,6 @@ import datetime
 import os
 import shutil
 import sys
-import tempfile
 import unittest
 import unittest.mock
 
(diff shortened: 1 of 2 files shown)


tests/test_dockerhub.py
```diff
@@ -133,7 +136,7 @@ def test_fetch(self, mock_utcnow):
         self.assertEqual(item['uuid'], '0fa16dc4edab9130a14914a8d797f634d13b4ff4')
         self.assertEqual(item['origin'], 'https://hub.docker.com/grimoirelab/perceval')
         self.assertEqual(item['updated_on'], 1483228800.0)
-        self.assertEqual(item['category'], 'dockerhub-data')
+        self.assertEqual(item['category'], dockerhub.metadata_category(item))
```
sduenas: Keep this line as it was.

PR #287, opened 2018-01-18
[archive] Improve log message for data in archive not found

valeriocos:
This fix improves the log message generated when some data is not retrieved from the archive. It includes the uri, payload and headers of the request beyond its hash.


valeriocos:
Yes, probably it is better to avoid having a log with too much information

valeriocos:
I close this PR since we agree on avoiding charging the log with too much information

PR #296, opened 2018-01-18
[hyperkitty] Apply archive to Hyperkitty backend

valeriocos:
This PR allows the Hyperkitty backend to store fetched data to archives. Thus it allows to transparently play back previous fetch executions, enabling the recovery of the stored data as it was originally collected from the data source.


perceval/backends/core/mbox.py
```diff
@@ -84,23 +83,115 @@ def fetch(self, from_date=DEFAULT_DATETIME):
...
+    def fetch_items(self, **kwargs):
+        """Fetch the messages"""
+
+        from_date = kwargs['from_date']
+
         logger.info("Looking for messages from '%s' on '%s' since %s",
                     self.uri, self.dirpath, str(from_date))
 
-        mailing_list = MailingList(self.uri, self.dirpath)
+        self.client = MailingList(self.uri, self.dirpath)
```
sduenas: I wouldn't change the name of this variable. It's not exactly a client so I'd rather keep it as it is now.

perceval/backends/core/mbox.py
```diff
@@ -84,23 +83,115 @@ def fetch(self, from_date=DEFAULT_DATETIME):
...
+        mbox = _MBox(filepath, create=False)
+
+        for msg in mbox:
+            message = message_to_dict(msg)
+            yield message
+
+    def _init_client(self, from_archive=False):
+        """Init mailing list"""
+
+        return MailingList(self.uri, self.dirpath)
```
sduenas: This method it isn't used inside `MBox` so don't override it and keep it as in `Backend`.

perceval/backends/core/hyperkitty.py
```diff
@@ -108,6 +131,11 @@ def has_resuming(cls):
         """
         return True
 
+    def _init_client(self, from_archive=False):
+        """Init mailing list"""
+
+        return HyperKittyList(self.url, self.dirpath, self.archive, from_archive)
```
sduenas: Same as in MBox.

perceval/backends/core/hyperkitty.py
```diff
@@ -100,6 +115,14 @@ def has_caching(cls):
         """
         return False
 
+    @classmethod
```
sduenas: I think this backend doesn't need support for archiving at this moment. Let's see how it works first with the HTTP backends and then decided how to support this feature in Mailing Lists.

valeriocos:
This PR is replaced by [PR 304](https://github.com/grimoirelab/perceval/pull/304)

PR #297, opened 2018-01-22
[backend] Add fetch, fetch_from_archive and fetch_items

valeriocos:
Generic backend methods to get the items from a remote repository (fetch) from the archive (fetch_from_archive) are added to the backend. They call fetch_items, which has to be implemented by the user to define the fetching process.

### DIFF
--- a/perceval/backend.py
+++ b/perceval/backend.py
@@ -35,7 +35,7 @@ from grimoirelab.toolkit.datetime import str_to_datetime
 
 from .archive import Archive
 from .cache import Cache, setup_cache
-from .utils import DEFAULT_DATETIME
+from .errors import ArchiveError
 from ._version import __version__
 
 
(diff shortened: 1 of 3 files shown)


perceval/backend.py
```diff
@@ -106,11 +106,45 @@ def archive(self, obj):
 
         self._archive = obj
 
-    def fetch(self, from_date=DEFAULT_DATETIME):
+    def fetch_items(self, **kwargs):
         raise NotImplementedError
 
-    def fetch_from_cache(self):
-        raise NotImplementedError
+    def fetch(self, category, **kwargs):
```
sduenas: Both `fetch` and `fetch_from_archive` must iterate over the items generated by `fetch_items`. Each item should be 
"decorated" with infomation from `metadata` function. It should be something like:

```
def fetch(self, category, **kwargs):
    if self.archive:
        self.archive.init_metadata(self.origin, self.__name__, self.version, category, kwargs)

    self.client = self._init_client()

    for item in self.fetch_items(**kwargs):
        yield metadata(item)
```

In fact, `fetch` and `fetch_from_archive` can use again a generic method with the code from above. What the code will change is the way archive is initialized and used.

perceval/backend.py
```diff
@@ -168,10 +205,12 @@ class BackendCommandArgumentParser:
     """
     def __init__(self, from_date=False, to_date=False, offset=False,
                  basic_auth=False, token_auth=False,
-                 cache=False, aliases=None):
+                 cache=False, aliases=None, category=None, kwargs=None):
```
sduenas: Why do you need `kwargs` here?

tests/test_backend.py
```diff
@@ -164,6 +217,73 @@ def test_cache_value_error(self):
             b.cache = 8
 
 
+class TestBackendArchive(unittest.TestCase):
```
sduenas: The test case is ok but I would make it generic. Subclassing it for each backend should be easy.

perceval/backend.py
```diff
@@ -106,11 +106,71 @@ def archive(self, obj):
...
+        The method retrieves items from a repository.
+
+        :param category: the category of the items fetched
+        :param kwargs: a list of other parameters (e.g., from_date, offset, etc.
+        specific for each backend)
+
+        :returns: a generator of items
+        """
+        if self.archive:
+            self.archive.init_metadata(self.origin, Backend.__name__, self.version, category,
```
sduenas: Don't use `Backend` class. Use `self` instead. Otherwise, subclasses won't use their name. `Backend` will always be stored.

tests/test_backend.py
```diff
@@ -43,10 +43,78 @@
...
+    def fetch_from_cache(self):
+        for x in range(5):
+            item = {
+                'item': x,
+                'cache': True
+            }
+            yield item
+
+    def _init_client(self, from_archive=False):
+
```
sduenas: Remove this blank line

tests/test_backend.py
```diff
@@ -163,6 +215,75 @@ def test_cache_value_error(self):
...
+    def test_fetch_items(self):
+        """Test whether an NotImplementedError exception is thrown"""
+
+        b = Backend('test')
+
+        with self.assertRaises(NotImplementedError):
+            b.fetch_items()
+
+    def test_init_client(self):
```
sduenas: This isn't a valid test because it might be failing because `fetch_items` is not implemented. Call to `_init_client` instead.

tests/test_backend.py
```diff
@@ -163,6 +215,75 @@ def test_cache_value_error(self):
...
+        """Test whether an NotImplementedError exception is thrown"""
+
+        b = Backend('test')
+
+        with self.assertRaises(NotImplementedError):
+            b.fetch_items()
+
+
+class TestBackendArchive(unittest.TestCase):
```
sduenas: I think this test should go in a different file that won't run. For instance, `base.py` in `tests` directory. In the same way, the test can be renamed to `TestCaseBackendArchive`.

The method `test_fetch_from_archive` won't receive any parameters but `self` (as other test methods). `backend` object will be called using `self.backend`. Derivated classes should define `setUp` method as follows (i.e Bugzilla):

```
class TestBugzillaBackendArchive(TestCaseBackendArchive):

    def setUp(self):
        super().setUp()
        self.backend = Bugzilla(archive=self,.archive, ...)

        # Setup here HTTP server using httpretty.
```

sduenas: You should define a `TestMockedBackendArchive` in this file too.

PR #299, opened 2018-01-24
[backend] Fix backend name in fetch method

valeriocos:
This patch fixes the name of the backend used to initialize the archive metadata.

### DIFF
--- a/perceval/backend.py
+++ b/perceval/backend.py
@@ -68,7 +68,7 @@ class Backend:
     :raises ValueError: raised when `cache` is not an instance of
         `Cache` class
     """
-    version = '0.5'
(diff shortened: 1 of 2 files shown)


perceval/backend.py
```diff
@@ -121,7 +121,7 @@ def fetch(self, category, **kwargs):
         :returns: a generator of items
         """
         if self.archive:
-            self.archive.init_metadata(self.origin, self.__name__, self.version, category,
+            self.archive.init_metadata(self.origin, self.__class__.__name__, self.version, category,
```
sduenas: To avoid further problems, why don't you create a new test which will check the initialization of an archive object?

PR #302, opened 2018-01-25
[gerrit] Apply archive to Gerrit backend

valeriocos:
This PR allows the Gerrit backend to store fetched data to archives. Thus it allows to transparently play back previous fetch executions, enabling the recovery of the stored data as it was originally collected from the data source.
Note that the code has not been tested

### DIFF
--- a/perceval/backends/core/gerrit.py
+++ b/perceval/backends/core/gerrit.py
@@ -31,9 +31,8 @@ from grimoirelab.toolkit.datetime import datetime_to_utc
 
 from ...backend import (Backend,
                         BackendCommand,
-                        BackendCommandArgumentParser,
-                        metadata)
-from ...errors import BackendError, CacheError
+                        BackendCommandArgumentParser)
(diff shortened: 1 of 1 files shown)


perceval/backends/core/gerrit.py
```diff
@@ -311,6 +324,21 @@ def __init__(self, repository, user, max_reviews, blacklist_reviews=[],
...
 
     def __execute(self, cmd):
+        """ Execute gerrit command"""
+
+        if self.from_archive:
+            data = self.archive.retrieve(cmd, "", "")
+        else:
+            try:
+                data = self.__execute_remote(cmd)
+                self.archive.store(cmd, "", "", data)
```
sduenas: `payload` and `hearders` parameters should be set to `None`.

valeriocos:
This PR is replaced by [PR 303](https://github.com/grimoirelab/perceval/pull/303)

PR #305, opened 2018-01-26
[pipermail] Apply archive to Pipermail backend

valeriocos:
This code enable archive functionalities to the Pipermail backend to store fetched data in the future.


perceval/backends/core/pipermail.py
```diff
@@ -81,6 +82,13 @@ def fetch(self, from_date=DEFAULT_DATETIME):
 
         :returns: a generator of messages
         """
+        super().fetch(from_date)
```
sduenas: Fetch doesn't return anything. This makes test to fail.

PR #308, opened 2018-01-26
[nntp] Apply archive to NNTP backend

valeriocos:
This code allows the NNTP backend to store fetched data to archives. Thus it allows to transparently play back previous fetch executions, enabling the recovery of the stored data as it was originally collected from the data source.

### DIFF
--- a/perceval/backends/core/nntp.py
+++ b/perceval/backends/core/nntp.py
@@ -20,7 +20,6 @@
 #     Santiago Dueñas <sduenas@bitergia.com>
 #
 
-import functools
 import io
 import logging
 import nntplib
@@ -31,9 +30,8 @@ from grimoirelab.toolkit.datetime import str_to_datetime
 
 from ...backend import (Backend,
                         BackendCommand,
-                        BackendCommandArgumentParser,
-                        metadata)
-from ...errors import CacheError, ParseError
+                        BackendCommandArgumentParser)
+from ...errors import ArchiveError, ParseError
 from ...utils import message_to_dict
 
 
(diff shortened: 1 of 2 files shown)


perceval/backends/core/nntp.py
```diff
@@ -277,6 +216,118 @@ def parse_article(raw_article):
...
+        """
+        if method == NNTTPClient.GROUP:
+            data = self.handler.group(args)
+        elif method == NNTTPClient.OVER:
+            data = self.handler.over(args)
+        elif method == NNTTPClient.ARTICLE:
+            data = self.fetch_article(args)
+
+        if self.archive:
+            self.archive.store(method, args, "", data)
```
sduenas: When storing values in the archive, set to `None` those parameters that will never be used not an empty string.

perceval/backends/core/nntp.py
```diff
@@ -277,6 +216,118 @@ def parse_article(raw_article):
...
+
+        return data
+
+    def fetch_from_archive(self, method, args):
+        """Fetch data from the archive
+
+        :param method: the name of the command to execute
+        :param args: the arguments required by the command
+        """
+        data = self.archive.retrieve(method, args, "")
```
sduenas: Same as above.

perceval/backends/core/nntp.py
```diff
@@ -277,6 +216,118 @@ def parse_article(raw_article):
...
+
+        if type(data) == nntplib.NNTPTemporaryError:
```
sduenas: Don't use `type()` to check the type of an object. Use `isinstance()` instead.

PR #317, opened 2018-02-09
Add archive manager

sduenas:
The idea behind the archive manager is to handle a set of `Archive` objects in Perceval. This class manages the creation, deletion and access to this objects

### DIFF
--- a/perceval/archive.py
+++ b/perceval/archive.py
@@ -76,7 +76,7 @@ class Archive:
                            "origin TEXT, " \
                            "backend_name TEXT, " \
                            "backend_version TEXT, " \
-                           "item_category TEXT, " \
+                           "category TEXT, " \
                            "backend_params BLOB, " \
                            "created_on TEXT)"
 
@@ -88,7 +88,7 @@ class Archive:
         self.origin = None
         self.backend_name = None
         self.backend_version = None
-        self.item_category = None
+        self.category = None
         self.backend_params = None
         self.created_on = None
 
@@ -103,7 +103,7 @@ class Archive:
             conn.close()
 
     def init_metadata(self, origin, backend_name, backend_version,
-                      item_category, backend_params):
+                      category, backend_params):
         """Init metadata information.
 
         Metatada is composed by basic information needed to identify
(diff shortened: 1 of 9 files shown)


perceval/archive.py
```diff
@@ -342,3 +343,123 @@ def _count_table_rows(self, table_name):
             cursor.close()
 
         return row[0]
```
acs: LGTM the change from item_category to category. It is something that can bite us, but category is not used in general now, so it is the moment to do this change.

perceval/archive.py
```diff
@@ -342,3 +343,123 @@ def _count_table_rows(self, table_name):
...
+
+        :raises ArchiveManangerError: when an error occurs removing the
+            archive
+        """
+        try:
+            Archive(archive_path)
+        except ArchiveError as e:
+            raise ArchiveManagerError(cause=str(e))
+
+        os.remove(archive_path)
```
acs: Hmmm, but this just removes a file (archive_path must be a path to a file). I have supposed that in the Archive there will be several files. Am I missing something?

perceval/archive.py
```diff
@@ -342,3 +343,123 @@ def _count_table_rows(self, table_name):
...
+
+            yield archive_path, archive.created_on
+
+    def _search_files(self):
+        """Retrieve the file paths stored under the base path."""
+
+        for root, _, files in os.walk(self.dirpath):
+            for filename in files:
+                location = os.path.join(root, filename)
+                yield location
```
acs: It is really cool this search feature to find the archives for a given data source. It could also be interesting to search just for a origin, or for a backend_name, or category. So you can say: search the archive files for `git`. But I am not sure that I have a valid use case for this feature, so then current implementation is ok for me.

perceval/archive.py
acs: Looking to the code below, it seems that for the same data source, several Archive objects exists. So the relation is one file per each Archive object. And the ArchiveManager is the one managing all those Archive objects for a specific configuration: origin, backend, category. Right?

sduenas: Each time `perceval` runs, it creates a `Archive` object/file for that process. All the items fetched on that execution are stored in that archive. `ArchiveManager` manages all those `Archive` objects, no matter which the origin or backend were.

perceval/archive.py
sduenas: Probably that makes sense but as you say we can keep it as is and add other features in a near future.

PR #329, opened 2018-02-20
[archive-manager] Add default date value when searching for archives

valeriocos:
This code enables to not explicitly define a date (archived_after) to look for archives when using the search functionalities. By default, the archived_after is set to 1970-01-01

### DIFF
--- a/perceval/archive.py
+++ b/perceval/archive.py
@@ -34,6 +34,7 @@ from grimoirelab.toolkit.datetime import (datetime_utcnow,
                                           str_to_datetime)
 
 from .errors import ArchiveError, ArchiveManagerError
+from .utils import DEFAULT_DATETIME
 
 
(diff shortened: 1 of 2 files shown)


valeriocos:
If you want to retrieve all archives for a given backend, you may not want to set a date.
Anyway if you think this change is not needed in Perceval, it could go in arthur.  When settings the args for the archive, the archived_after should be set to a default value (if not defined)

sduenas:
My idea is to force the developer or the user to give a date. As the code always needs a date to look for archives, I think it makes more sense to give it. To get all the archives, give the default date.
Moreover, I think it should work in the same way in Arthur.

valeriocos:
ok, sounds good to me. I close this PR and align the Arthur code to this requirement

PR #333, opened 2018-02-27
[askbot] Add data about accepted answers

valeriocos:
This code adds the information about answers which have been marked as accepted.

### DIFF
--- a/perceval/backends/core/askbot.py
+++ b/perceval/backends/core/askbot.py
@@ -52,7 +52,7 @@ class Askbot(Backend):
     :param tag: label used to mark the data
     :param archive: archive to store/retrieve items
     """
-    version = '0.4.0'
+    version = '0.5.0'
 
     def __init__(self, url, tag=None, archive=None):
         origin = url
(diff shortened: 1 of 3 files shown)


perceval/backends/core/askbot.py
```diff
@@ -52,7 +52,7 @@ class Askbot(Backend):
     :param tag: label used to mark the data
     :param archive: archive to store/retrieve items
     """
-    version = '0.4.0'
+    version = '0.4.1'
```
sduenas: This PR modifies the structure of the items returned by this backend adding a new field `accepted`. For that reason, the version of the backend must be incremented to `0.5`.

tests/test_askbot.py
```diff
@@ -473,6 +473,7 @@ def test_fetch(self):
         json_comments = json.loads(comments)
 
         self.assertEqual(len(questions[0]['data']['answers']), len(questions[0]['data']['answer_ids']))
+        self.assertTrue(questions[0]['data']['answers'][0]['accepted'])
```
sduenas: Don't we have any test case when this value is `False`? We should check that too.

valeriocos: yes, I'm going to add it
```
self.assertFalse(questions[1]['data']['answers'][0]['accepted'])
```

PR #334, opened 2018-02-28
[backends] Align fetch signature to the backend generic one

valeriocos:
This PR aligns the signature of the fetch method of each backend to the one provided by the generic backend

### DIFF
--- a/tests/test_jira.py
+++ b/tests/test_jira.py
@@ -319,7 +319,7 @@ class TestJiraBackend(unittest.TestCase):
 
         jira = Jira(JIRA_SERVER_URL)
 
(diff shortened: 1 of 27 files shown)


tests/test_jira.py
```diff
@@ -319,7 +319,7 @@ def test_fetch_from_date(self):
 
         jira = Jira(JIRA_SERVER_URL)
 
-        issues = [issue for issue in jira.fetch(from_date)]
+        issues = [issue for issue in jira.fetch(from_date=from_date)]
```
sduenas: This shouldn't be part of this PR.

sduenas: This was needed because `category` is now the first parameter of `fetch`

PR #342, opened 2018-03-03
[mediawiki] Force sort request payload parameters

valeriocos:
This patch sorts payload parameters used for api remote requests to allow the correct retrieval of archived data. Furthermore, it includes in the fetched args stored in the archive also the value of the _test_mode variable. Thus, simplifying the generation of archives for testing purposes.

### DIFF
--- a/perceval/backends/core/mediawiki.py
+++ b/perceval/backends/core/mediawiki.py
@@ -20,13 +20,14 @@
 #     Alvaro del Castillo <acs@bitergia.com>
 #
 
-import datetime
 import json
 import logging
 
 import dateutil
 
-from grimoirelab.toolkit.datetime import datetime_to_utc, str_to_datetime
+from grimoirelab.toolkit.datetime import (datetime_to_utc,
+                                          datetime_utcnow,
+                                          str_to_datetime)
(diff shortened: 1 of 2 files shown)


perceval/backends/core/mediawiki.py
```diff
@@ -99,7 +99,7 @@ def fetch(self, category=CATEGORY_PAGE, from_date=DEFAULT_DATETIME, reviews_api=
         else:
             from_date = datetime_to_utc(from_date)
 
-        kwargs = {"from_date": from_date, "reviews_api": reviews_api}
+        kwargs = {"from_date": from_date, "reviews_api": reviews_api, "test_mode": self._test_mode}
```
sduenas: `_test_mode` is obsolete. It shouldn't be included in `kwargs` dictionary.
Originally, `_test_mode` was added to bypass a check `MAX_RECENT_DAYS` variable checking while testing this module. 
A better approach would be to mock the call of `datetime.datetime.now(dateutil.tz.tzlocal())` to use always the same date. (See https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/backends/core/mediawiki.py#L311)

PR #346, opened 2018-03-06
[gerrit] Add tests for gerrit backend

valeriocos:
This PR includes tests to check the gerrit backend functionalities. Currently the tests cover only gerrit versions >= 2.10


tests/test_gerrit.py
```diff
@@ -0,0 +1,407 @@
...
+                "'(status:open OR status:closed)' --all-approvals --comments --format=JSON --start=4"
+
+
+def read_file(filename, mode='r'):
+    with open(os.path.join(os.path.dirname(os.path.abspath(__file__)), filename), mode) as f:
+        content = f.read()
+    return content
+
+
+class MockedCommand():
```
sduenas: I see the point of this object but I don't think is the best way to mock a function. I suggest to use `unittest.mock` module instead with the help of a class similar to this one.
This call `setattr(obj, member, mock)` can propagate side-effects when something fails in a test.

tests/test_gerrit.py
```diff
@@ -0,0 +1,396 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, 51 Franklin Street, Fifth Floor, Boston, MA 02110-1335, USA.
+#
+# Authors:
+#     Valerio Cosentino <valcos@bitergia.com>
+#
+
+import datetime
+import mock
```
sduenas: Remove `mock` library. We use `unittest.mock` instead.

tests/test_gerrit.py
```diff
@@ -0,0 +1,396 @@
...
+import os
+import unittest.mock
```
sduenas: You also have add a line with `import unnitest`

tests/test_gerrit.py
```diff
@@ -0,0 +1,396 @@
...
+        """Test if it returns True when has_archiving is called"""
+
+        self.assertEqual(Gerrit.has_archiving(), True)
+
+    def test_has_resuming(self):
+        """Test if it returns False when has_resuming is called"""
+
+        self.assertEqual(Gerrit.has_resuming(), False)
+
+    @mock.patch('subprocess.check_output', mock_check_ouput)
```
sduenas: These calls should be done with `unittest.mock.patch`.

tests/test_gerrit.py
```diff
@@ -0,0 +1,396 @@
...
+
+from perceval.backends.core.gerrit import (CATEGORY_REVIEW, MAX_REVIEWS, PORT,
+                                           Gerrit,
+                                           GerritCommand,
+                                           GerritClient)
+
+from base import TestCaseBackendArchive
+
+
+GERRIT_REPO = "gerrit.acme.org"
```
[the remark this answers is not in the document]
sduenas: You still need to change this.

PR #348, opened 2018-03-09
[backend] Set category when calling fetch and fetch_from_archive

valeriocos:
This variable is needed to drive the fetching operation in backends with multiple categories (e.g. crates).

### DIFF
--- a/perceval/backend.py
+++ b/perceval/backend.py
@@ -72,7 +72,7 @@ class Backend:
     :raises ValueError: raised when `archive` is not an instance of
         `Archive` class
     """
-    version = '0.6.0'
+    version = '0.6.1'
 
     CATEGORIES = []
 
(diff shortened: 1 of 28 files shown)


valeriocos:
Yes sure (sorry I should have given more context)

An example is with the [crates backend](https://github.com/chaoss/grimoirelab-perceval-mozilla/blob/master/perceval/backends/mozilla/crates.py) (but I think it could happen with all backends which have multiple categories). In the case of crates, the variable [self.category](https://github.com/chaoss/grimoirelab-perceval-mozilla/blob/master/perceval/backends/mozilla/crates.py#L88) is initialized in the **fetch** method and used in **fetch_items**. However, when the execution starts from **fetch_from_archive**, that variable is not set, thus **fetch_items** will raise an error.

The information about the items category should be set when the fetch operation (i.e., from the remote data source or the archive) is executed (only at that time the backend should know the category of items that will be fetched). Thus, **self.category** should be set in the generic **fetch** or **fetch_from_archive**.

sduenas:
I understand your point better now. In any case, I find this solution a bit odd because this code is setting an object attribute needed by the method which is being called a few lines below.
Doesn't make more sense to pass the category to fetch_items too?

Issue #355, opened 2018-03-15
[github] Negative time sleep when calculating sleep_for_rate_limit


valeriocos:
The error seems to be related to the use of shared tokens between several processes (e.g., percevals or mordreds). The error doesn't pop up when every process uses a non-shared token 
```
mordred_1 | File "/usr/local/lib/python3.5/dist-packages/perceval/backends/core/github.py", line 467, in fetch
mordred_1 | self.sleep_for_rate_limit()
mordred_1 | File "/usr/local/lib/python3.5/dist-packages/perceval/client.py", line 242, in sleep_for_rate_limit
mordred_1 | time.sleep(seconds_to_reset)
mordred_1 | ValueError: sleep length must be non-negative
```

PR #356, opened 2018-03-15
[client] Avoid negative sleep time error for rate limit reset

valeriocos:
This patch fixes issue #355. It prevents to sleep for a negative number of seconds when waiting for tokens rate limit reset. The error seems to occur when a given token is shared among several fetching processes.

### DIFF
--- a/perceval/client.py
+++ b/perceval/client.py
@@ -53,7 +53,7 @@ class HttpClient:
     :param sleep_time: time to sleep in case
         of connection problems
     """
-    version = '0.1.3'
+    version = '0.1.4'
 
     DEFAULT_SLEEP_TIME = 1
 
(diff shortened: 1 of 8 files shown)


perceval/client.py
```diff
@@ -236,6 +236,8 @@ def sleep_for_rate_limit(self):
         """
         if self.rate_limit is not None and self.rate_limit <= self.min_rate_to_sleep:
             seconds_to_reset = self.calculate_time_to_reset()
+            seconds_to_reset = 1 if seconds_to_reset < 0 else seconds_to_reset
```
sduenas: If seconds to rest is zero or less than zero why to sleep one second? It should sleep 0.

valeriocos: you are right, it should be 0

Issue #324, opened 2018-02-14
Add --to-date to plugins


pwnfoo:
`--from-date` is a really awesome functionality but there is no option to mention an end date in any plugins.

It'd be awesome to have a `--to-date` or `--end-date` functionality to pull data within a given timeframe. This will he helpful for those who wants to pull data for just one year out of x years of data out of a given source.

valeriocos:
Hi @sachinkamath, 
In the generic backend the option *--to-date* is [available](https://github.com/chaoss/grimoirelab-perceval/blob/9220d045a63a8d7a86056405a29005056785aa53/perceval/backend.py#L226
), however at the moment no specific backend supports it.

Indeed this could be a nice functionality to include in perceval, what do you think @sduenas?

PR #360, opened 2018-03-20
[git] Add --to-date to git backend and its tests

anveshc05:
This partially solves #324 .

### DIFF
--- a/perceval/backends/core/git.py
+++ b/perceval/backends/core/git.py
@@ -37,7 +37,7 @@ from ...backend import (Backend,
                         BackendCommand,
                         BackendCommandArgumentParser)
 from ...errors import RepositoryError, ParseError
-from ...utils import DEFAULT_DATETIME
+from ...utils import DEFAULT_DATETIME, DEFAULT_LAST_DATETIME
 
 CATEGORY_COMMIT = 'commit'
 
@@ -75,8 +75,8 @@ class Git(Backend):
         self.uri = uri
         self.gitpath = gitpath
 
-    def fetch(self, category=CATEGORY_COMMIT, from_date=DEFAULT_DATETIME, branches=None,
-              latest_items=False):
+    def fetch(self, category=CATEGORY_COMMIT, from_date=DEFAULT_DATETIME, to_date=DEFAULT_LAST_DATETIME,
+              branches=None, latest_items=False):
         """Fetch commits.
 
         The method retrieves from a Git repository or a log file
@@ -104,6 +104,7 @@ class Git(Backend):
         :param category: the category of items to fetch
         :param from_date: obtain commits newer than a specific date
             (inclusive)
+        :param to_date: obtain commits older than a specific date
         :param branches: names of branches to fetch from (default: None)
         :param latest_items: sync with the repository to fetch only the
             newest commits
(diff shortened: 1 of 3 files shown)


anveshc05:
@jgbarah @valeriocos please review this.

perceval/utils.py
```diff
@@ -44,6 +44,9 @@
 
 DEFAULT_DATETIME = datetime.datetime(1970, 1, 1, 0, 0, 0,
                                      tzinfo=dateutil.tz.tzutc())
+LAST_DATETIME = datetime.datetime(2050, 1, 1, 0, 0, 0,
+                                  tzinfo=dateutil.tz.tzutc())
+CURRENT_DATETIME = datetime.datetime.now(dateutil.tz.tzutc())
```
sduenas: This approach will generate unexpected results when `perceval` is used as a library. This date will be generated when the module is loaded in Python. That means it will always use the same date in long running processes. For instance, in dameons.

`DEFAULT_LAST_DATE` should be a better solution but I would set it to `2100-01-01`, as we have in other places.

tests/test_git.py
```diff
@@ -173,6 +173,54 @@ def test_fetch(self):
 
         shutil.rmtree(new_path)
 
+    def test_fetch_till_date(self):
```
sduenas: I miss a test checking a range with `from_date` and `to_date`.

perceval/backends/core/git.py
```diff
@@ -112,8 +113,10 @@ def fetch(self, category=CATEGORY_COMMIT, from_date=DEFAULT_DATETIME, branches=N
         """
         if not from_date:
             from_date = DEFAULT_DATETIME
+        if not to_date:
+            to_date = CURRENT_DATETIME
 
-        kwargs = {'from_date': from_date, 'branches': branches, 'latest_items': latest_items}
+        kwargs = {'from_date': from_date, 'to_date': to_date, 'branches': branches, 'latest_items': latest_items}
```
sduenas: This line is getting long. Format it like:
```
kwargs = {
    'from_date: from_date,
    'to_date': to_date,
    ...
}
```

perceval/backends/core/git.py
```diff
@@ -126,6 +129,7 @@ def fetch_items(self, category, **kwargs):
 
         :returns: a generator of items
         """
+        to_date = kwargs['to_date']
```
sduenas: Put this line below `from_date` to follow the same order you used in the parameters.

anveshc05:
@sduenas I have made the changes. Please review this

perceval/backends/core/git.py
```diff
@@ -245,41 +254,47 @@ def __fetch_from_log(self):
...
         else:
             branches_text = ", ".join(branches)
 
-        logger.info("Fetching commits: '%s' git repository from %s; %s branches",
-                    self.uri, str(from_date), branches_text)
-
         # Ignore default datetime to avoid problems with git
         # or convert to UTC
+        if to_date == DEFAULT_LAST_DATETIME:
+            logger.info("Fetching commits: '%s' git repository from %s to CURRENT DATETIME; %s branches",
```
sduenas: I suggest to move this logging message outside the `if` block. It makes more sense there and we'll only need one message. Chance `CURRENT DATETIME` by the value of `to_date`. I'd rather be clear on what parameters the backend is using to fetch new items.

anveshc05:
@sduenas Made the changes. Please review.

sduenas:
Merged. Thanks!

anveshc05:
Thanks for helping.

Issue #369, opened 2018-04-03
Not a gzipped file (b'00')


dpose:
I've run `perceval git --latest-items gerrit_url` where `gerrit_url` is the url of a git repository on gerrit and I've found the error below:

```
2018-03-26 16:24:08,052 - grimoire_elk.arthur - ERROR - Error feeding ocean from git (https://gerrit.domain.org/gerrit/repo): Not a gzipped file (b'00')
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/dist-packages/dulwich/protocol.py", line 200, in read_pkt_line
    sizestr = read(4)
  File "/usr/lib/python3.5/gzip.py", line 274, in read
    return self._buffer.read(size)
  File "/usr/lib/python3.5/_compression.py", line 68, in readinto
    data = self.read(len(byte_view))
  File "/usr/lib/python3.5/gzip.py", line 461, in read
    if not self._read_gzip_header():
  File "/usr/lib/python3.5/gzip.py", line 409, in _read_gzip_header
    raise OSError('Not a gzipped file (%r)' % magic)
OSError: Not a gzipped file (b'00')

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.5/dist-packages/grimoire_elk/arthur.py", line 193, in feed_backend
    ocean_backend.feed(latest_items=latest_items)
  File "/usr/local/lib/python3.5/dist-packages/grimoire_elk/ocean/elastic.py", line 204, in feed
    self.feed_items(items)
  File "/usr/local/lib/python3.5/dist-packages/grimoire_elk/ocean/elastic.py", line 213, in feed_items
    for item in items:
  File "/usr/local/lib/python3.5/dist-packages/perceval/backend.py", line 129, in fetch
    for item in self.fetch_items(category, **kwargs):
  File "/usr/local/lib/python3.5/dist-packages/perceval/backends/core/git.py", line 140, in fetch_items
    latest_items)
  File "/usr/local/lib/python3.5/dist-packages/perceval/backends/core/git.py", line 258, in __fetch_from_repo
    commits = self.__fetch_newest_commits_from_repo(repo)
  File "/usr/local/lib/python3.5/dist-packages/perceval/backends/core/git.py", line 289, in __fetch_newest_commits_from_repo
    hashes = repo.sync()
  File "/usr/local/lib/python3.5/dist-packages/perceval/backends/core/git.py", line 900, in sync
[...]

PR #370, opened 2018-04-03
Fix the error "Not a gzipped file (b'00')"

dpose:
It freezes the version of the dulwich package installed as dependency
of perceval.
From now, a version >=0.18.5 and <0.19 will be installed instead of
the latest one.
https://github.com/chaoss/grimoirelab-perceval/issues/369

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,6 +2,6 @@ python-dateutil>=2.6.0
 requests>=2.7.0
 beautifulsoup4>=4.3.2
 feedparser>=5.1.3
-dulwich>=0.18.5
+dulwich>=0.18.5, <0.19
 urllib3>=1.22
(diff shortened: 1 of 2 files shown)


valeriocos:
The PR looks good to me. I have just some minor remarks.
Does the error appear with other versions of dulwich previous to the latest one (0.19)? If this is not the case, could you modify the dep constraint within the **setup.py**?
Could you also update the dep constraint in the **requirements.txt**?

dpose:
I tested it with the two previous versions to 0.19 (0.18.5 and 0.18.6) and it worked successfully with both of them. You are right @valeriocos , using the newest version possible is better (0.18.6 in this case).
I've update the requirements file too.

sduenas:
@dpose I think @valeriocos meant that you should add a range with the versions of Dulwich which work.

dpose:
@sduenas @valeriocos 

Understood! :ok_hand:

PR #373, opened 2018-04-03
[archive] Sanitize archive

valeriocos:
This patch enables removing credentials information available in url, payload and headers when storing data to the archive. Thus, it enables sharing archives between users, avoiding credentials exposure.

### DIFF
--- a/perceval/backends/core/bugzilla.py
+++ b/perceval/backends/core/bugzilla.py
@@ -60,7 +60,7 @@ class Bugzilla(Backend):
     :param tag: label used to mark the data
     :param archive: archive to store/retrieve items
     """
-    version = '0.10.2'
+    version = '0.10.3'
 
     CATEGORIES = [CATEGORY_BUG]
 
@@ -554,6 +554,28 @@ class BugzillaClient(HttpClient):
         req = self.fetch(url, payload=params)
 
         return req.text
+
+    @staticmethod
+    def sanitize_for_archive(url, headers, payload):
+        """Sanitize payload of a HTTP request by removing the login and password information
+        before storing/retrieving archived items
+
+        :param: url: HTTP url request
+        :param: headers: HTTP headers request
+        :param: payload: HTTP payload request
+
+        :returns url, headers and the sanitized payload
+        """
+        if BugzillaClient.PBUGZILLA_LOGIN in payload:
+            payload.pop(BugzillaClient.PBUGZILLA_LOGIN)
+
+        if BugzillaClient.PBUGZILLA_PASSWORD in payload:
+            payload.pop(BugzillaClient.PBUGZILLA_PASSWORD)
+
+        if BugzillaClient.PLOGIN in payload:
+            payload.pop(BugzillaClient.PLOGIN)
+
+        return url, headers, payload
 
     def __fetch_version(self):
         response = self.metadata()
--- a/perceval/backends/core/bugzillarest.py
+++ b/perceval/backends/core/bugzillarest.py
@@ -60,7 +60,7 @@ class BugzillaREST(Backend):
     :param tag: label used to mark the data
     :param archive: archive to store/retrieve items
     """
-    version = '0.8.2'
+    version = '0.8.3'
 
     CATEGORIES = [CATEGORY_BUG]
 
@@ -424,6 +424,28 @@ class BugzillaRESTClient(HttpClient):
                                     code=result['code'])
 
         return r.text
+
+    @staticmethod
+    def sanitize_for_archive(url, headers, payload):
+        """Sanitize payload of a HTTP request by removing the login, password and token information
+        before storing/retrieving archived items
+
+        :param: url: HTTP url request
+        :param: headers: HTTP headers request
+        :param: payload: HTTP payload request
+
+        :returns url, headers and the sanitized payload
+        """
+        if BugzillaRESTClient.PBUGZILLA_LOGIN in payload:
+            payload.pop(BugzillaRESTClient.PBUGZILLA_LOGIN)
+
+        if BugzillaRESTClient.PBUGZILLA_PASSWORD in payload:
+            payload.pop(BugzillaRESTClient.PBUGZILLA_PASSWORD)
+
+        if BugzillaRESTClient.PBUGZILLA_TOKEN in payload:
+            payload.pop(BugzillaRESTClient.PBUGZILLA_TOKEN)
+
+        return url, headers, payload
 
 
 class BugzillaRESTCommand(BackendCommand):
--- a/perceval/backends/core/discourse.py
+++ b/perceval/backends/core/discourse.py
@@ -52,7 +52,7 @@ class Discourse(Backend):
     :param tag: label used to mark the data
     :param archive: archive to store/retrieve items
     """
-    version = '0.9.2'
+    version = '0.9.3'
 
     CATEGORIES = [CATEGORY_TOPIC]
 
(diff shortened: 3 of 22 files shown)


perceval/client.py
```diff
@@ -133,8 +133,19 @@ def fetch(self, url, payload=None, headers=None, method=GET, stream=False, verif
...
         return response
 
+    @staticmethod
+    def sanitize_for_archive(url, headers, payload):
+        """Sanitize the URL, headers and payload of a HTTP request before storing/retrieving items
+
+        :param: url: HTTP url request
+        :param: headers: HTTP headers request
+        :param: payload: HTTP payload request
+        """
```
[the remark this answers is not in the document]
sduenas: This method is not tested in `test_client.py`.

tests/test_client.py
```diff
@@ -310,6 +319,26 @@ def test_fetch_from_archive_exception(self):
         with self.assertRaises(requests.exceptions.HTTPError):
             _ = client.fetch(CLIENT_SPIDERMAN_URL)
 
+    @httpretty.activate
+    def test_sanitize_for_archive(self):
```
sduenas: Not sure what you are trying to test here. I would check that calling to `satanize_for_archive` method (not the overriden version in MockedClient) returns the expected result.

tests/test_stackexchange.py
```diff
@@ -415,6 +415,22 @@ def request_callback(method, uri, headers):
...
+        payload = {'order': 'desc',
+                   'site': 'stackoverflow',
+                   'sort': 'activity',
+                   'pagesize': 1,
+                   'key': 'aaa',
+                   'filter': 'Bf*y*ByQD_upZqozgU6lXL_62USGOoV3)MFNgiHqHpmO_Y-jHR',
+                   'page': 1,
+                   'tagged': 'python'}
+
+        url, headers, payload = StackExchangeClient.sanitize_for_archive(None, None, payload)
```
sduenas: Instead of using a `None` value for `url` and `headers`, I suggest to check it when these parameters have values too. Thus, you can be sure that both values are not modified by the method.

This comment is applied to the rest of the backends which override this method.

tests/test_stackexchange.py
```diff
@@ -415,6 +415,22 @@ def request_callback(method, uri, headers):
...
+        with self.assertRaises(KeyError):
+            payload.pop("key")
+
```
sduenas: I think you should check if `payload` contains the same keys and values minus `key` key. The method could be modifying the whole dictionary and this check won't catch that error.

This comment is applied to the rest of the backends which override this method.

perceval/backends/core/stackexchange.py
```diff
@@ -279,6 +254,43 @@ def get_questions(self, from_date):
                                   nquestions,
                                   tquestions)
 
+    @staticmethod
+    def sanitize_for_archive(url, headers, payload):
+        """Sanitize the URL, headers and payload of a HTTP request before storing/retrieving items
+
```
sduenas: Add a `:returns` not specifying what this method returns. I think it would be useful to add paragraph, in the description of the method, about what parts are satanized and in which way.

This comment is applied to the rest of the backends which override this method.

perceval/backends/core/slack.py
```diff
@@ -357,6 +357,18 @@ def user(self, user_id):
...
 
+    @staticmethod
+    def sanitize_for_archive(url, headers, payload):
+        """Sanitize the URL, headers and payload of a HTTP request before storing/retrieving items
+
+        :param: url: HTTP url request
+        :param: headers: HTTP headers request
+        :param: payload: HTTP payload request
+        """
+        payload.pop(SlackClient.PTOKEN) if SlackClient.PTOKEN in payload else payload
```
sduenas: `else` statement is not needed here

perceval/backends/core/redmine.py
```diff
@@ -406,6 +406,18 @@ def user(self, user_id):
...
 
+    @staticmethod
+    def sanitize_for_archive(url, headers, payload):
+        """Sanitize the URL, headers and payload of a HTTP request before storing/retrieving items
+
+        :param: url: HTTP url request
+        :param: headers: HTTP headers request
+        :param: payload: HTTP payload request
+        """
+        payload.pop(RedmineClient.PKEY) if RedmineClient.PKEY in payload else payload
```
sduenas: `else` clause is not needed here.

tests/test_phabricator.py
```diff
@@ -828,6 +828,20 @@ def request_callback(method, uri, headers):
...
+                   'output': 'json',
+                   'params': '{"__conduit__": {"token": "aaaa"}, '
+                             '"phids": ["PHID-APPS-PhabricatorHeraldApplication", '
+                             '"PHID-APPS-PhabricatorMockApplication"]}'}
+
+        url, headers, payload = ConduitClient.sanitize_for_archive('http://example.com/api/phid.query', None, payload)
+        with self.assertRaises(KeyError):
+            params = json.loads(payload['params'])
+            params.pop("__conduit__")
+
```
sduenas: I miss a check here when '__conduit__' is not inside `payload` parameter.

perceval/backends/core/meetup.py
```diff
@@ -410,6 +410,19 @@ def rsvps(self, group, event_id):
...
 
+    @staticmethod
+    def sanitize_for_archive(url, headers, payload):
+        """Sanitize the URL, headers and payload of a HTTP request before storing/retrieving items
+
+        :param: url: HTTP url request
+        :param: headers: HTTP headers request
+        :param: payload: HTTP payload request
+        """
+        payload.pop(MeetupClient.PKEY) if MeetupClient.PKEY in payload else payload
```
sduenas: None of the `else` clauses are needed.

perceval/backends/core/discourse.py
```diff
@@ -335,6 +335,18 @@ def post(self, post_id):
...
 
+    @staticmethod
+    def sanitize_for_archive(url, headers, payload):
+        """Sanitize the URL, headers and payload of a HTTP request before storing/retrieving items
+
+        :param: url: HTTP url request
+        :param: headers: HTTP headers request
+        :param: payload: HTTP payload request
+        """
+        payload.pop(DiscourseClient.PKEY) if DiscourseClient.PKEY in payload else payload
```
sduenas: `else` clause is not needed here.

perceval/backends/core/bugzillarest.py
```diff
@@ -425,6 +425,20 @@ def call(self, resource, params):
...
 
+    @staticmethod
+    def sanitize_for_archive(url, headers, payload):
+        """Sanitize the URL, headers and payload of a HTTP request before storing/retrieving items
+
+        :param: url: HTTP url request
+        :param: headers: HTTP headers request
+        :param: payload: HTTP payload request
+        """
+        payload.pop(BugzillaRESTClient.PBUGZILLA_LOGIN) if BugzillaRESTClient.PBUGZILLA_LOGIN in payload else payload
```
sduenas: None of the `else` clauses are needed here.

perceval/backends/core/bugzilla.py
```diff
@@ -555,6 +555,20 @@ def call(self, cgi, params):
...
 
+    @staticmethod
+    def sanitize_for_archive(url, headers, payload):
+        """Sanitize the URL, headers and payload of a HTTP request before storing/retrieving items
+
+        :param: url: HTTP url request
+        :param: headers: HTTP headers request
+        :param: payload: HTTP payload request
+        """
+        payload.pop(BugzillaClient.PBUGZILLA_LOGIN) if BugzillaClient.PBUGZILLA_LOGIN in payload else payload
```
sduenas: None of the `else` clauses are needed.

PR #375, opened 2018-04-10
[jenkins] Control detail level of Jenkins data

valeriocos:
This patch allows to control the granularity of the data returned by the Jenkins API. By default this value (detail_depth) is set to 1. Note that increasing the detail_depth may considerably slow down the fetch operation and cause connection broken errors.

### DIFF
--- a/perceval/backends/core/jenkins.py
+++ b/perceval/backends/core/jenkins.py
@@ -34,6 +34,7 @@ from ...client import HttpClient
 
 CATEGORY_BUILD = "build"
 SLEEP_TIME = 10
+DETAIL_DEPTH = 1
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


perceval/backends/core/jenkins.py
```diff
@@ -222,8 +231,7 @@ def get_builds(self, job_name):
             logging.info("Not getting blacklisted job: %s", job_name)
             return
 
-        # depth=2 to get builds details
-        url_build = urijoin(self.base_url, "job", job_name, "api", "json?depth=2")
+        url_build = urijoin(self.base_url, "job", job_name, "api", ("json?depth=%s" % self.detail_depth))
```
sduenas: Why don't put `depth` inside `payload` parameter?

PR #376, opened 2018-04-13
[phabricator] Handle project info in core:edge transactions

valeriocos:
This patch allows to handle different JSON structures (dict and list) of how project data appears in old and new values within the transactions core:edge.

### DIFF
--- a/perceval/backends/core/phabricator.py
+++ b/perceval/backends/core/phabricator.py
@@ -50,7 +50,7 @@ class Phabricator(Backend):
     :param tag: label used to mark the data
     :param archive: archive to store/retrieve items
     """
-    version = '0.9.5'
+    version = '0.10.0'
 
     CATEGORIES = [CATEGORY_TASK]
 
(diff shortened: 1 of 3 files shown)


perceval/backends/core/phabricator.py
```diff
@@ -355,7 +355,9 @@ def __resolve_subsribers_ids(self, lst):
         resolved_lst = []
         for e in lst:
             resolved = e
-            if e.startswith('PHID-PROJ'):
+            if not e:
+                pass
```
sduenas: I think this should be a `continue` and not a `pass`.

perceval/backends/core/phabricator.py
```diff
@@ -364,17 +366,38 @@ def __resolve_subsribers_ids(self, lst):
...
         return resolved_lst
 
-    def __resolve_project_ids(self, dct):
-        if not dct:
-            return dct
+    def __resolve_project_ids(self, obj):
+        if not obj:
+            return obj
+
+        if type(obj) == dict:
```
sduenas: Use `isinstance` instead of type.

perceval/backends/core/phabricator.py
```diff
@@ -364,17 +366,38 @@ def __resolve_subsribers_ids(self, lst):
...
+                project_info = self.__get_or_fetch_project(content['dst'])
+                projects.append(project_info)
+
+        return projects
+
+    def __resolve_project_ids_from_list(self, lst):
+        projects = []
+
+        for e in lst:
+
```
sduenas: Remove this blank line.

PR #377, opened 2018-04-19
[graal] A generic backend to support source code analysis

valeriocos:
This PR proposes a new backend Graal, a Generic Repository AnALyzer. Graal extends the Git backend, thus it fetches the commits from a local Git repository, and enables to add the result of the analysis within the `analysis` attribute of the Perceval item returned. 

To initialize this class, you have to provide the URI repository and a value for `git_path`, where the repository will be cloned. For each target commit (by default all of them), a snapshot of the repository can be created to ease the analysis. The parameter `snapshot` allows to manage the snapshot as a tar object, tar file or folder. In the last two cases, the snapshot is stored in the folder pointed by `snapshots_path`.
Note that you can customize the filter to select commits by redefining the method `_filter_commit(self, commit, files)`. Furthermore, you can plug your analysis by redefining the method `_analyse(self, commit, snapshot)` as well as tweak the item generated by redefining the method `_post(commit)`.

The use of Graal is showcased with a backend to calculate code complexity information (e.g., Cyclomatic Complexity, LOC, commented lines, number of functions) for several popular languages (e.g., Java, Python, JavaScript). If the parameter `functions` is set to true, the analysis can be performed also at the level of functions/methods.


valeriocos:
@sduenas @jgbarah, this can be a good base for discussion, please have a look at the code. Note that I didn't provide tests, since the implementation may change after discussing.

PR #378, opened 2018-04-20
[setup] Fix to let setup.py test run

jgbarah:
The command

```
python setup.py test
```

was not working.

This patch makes it work.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -2,6 +2,6 @@
 description-file = README.md
 
 [flake8]
-exclude = .git, __pycache__, build, dist, docs, docker
+exclude = .git, .eggs, __pycache__, build, dist, docs, docker
 ignore = E129, E402, F841, C901
 max-line-length = 130
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,8 @@ import re
 import sys
 import unittest
 
-from setuptools import setup, Command
+from setuptools import setup
+from setuptools.command.test import test as TestClass
 
 here = os.path.abspath(os.path.dirname(__file__))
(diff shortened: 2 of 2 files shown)


jgbarah:
> I have just a question, should the perceval.egg-info folder be removed after executing the command python setup.py test?

I think it doesn't matter if you're using pip install -e . as a developer, but very likely it is safer to just remove it after the tests are passed...

jgbarah:
>  We already have requirements_tests.py which should work the same way. Do still need that? Do we need both?

I would completely skip requirements in `requirements*`, and stick only to those in setup.py. If the patch works as intended, for travis (and any other testing, for that matter), you just need to run `python setup.py test` and testing requirements should be installed. For development, you can just run `pip install -e .` and you will get every dependency installed as well. If for some reason you still need `requirements.txt`, you can just remove everything in it, and substitute with:

```
 -e .
```

In that case, `pip -r requirements.txt` will just use `setup.py` to get the list of packages to pre-install.

The only problem I see is that with the current patch, if we remove `test_requirements.txt', there is no easy way (AFAIK) to install those test dependencies by calling pip directly. There is a nice solution in the first answer to [pip install test dependencies for tox from setup.py](https://stackoverflow.com/questions/29870629/pip-install-test-dependencies-for-tox-from-setup-py), though. Or we could use the option in the other answer, which in sort is to keep `requirerments*` files and then read them from `setup.py`.

jgbarah:
@sduenas ?

jgbarah:
Now, rebased version uploaded, @sduenas. Please, have a look...

PR #380, opened 2018-04-30
[graal] A generic backend to support source code analysis (faster than #377)

valeriocos:
This PR proposes a new backend Graal, a Generic Repository AnALyzer. Graal extends the Git backend, thus it fetches the commits from a local Git repository, and enables to add the result of the analysis within the analysis attribute of the Perceval item returned.

To initialize this class, you have to provide the local path of a Git repository (URI), a value for `git_path`, where the repository will be mirrored, and the path where a working tree will be created. The working tree is added to the mirror and removed after the analysis is over. The use of a working tree allows to leverage on the command `git checkout` which is faster than `git archive` (used in PR #377).

For each target commit (by default all of them), a checkout version of the repository is created at `worktreepath` to ease the analysis. Note that you can customize the filter to select commits, by redefining the method `_filter_commit(self, commit, files)`. Furthermore, you can plug your analysis by redefining the method `_analyse(self, commit)` as well as tweak the item generated by redefining the method `_post(commit)`.

The use of Graal is showcased with a backend to calculate code complexity information (e.g., Cyclomatic Complexity, LOC, commented lines, number of functions) for several popular languages (e.g., Java, Python, JavaScript). If the parameter functions is set to true, the analysis can be performed also at the level of functions/methods.


valeriocos:
The code of this PR has been moved at https://github.com/valeriocos/graal

valeriocos:
Just for practical reasons. Anyway if this is a problem, I can update the PR and reopen it.

PR #386, opened 2018-05-28
[pipermail] Disable SSL verification

valeriocos:
This code allows to disable SSL verifications for calls to a pipermail server, via the argument `verify` (default value is set to true). Note that if verification is disabled, warnings are returned in the log.

### DIFF
--- a/perceval/backends/core/pipermail.py
+++ b/perceval/backends/core/pipermail.py
@@ -60,16 +60,18 @@ class Pipermail(MBox):
 
     :param url: URL to the Pipermail archiver
     :param dirpath: directory path where the mboxes are stored
+    :param verify: allows to disable SSL verification
     :param tag: label used to mark the data
     :param archive: archive to store/retrieve items
     """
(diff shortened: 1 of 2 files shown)


perceval/backends/core/pipermail.py
```diff
@@ -30,6 +30,8 @@
 import bs4
 import dateutil
 import requests
+import urllib3
+urllib3.disable_warnings()
```
sduenas: Warnings shouldn't be hidden on a library. Let users to do that.

perceval/backends/core/pipermail.py
```diff
@@ -199,7 +201,7 @@ def fetch(self, from_date=DEFAULT_DATETIME):
 
         from_date = datetime_to_utc(from_date)
 
-        r = requests.get(self.url)
+        r = requests.get(self.url, verify=False)
```
sduenas: In my opinion, this should be an optional behaviour and configured passing the value to the class.

PR #387, opened 2018-05-28
[tests] Increase test coverage pipermail

valeriocos:
This code allows to reach 100% line coverage for the pipermail backend.

### DIFF
--- a/tests/test_pipermail.py
+++ b/tests/test_pipermail.py
@@ -49,6 +49,13 @@ def read_file(filename, mode='r'):
     return content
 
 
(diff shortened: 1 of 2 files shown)


tests/test_pipermail.py
```diff
@@ -49,6 +50,14 @@ def read_file(filename, mode='r'):
     return content
 
 
+class PipermailListMocked(PipermailList):
```
sduenas: Do you think this is needed. I would raise an `OSError` exception. It's more clear than use `chmod` to raise that error.

PR #393, opened 2018-06-11
Added PATH environment in git backend

dandanwei:
Update the git backend to allow the sub process which executes the git command to use the same PATH environment variable as the parent process. Without this update it will use /usr/bin/git by default, which could be different from the actual git that is used and defined by PATH.

### DIFF
--- a/perceval/backends/core/git.py
+++ b/perceval/backends/core/git.py
@@ -783,7 +783,8 @@ class GitRepository:
         self.gitenv = {
             'LANG': 'C',
             'PAGER': '',
-            'HOME': os.getenv('HOME', '')
+            'HOME': os.getenv('HOME', ''),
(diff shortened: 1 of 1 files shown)


sduenas:
I'm not sure if it's the best solution for this problem. With your change, tests are broken in Travis CI because `git` command couldn't be found.

dandanwei:
Thanks for the feedback! I will check and come back later.

sduenas:
I've been thinking we can have a special env variable for those cases. For example, something like `PERCEVAL_GIT_PATH`, when this is defined, the child process will use that path to run `git`. Otherwise, it will work as git is configured in the system.

dandanwei:
That is a very good idea! I can come back with that.

PR #401, opened 2018-07-03
[args] Add new argument --json-line

jgbarah:
The new `--json-line` argument produces an output which is a valid JSON line per item. This makes it easier to import data produced by Perceval from scripts.


jgbarah:
> It would be possible to add a test like...

Done.

valeriocos:
Thank you @jgbarah ! Please fix the flake8 error to make the tests pass

jgbarah:
> Please fix the flake8 error to make the tests pass

Done. Sorry for not checking earlier.

tests/test_backend.py
```diff
@@ -824,6 +824,35 @@ def test_run_not_supported_archive(self):
...
+                '--archive-path', self.test_path,
+                '--from-date', '2015-01-01', '--tag', 'test',
+                '--output', self.fout_path, 'http://example.com/',
+                '--json-line']
+
+        cmd = MockedBackendCommand(*args)
+        cmd.run()
+        cmd.outfile.close()
+
+#        items = [item for item in convert_cmd_output_to_json(self.fout_path)]
```
sduenas: Can you remove this line before merging it?

tests/test_backend.py
```diff
@@ -824,6 +824,35 @@ def test_run_not_supported_archive(self):
...
+        with open(self.fout_path) as fout:
+            items = [item for item in fout]
```
sduenas: I think is better to use `fout.readlines()` if we're not to do anything with the lines.

tests/test_backend.py
jgbarah: Done.

tests/test_backend.py
jgbarah: Done.

jgbarah:
I think I addressed all suggestions. Sorry for the delay.

sduenas:
@jgbarah, I think something is wrong because I don't see my comments were addressed.

jgbarah:
Ooops. Sorry @sduenas, my bad. Now, changes should be addressed.

PR #402, opened 2018-07-05
[backends] Add backend for groups.io

valeriocos:
This PR includes a backend to fetch mbox archives available on Groups.io. It inherits part of its functionalities from the Mbox backend, and extend the latter to handle zip-compressed mbox archives.
In order to use the backend, the following parameters are needed:
- group_name: name of a group on Groups.io
- api_token: a token to query the Groups.io API.


valeriocos:
I don't understand why the coverage decreased, since: 
- no lines were removed 
- the groups.io backend is 100% covered
- tests for the new functionalities in mbox and utils were provided 
- new tests for mbox were added to cover no-tested portion of code

README.md
```diff
@@ -190,6 +191,11 @@ $ perceval github elastic logstash --from-date '2016-01-01'
 $ perceval gitlab elastic logstash --from-date '2016-01-01'
 ```
 
+### Groups.io
+```
+$ perceval groupsio 'updates' -t 'abcdefghijk' --from-date '2016-01-01'
```
sduenas: The help output says `groups.io` is the way to call the backend but here you're using `groupsio`.

tests/test_groupsio.py
```diff
@@ -0,0 +1,391 @@
...
+        self.assertEqual(message['data']['Message-ID'], '<1530334100637290090.7026@groups.io>')
+        self.assertEqual(message['origin'], 'https://groups.io/g/acme_group')
+        self.assertEqual(message['uuid'], '787835d4cc343299b8d1a427fa580365bb7feb0d')
+        self.assertEqual(message['updated_on'], 1530334100.0)
+        self.assertEqual(message['category'], 'message')
+        self.assertEqual(message['tag'], 'https://groups.io/g/acme_group')
+
+    @httpretty.activate
+    def test_fetch_empty(self):
+        """Test whether it works when a group does not store any mbox"""
```
sduenas: I think this documentation message is wrong because the code stores the mbox but the mbox is empty.

tests/test_mbox.py
```diff
@@ -50,19 +52,26 @@ def setUpClass(cls):
...
                 mod = gzip
+            else:
+                mod = zipfile
 
-            with open(os.path.join(os.path.dirname(os.path.abspath(__file__)), 'data/mbox/mbox_single.mbox'), 'rb') as f_in:
-                with mod.open(fname, 'wb') as f_out:
-                    shutil.copyfileobj(f_in, f_out)
+            if mod == zipfile:
+                with zipfile.ZipFile(fname, 'w') as f_out:
+                    f_out.write(os.path.join(os.path.dirname(os.path.abspath(__file__)), 'data/mbox/mbox_single.mbox'))
```
sduenas: These lines are too long. Reformat then in with something like:

``` 
f_out.write(os.path.join(os.path.dirname(os.path.abspath(__file__)),
            'data/mbox/mbox_single.mbox'))
```
```	
with open(os.path.join(os.path.dirname(os.path.abspath(__file__)),
         'data/mbox/mbox_single.mbox'), 'rb') as f_in:
```

perceval/backends/core/mbox.py
```diff
@@ -367,6 +368,11 @@ def container(self):
             return bz2.open(self.filepath, mode='rb')
         elif self.compressed_type == 'gz':
             return gzip.open(self.filepath, mode='rb')
+        elif self.compressed_type == "zip":
+            _zip = zipfile.ZipFile(self.filepath)
+            if len(_zip.infolist()) > 1:
+                logger.error("Zip %s contains more than one file, only the first is analyzed", self.filepath)
```
sduenas: I would change `is analyzed` by `is uncompressed`.

README.md
```diff
@@ -20,6 +20,7 @@ are:
     git              Fetch commits from Git
     github           Fetch issues from GitHub
     gitlab           Fetch issues from GitLab
+    groups.io        Fetch messages from Groups.io
```
sduenas: File `bin/perceval` also needs to be updated to include this documentation.

PR #411, opened 2018-07-19
[backends] Add backend for GoogleHits

valeriocos:
This code includes a backend to fetch hits for a given list of keywords from the Google API. In order to use the backend, the only required parameter is a list of keywords.


perceval/backends/core/googlehits.py
```diff
@@ -0,0 +1,251 @@
...
+        of connection problems
+    """
+    version = '0.1.0'
+
+    CATEGORIES = [CATEGORY_HITS]
+
+    def __init__(self, keywords, tag=None, archive=None,
+                 max_retries=MAX_RETRIES, sleep_time=DEFAULT_SLEEP_TIME):
+        # sort keywords to produce always the same hash for the same keywords
+        keywords.sort()
```
sduenas: This sorts the original list. Code must not change inputs when the user doesn't expect it. It would be better to copy the list into `self.keywords` and sort this one.

perceval/backends/core/googlehits.py
```diff
@@ -0,0 +1,251 @@
...
+        super().__init__(GOOGLE_SEARCH_URL, extra_status_forcelist=[429],
+                         sleep_time=sleep_time, max_retries=max_retries,
+                         archive=archive, from_archive=from_archive)
+
+    def hits(self, keywords):
+        """Fetch information about a list of keywords."""
+
+        for keyword in keywords:
+            logger.info("Fetching hits for '%s'", keyword)
+            params = {'q': keyword}
```
sduenas: I'm not sure if this is the behaviour we want. This means we're making a query for each keyword (OR query) instead a single query containing all the keywords (AND query). This contradicts the behaviour we have in `stackexchange` where we make request with all the keywords.

My idea is that if someone wants to make a OR query, she should run perceval one time for each keyword.

perceval/backends/core/googlehits.py
```diff
@@ -0,0 +1,251 @@
...
+
+            # Make the request
+            req = self.fetch(GOOGLE_SEARCH_URL, payload=params)
+
+            # Create the soup and get the desired div
```
sduenas: Is not better to move the parsing code to the backend as we have in other backends?

tests/test_googlehits.py
```diff
@@ -0,0 +1,242 @@
+# -*- coding: utf-8 -*-
```
sduenas: I miss tests checking when there are no results using the given keywords.

PR #412, opened 2018-07-20
[gitlab] Sanitize requests before archiving

valeriocos:
This code removes token information before archiving the GitLab data, thus it enables to read the archive afterwards without the need of passing the token.


valeriocos:
sure @sduenas , I have just splitted it in two (https://github.com/chaoss/grimoirelab-perceval/pull/415)

PR #417, opened 2018-07-21
[doc] Update install from source code section in README.md

valeriocos:
This code updates the information to install perceval from source code. Thus, the new way of installing perceval is: `pip3 install .`

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,12 +88,17 @@ To install from the source code you will need to clone the repository first:
 $ git clone https://github.com/chaoss/grimoirelab-perceval.git
 ```
 
-In this case, [setuptools](http://setuptools.readthedocs.io/en/latest/) package will be required.
-Make sure it is installed before running the next commands:
-
+Then you can execute the following commands:
 ```
 $ pip3 install -r requirements.txt
-$ python3 setup.py install
+$ pip3 install .
+```
+
+In case you are a developer, you should execute the following commands to install Perceval in your working directory (option `-e`) and the packages of requirements_tests.txt.
+```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -93,7 +93,7 @@ Make sure it is installed before running the next commands:
 
 ```
 $ pip3 install -r requirements.txt
-$ python3 setup.py install
+$ pip3 install .
```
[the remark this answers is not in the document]
valeriocos: I modfied the README according to your comment

README.md
```diff
@@ -93,7 +93,7 @@ Make sure it is installed before running the next commands:
 
 ```
 $ pip3 install -r requirements.txt
-$ python3 setup.py install
+$ pip3 install . (-e for developer build)
```
sduenas: I think it will be more clear to describe developer installation in a different paragraph. For instance, developers need to install the dependencies of **requirements_tests.txt**.

jgbarah: I'm sorry, but I don't get it:

* If you want to install from the repo, and you're a developer, instructions should be (AFAIU)

```
pip3 install -e .
```

No need to install `requirements.txt` in this case.

If you're not a developer, it would be better, I think to produce the package and then install it (but I'm not sure why you would like to install it from source if you're not a developer):

```
python3 setup.py sdist
pip3 install dist/package_name
```

Again, no need for installing requirements.

Or maybe I'm missing something?

jgbarah:
But then, why do we need to install `requirements.txt` in addition to running `pip install`?

PR #424, opened 2018-08-27
[googlehits] Modify item id generation

valeriocos:
This code modifies the way how ids for google hits items are generated. In order to track the evolution of a set of keywords, the id should contain the time when the search was performed.

### DIFF
--- a/perceval/backends/core/googlehits.py
+++ b/perceval/backends/core/googlehits.py
@@ -21,7 +21,6 @@
 #
 
 import bs4
-import hashlib
 import logging
 import re
 
@@ -29,7 +28,8 @@ from grimoirelab_toolkit.datetime import datetime_utcnow
 
 from ...backend import (Backend,
                         BackendCommand,
-                        BackendCommandArgumentParser)
+                        BackendCommandArgumentParser,
+                        uuid)
 from ...client import HttpClient
 from ...errors import BackendError
 
@@ -59,7 +59,7 @@ class GoogleHits(Backend):
     :param sleep_time: time to sleep in case
         of connection problems
     """
-    version = '0.1.0'
+    version = '0.2.0'
 
     CATEGORIES = [CATEGORY_HITS]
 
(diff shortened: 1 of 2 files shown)


tests/test_googlehits.py
```diff
@@ -123,6 +123,13 @@ def test_has_resuming(self):
 
         self.assertEqual(GoogleHits.has_resuming(), True)
 
+    # def test_fetch_Live(self):
```
sduenas: Remove this patch from the code, please.

perceval/backends/core/googlehits.py
```diff
@@ -130,7 +130,7 @@ def has_resuming(cls):
     def metadata_id(item):
         """Extracts the identifier from a GoogleHit item."""
 
-        return item['id']
+        return item['id'] + '-' + str(item['fetched_on'])
```
sduenas: Is this the way we do this in other backends? Do we concat fields using a `- `?

valeriocos: we generally have just the `id` without any concatenation. I can modify the googlehits `__uuid` method to include the `fetched_on` value. What do you think?

valeriocos: I have modifed the PR, now the method `__uuid` takes care of joining the `fetched_on` attribute.

perceval/backends/core/googlehits.py
```diff
@@ -174,34 +174,30 @@ def __parse_hits(self, hit_raw):
...
-            'keywords': self.keywords,
-            'type': 'googleSearchHits'
-        }
+        hits_json['hits'] = hits
 
         return hits_json
 
     @staticmethod
-    def __uuid(*args):
+    def __uuid(args, fetched_on):
```
sduenas: Why do you do this? `*args` y a list of arguments. Give `fetched_on` on that list of arguments.

perceval/backends/core/googlehits.py
```diff
@@ -174,34 +174,30 @@ def __parse_hits(self, hit_raw):
...
         """Generate a UUID based on the given parameters."""
 
+        args.append(str(fetched_on))
         s = '-'.join(args)
```
sduenas: On `supybot` backend we don't use any characters to concatenate strings when we generate the `id`. Not sure what it's better.

tests/test_googlehits.py
```diff
@@ -151,10 +151,13 @@ def test_fetch(self, mock_utcnow):
 
         self.assertEqual(len(items), 1)
         item = items[0]
+
+        import json
```
sduenas: Don't import this in the middle of the code, please. Moreover, don't print anything when tests are run.

perceval/backends/core/googlehits.py
valeriocos: I believe it's clearer this way

perceval/backends/core/googlehits.py
valeriocos: I took as reference https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/backend.py#L444

perceval/backends/core/googlehits.py
valeriocos: I believe this way is clearer. However I changed the way of generating the uuid as you suggested. I used the `uuid` method in [backend.py](https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/backend.py#L444)

perceval/backends/core/googlehits.py
valeriocos: this is gone, now uuids rely on the method `uuid` in backend.py

PR #430, opened 2018-09-04
remove unused and not working config_file option

### LINKED ISSUE #428: Config option in perceval: should we remove it ?
After i did some research on the configuration option `-c` for perceval to me look like that this functionality is not working well + undocumented.
( see pr on wip: https://github.com/chaoss/grimoirelab-perceval/pull/427)

What is not working a part of documentation:

if you see here `defaults` var https://github.com/chaoss/grimoirelab-perceval/blob/master/bin/perceval#L99
this var is not used for the backend meta-class.. i don"t see any method that take advantage of that data ( i tried it out with this  cfg file:)

``` 
[github]
owner = "chaoss"
repository = "grimoirelab-perceval"
```
and calling
```
perceval -c perceval.cfg 
```
but perceval was complaining  that repository was missing

Ok we can implement/fix that feature but..  : 

To me since we have https://github.com/chaoss/grimoirelab-kingarthur which is the scheduler working with files for scheduling various instances of perceval, i don"t see this option `-c` as really needed.

Additionally it seems to me to be a ghost feature which we can safely remove.
https://github.com/chaoss/grimoirelab-perceval/commit/589bc515db1632959f6edcb0c4bc2fe43ee51b74
(just looking the initial implementation was more kind of debugging/printing args)

Personally i would vote to remove this config parameter for perceval since for that there is king-arthur ( and we spare an additional option that can led to complexity), but i just opened the issues to discuss it with you

MalloZup:
remove unused and not working config_file option for perceval cli

fix #428 
CC
@sduenas @valeriocos

### DIFF
--- a/bin/perceval
+++ b/bin/perceval
@@ -22,9 +22,7 @@
 #
 
 import argparse
-import configparser
 import logging
-import os.path
 import sys
 
 import perceval
@@ -32,7 +30,7 @@ import perceval.backend
 import perceval.backends.core
 
 PERCEVAL_USAGE_MSG = \
-"""%(prog)s [-c <file>] [-g] <backend> [<args>] | --help | --version"""
+"""%(prog)s [-g] <backend> [<args>] | --help | --version"""
 
 PERCEVAL_DESC_MSG = \
 """Send Sir Perceval on a quest to retrieve and gather data from software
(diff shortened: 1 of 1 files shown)


sduenas:
Thanks @MalloZup.

I merged your PR rebasing it to master.

MalloZup:
thanks @sduenas :+1:

PR #435, opened 2018-09-13
[gitlab] Add merge_request category

valeriocos:
This code enhances the GitLab backend to extract merge requests data, which includes comments and reactions (emojis) on merge requests, commits and changes. Note that, due to recent changes on the GitLab API, the way of filtering issues and merge requests has changed accordingly.


valeriocos:
we could add a parameter to include the diff data, by default diffs are excluded. what do you think?

wrt the error you mention, I look into it.

sduenas:
You know my big concern about having first-citizen items. In my opinion, items of any given category should have the same structure. I think it would be better to remove that field from the items. What do you think?

Issue #436, opened 2018-09-18
[mediawiki] Duplicate entry when inserting into the archive


sduenas:
When retrieving data using `mediawiki` backend, I got this error. It raises an error because it founds a duplicate entry in the archive.

```
$ perceval mediawiki https://www.mediawiki.org/w/ --from-date 2018-09-09 > /dev/null
[2018-09-18 14:38:33,523] - Sir Perceval is on his quest.
[2018-09-18 14:38:34,003] - MediaWiki version: [1, 32]
[2018-09-18 14:38:34,003] - Looking for pages at url 'https://www.mediawiki.org/w/'
Traceback (most recent call last):
  File "/home/sduenas/devel/grimoirelab/perceval/perceval/archive.py", line 181, in store
    payload_dump, headers_dump, data_dump))
sqlite3.IntegrityError: UNIQUE constraint failed: archive.hashcode

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/sduenas/devel/grimoirelab/perceval/perceval/backend.py", line 385, in run
    for item in items:
  File "/home/sduenas/devel/grimoirelab/perceval/perceval/backend.py", line 498, in fetch
    raise e
  File "/home/sduenas/devel/grimoirelab/perceval/perceval/backend.py", line 492, in fetch
    for item in items:
  File "/home/sduenas/devel/grimoirelab/perceval/perceval/backend.py", line 127, in fetch
    for item in self.fetch_items(category, **kwargs):
  File "/home/sduenas/devel/grimoirelab/perceval/perceval/backends/core/mediawiki.py", line 131, in fetch_items
    for page_reviews in fetcher:
  File "/home/sduenas/devel/grimoirelab/perceval/perceval/backends/core/mediawiki.py", line 276, in fetch_incremental_changes
    page_reviews = self.__get_page_reviews(page)
  File "/home/sduenas/devel/grimoirelab/perceval/perceval/backends/core/mediawiki.py", line 238, in __get_page_reviews
    revisions_raw = self.client.get_revisions(page['title'])
  File "/home/sduenas/devel/grimoirelab/perceval/perceval/backends/core/mediawiki.py", line 454, in get_revisions
    return self.call(params)
  File "/home/sduenas/devel/grimoirelab/perceval/perceval/backends/core/mediawiki.py", line 367, in call
    req = self.fetch(self.base_url, payload=params)
[...]

PR #437, opened 2018-09-26
[mediawiki] Optimize number of calls to API

valeriocos:
This code targets #436, thus it optmizes the number of calls to fetch revision info, in particular if a page doesn't have any revision, the call is not performed. Furthermore, this code avoids errors when creating the archive (which doesn't allow to store duplicated API calls).

### DIFF
--- a/perceval/backends/core/mediawiki.py
+++ b/perceval/backends/core/mediawiki.py
@@ -211,6 +211,7 @@ class MediaWiki(Backend):
         logger.info("Looking for pages at url '%s'", self.url)
 
         npages = 0  # number of pages processed
+        tpages = 0  # number of total pages
         pages_done = []  # pages already retrieved in reviews API
 
         namespaces_contents = self.__get_namespaces_contents()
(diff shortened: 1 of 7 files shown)


tests/test_mediawiki.py
```diff
@@ -200,7 +200,7 @@ def _test_fetch_version(self, version, mock_utcnow, from_date=None, reviews_api=
         elif version == "1.23" or not reviews_api:
             if not from_date:
                 # 2 pages per each of the 5 name spaces
-                self.assertEqual(len(pages), 10)
+                self.assertEqual(len(pages), 2)
```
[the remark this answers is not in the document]
valeriocos: Me neither, the PR wasn't ready to review, so currently I cannot answer your question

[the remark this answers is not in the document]
valeriocos: The pages have the same IDs, thus instead of issuing 10 calls, only 2 are performed.

valeriocos:
The method `get_recent_pages` returns the recent changes over pages (every change contains different IDs such as `pageid`, `revid`, `old_revid`), thus the same `pageid` can appear more than once. 

Later, for each page retrieved, all revisions are collected based on the name of the page, however now a check over the `revid` is performed to avoid issuing requests for pages that don't have revisions, thus limiting the global number of calls to the mediawiki API.

perceval/backends/core/mediawiki.py
```diff
@@ -251,32 +260,44 @@ def __fetch_pre1_27(self, from_date=None):
...
+                rccontinue = None
                 if 'query-continue' in data_json:
                     # < 1.27
                     rccontinue = data_json['query-continue']['recentchanges']['rccontinue']
                 elif 'continue' in data_json:
                     # >= 1.27
                     rccontinue = data_json['continue']['rccontinue']
-                else:
-                    rccontinue = None
+
```
sduenas: I would keep the `else` block here. I think is more clear and if we have to change something else, we won't break anything.

perceval/backends/core/mediawiki.py
```diff
@@ -303,7 +325,17 @@ def fetch_all_pages(namespaces_contents):
...
                     for page in pages_json:
-                        yield self.__get_page_reviews(page)
+                        if page['pageid'] in pages_done:
+                            # The page was already returned for previous revisions
+                            continue
+
+                        pages_done.append(page['pageid'])
+                        page_reviews = self.__get_page_reviews(page)
+                        if not page_reviews:
+                            # Page without reviews are not managed
```
sduenas: I think we need to add a warning message to inform this page was skipped. Like in https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/backends/core/mbox.py#L213

tests/test_mediawiki.py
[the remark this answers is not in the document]
sduenas: Understood.

tests/test_mediawiki.py
```diff
@@ -212,13 +213,24 @@ class TestMediaWikiBackend_1_23(TestMediaWikiBackend):
     """MediaWiki backend tests for MediaWiki 1.23 version"""
 
     def test_fetch(self):
-        self._test_fetch_version("1.23")
-        self._test_fetch_version("1.23", reviews_api=True)
+        with self.assertLogs(logger, level='WARNING') as cm:
```
sduenas: The usage of `cm` is not clear. What's the meaning of cm? Context Manager? And why do you have to access to the message in this way: `cm[-1][2]`?

valeriocos: cm (which stands for context manager) is a common variable used in combination with `assertLogs` (https://docs.python.org/3.4/library/unittest.html#unittest.TestCase.assertLogs). 
The log messages can be accessed in different ways, `cm[-1][-2]` is one of the possible ones (which is used also in other tests in Perceval). I use the index `[-2]` to access a specific message in the logs.

PR #438, opened 2018-09-27
[gitlab] Fix typo initiazaling common MRs fields

sduenas:
Merge Requests store `versions_data` field and not `commits_data`. This patch fixes this typo.


sduenas:
> LGTM, could you also update the second line of the comment at line 250: ...thus we inflate it with other data (e.g., notes, emojis, ---> commits <---) ?

Done.

PR #444, opened 2018-10-14
[groupsio] Fetch archives from group-specific domains

valeriocos:
This code allows the groupsio backend to fetch archives of organizations having their own domains (e.g., lists.cloudfoundry.org). Given a group name and a token, the corresponding group id is searched among the list of subscriptions of the token. If the group id is not found, an error is thrown.
Groupsio test code has been changed accordingly.


perceval/backends/core/groupsio.py
```diff
@@ -162,18 +170,53 @@ def fetch(self):
...
+        :returns: an iterator of subscriptions
+        """
+        url = urijoin(GROUPSIO_API_URL, self.GET_SUBSCRIPTIONS)
+        logger.debug("Get groupsio paginated subscriptions from " + url)
+
+        payload = {
+            "limit": per_page
+        }
+
+        r = self.__fetch(url, payload)
```
sduenas: This whole logic can be merged into a single `while` loop instead of repeating the code twice.

PR #454, opened 2018-11-06
[git] Include tags and branches information

valeriocos:
This code enhances the git backend to include the names of tags and branches where a commit appears. Note that the performance of the backend is affected, 4 secs are now needed to parse 200 commits, against 1 sec that was required before.


valeriocos:
> I guess this doesn't work incrementally, does it?

No, it doesn't work incrementally.  

> Since this has a severe performance penalty, could it have an option to disable it when needed? 

I have just added a commit to make it optional

jgbarah:
> No, it doesn't work incrementally.

I think it would be convenient to state that crystal clear in the documentation, since that is a major difference with the regular working of the backends.

> I have just added a commit to make it optional

Great! Thanks.

Polaris000:
@valeriocos, why was this not merged?
Was it due to it not properly supporting increment?

valeriocos:
Hi @Polaris000 , it doesn't work incrementally and thus it is not really useful in production environments. However for one-shot analysis it works fine. If this case fits in your scenario feel free to reuse the code.

The problems with branches (and also tags) are:
- (i) they are pointers to commits and they can change at any moment (a merge, a rebase, etc.)
- (ii) the information retrieved by Perceval is just about commits, thus changes at branch level are not collected
- (iii) getting branch changes from Git is not trivial.
 
The solution GrimoireLab offers is to update the branch information for all commits stored in the ElasticSearch database for each enrichement phase. Please have a look at the [code](https://github.com/chaoss/grimoirelab-elk/blob/master/grimoire_elk/enriched/git.py#L762) and don't hesitate to write in case of any questions/doubts :)

Polaris000:
Thanks for the prompt reply as well as the clearly stated reasons, @valeriocos! 
Will ask in case of any queries.

PR #455, opened 2018-11-06
Updating instructions for querying stackexchange in README

filmaj:
`--token` doesnt work, but `-t` does.


valeriocos:
thank you @filmaj for the PR. You are right, `--token` doesn't work, 
the correct names are `--api-token`or `-t`

PR #463, opened 2018-12-04
[gitlab] Add retry on 503 HTTP errors

valeriocos:
This code allows the backend to reissue petititions to the GitLab API in case of 503 HTTP errors. This fix is needed to avoid the fetching process to be stopped due to upstream server problems.


chupy35:
if i got a 500 server error, is there a way of continue from that part if i am putting the output in a file?

valeriocos:
Hi @chupy35 , you can use the `from_date` param. It will retrieve the the issues/merge requests updated after a given date. If you are storing the data, just use as `from_date` the `updated_at` value of the last item before the 500 error.

PR #464, opened 2018-12-11
[discourse] Add params `sleep_time` and ` max_retries`

valeriocos:
This code enhances the Discourse backend, client and commandline with the params `sleep_time` and `max_retries`, which slow down the number of petititions issued to the API, and thus limit possible 429 HTTP errors. Tests have been changed accordingly.

### DIFF
--- a/perceval/backends/core/discourse.py
+++ b/perceval/backends/core/discourse.py
@@ -35,6 +35,9 @@ from ...client import HttpClient
 from ...utils import DEFAULT_DATETIME
 
 
+DEFAULT_SLEEP_TIME = 5
+MAX_RETRIES = 10
+
 CATEGORY_TOPIC = "topic"
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


perceval/backends/core/discourse.py
```diff
@@ -267,11 +280,18 @@ class DiscourseClient(HttpClient):
 
     :param url: URL of the Discourse site
     :param api_key: Discourse API access token
-    :param archive: an archive to store/read fetched data
+    :param sleep_time: time to sleep in case
+        of connection problems
+    :param max_retries: number of max retries to a data source
+        before raising a RetryError exception
+    :param archive: collect issues already retrieved from an archive
```
sduenas: This parameter is repeated in the description below. I guess the first one is incorrect.

perceval/backends/core/discourse.py
```diff
@@ -51,17 +54,25 @@ class Discourse(Backend):
     :param api_token: Discourse API access token
     :param tag: label used to mark the data
     :param archive: archive to store/retrieve items
+    :param max_retries: number of max retries to a data source
+        before raising a RetryError exception
+    :param sleep_time: time to sleep in case
```
sduenas: Can we define whether this is milisecs, seconds, minutes, etc? Same below in client docs.

valeriocos: I'm going to address this in another PR since the description of `sleep_time` is repeated in several backends.

valeriocos:
The PR to improve the docstring for the `sleep_time` param is available at:
https://github.com/chaoss/grimoirelab-perceval/pull/491

please @sduenas have a look at it

valeriocos:
Done, do you think the `max_retries` should be lowered too ?

PR #472, opened 2019-01-17
[github] Handle upstream server connection problems

valeriocos:
This code limits the connection problems with the github api. In a nutshell, the github backend puts now in place retrying mechanisms for the following http errors 403, 500, 502, 503. Note that this solution
doesn't guarantee to completely solve the connection problems, since the retry is executed a limited number of times (param `max_retries`), before raising an exception.


perceval/backends/core/github.py
```diff
@@ -483,7 +483,8 @@ def __init__(self, owner, repository, token,
             base_url = GITHUB_API_URL
 
         super().__init__(base_url, sleep_time=sleep_time, max_retries=max_retries,
-                         extra_headers=self._set_extra_headers(), archive=archive, from_archive=from_archive)
+                         extra_headers=self._set_extra_headers(), extra_status_forcelist=[403, 500, 502, 503],
```
sduenas: I think is better to move the list of status to a constant class. Therefore, it would be easier to update the list when developing or when using the client in some specific code.

valeriocos: I agree, should we do it also for the other backends that use `extra_status_forcelist`?
- [googlehits](https://github.com/chaoss/grimoirelab-perceval/blob/45dfec19322502762de6a71230153c6f54990724/perceval/backends/core/googlehits.py#L213)
- [jenkins](https://github.com/chaoss/grimoirelab-perceval/blob/44e442cff6acd05743542611070ec03a4a538c7d/perceval/backends/core/jenkins.py#L214)
- [meetup](https://github.com/chaoss/grimoirelab-perceval/blob/67919f659e4a80b62cd88f4fd720b462898b8492/perceval/backends/core/meetup.py#L338)
- [phabricator](https://github.com/chaoss/grimoirelab-perceval/blob/4196868e0f4b96cd8d8761b05ee87fb71d071dfa/perceval/backends/core/phabricator.py#L479)

valeriocos: I have just added some commits to align your comment with the rest of the code. I can move them to another PR.

PR #475, opened 2019-01-18
[git] Allow fetch without update repo

valeriocos:
This PR allows to fetch the commits of a repo without performing any update against the target repo. This new feature is enable by setting the fetch param `no_update` to true (by default is disabled). Tests have been created accordingly.

This change is needed to allow ELK to detect the commits that have been deleted in the repo and remove them from the target index (see https://github.com/chaoss/grimoirelab-elk/pull/538)

### DIFF
--- a/perceval/backends/core/git.py
+++ b/perceval/backends/core/git.py
@@ -64,7 +64,7 @@ class Git(Backend):
     :raises RepositoryError: raised when there was an error cloning or
         updating the repository.
     """
-    version = '0.10.2'
+    version = '0.11.0'
 
     CATEGORIES = [CATEGORY_COMMIT]
 
@@ -76,7 +76,7 @@ class Git(Backend):
         self.gitpath = gitpath
 
     def fetch(self, category=CATEGORY_COMMIT, from_date=DEFAULT_DATETIME, to_date=DEFAULT_LAST_DATETIME,
-              branches=None, latest_items=False):
+              branches=None, latest_items=False, no_update=False):
         """Fetch commits.
 
         The method retrieves from a Git repository or a log file
@@ -94,6 +94,9 @@ class Git(Backend):
         The parameter `latest_items` returns only those commits which
         are new since the last time this method was called.
 
+        The parameter `no_update` returns all commits without performing
+        an update of the repository before.
+
         Take into account that `from_date` and `branches` are ignored
         when the commits are fetched from a Git log file or when
         `latest_items` flag is set.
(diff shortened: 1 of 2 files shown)


perceval/backends/core/git.py
```diff
@@ -348,6 +355,9 @@ def setup_cmd_parser():
         group.add_argument('--latest-items', dest='latest_items',
                            action='store_true',
                            help="Fetch latest commits added to the repository")
+        group.add_argument('--no-update', dest='no_update',
```
sduenas: Probably, `--latest-items` and `--no-update` should be mutually exclusive. To use both at the same time doesn't make sense.

tests/test_git.py
```diff
@@ -584,6 +584,86 @@ def test_fetch_latest_items_from_empty_repository(self):
 
         shutil.rmtree(new_path)
 
+    def test_fetch_no_update(self):
+        """Test whether all commits are fetched from a Git repository"""
```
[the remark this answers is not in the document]
valeriocos: why the comment is wrong?

valeriocos:
@sduenas should I address your comments, rework the PR to implement and use `git rev-list` (see last commit) or you prefer to have two PRs one for the param `update` and the other one with the command `git rev-list`? Note that also in the related PR in ELK I added a [commit](https://github.com/chaoss/grimoirelab-elk/pull/538) to discuss which solution was better

tests/test_git.py
```diff
@@ -758,6 +835,15 @@ def test_setup_cmd_parser(self):
         self.assertEqual(parsed_args.git_path, '/tmp/gitpath')
         self.assertEqual(parsed_args.uri, 'http://example.com/')
         self.assertEqual(parsed_args.branches, ['master', 'testing'])
+        self.assertFalse(parsed_args.no_update)
```
sduenas: I would put the code below in a new tests. It's hard to understand why this is failing. It's failing because `--no-update` and `--latest-items` are mutually exclusive. Having this in a test named like `test_mutual_exclusive_update` or something like that will make live easier. Someone would read the test and will know what's going on.

tests/test_git.py
```diff
@@ -584,6 +584,81 @@ def test_fetch_latest_items_from_empty_repository(self):
 
         shutil.rmtree(new_path)
 
+    def test_fetch_no_update(self):
+        """Test whether the fetch method properly works when the no update flag is set"""
```
sduenas: I would reword it like "Test whether new commits aren't fetched when no_update flag is set"

PR #478, opened 2019-01-22
[git] Add rev-list command to GitRepository class

valeriocos:
This code adds support for the `rev-list` command, thus the list of commit hashes is retrieved from the repo without triggering any update. The rev-list is executed with the following params by default:
`git rev-list --topo-order --branches --tags --remotes=origin`.
However, using the param `branches`, it is possible to analysis a set of branches.

### DIFF
--- a/tests/test_git.py
+++ b/tests/test_git.py
@@ -1569,6 +1569,45 @@ class TestGitRepository(TestCaseGit):
 
         shutil.rmtree(new_path)
 
+    def test_rev_list(self):
+        """Test rev-list command"""
+
+        new_path = os.path.join(self.tmp_path, 'newgit')
+
+        repo = GitRepository.clone(self.git_path, new_path)
+        gitrev = repo.rev_list()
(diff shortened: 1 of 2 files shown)


tests/test_git.py
```diff
@@ -1569,6 +1569,58 @@ def test_sync_from_repos_without_refs(self):
...
+
+        shutil.rmtree(new_path)
+
+    def test_rev_list_empty(self):
+        """Test if no line is returned when the log is empty"""
+
+        new_path = os.path.join(self.tmp_path, 'newgit')
+
+        repo = GitRepository.clone(self.git_path, new_path)
+        gitlog = repo.log(from_date=datetime.datetime(2020, 1, 1, 1, 1, 1))
```
sduenas: I think the test is wrong. It isn't checking the output of `rev-list`.

valeriocos: sorry I forgot to delete that test

PR #479, opened 2019-01-22
[git] Redefine `create_git_repository` method as class method

valeriocos:
This code redefines the method `create_git_repository` as a class method, thus allowing to call it without initializing the class. Tests have been provided accordingly.


sduenas:
I won't accept this feature. A repository shouldn't be created using the backend class. If some library needs that, it should call directly to `GitRepository`.

Sorry because I explained myself wrong yesterday.

PR #484, opened 2019-02-20
[git] Add for-each-ref command to GitRepository class

valeriocos:
This PR adds support for the `for-each-ref` command, which iterates over all the references and output information. The command is executed with the param `--contains` that lists the references containing a given commit.
Test have been added accordingly.

### DIFF
--- a/perceval/backends/core/git.py
+++ b/perceval/backends/core/git.py
@@ -64,7 +64,7 @@ class Git(Backend):
     :raises RepositoryError: raised when there was an error cloning or
         updating the repository.
     """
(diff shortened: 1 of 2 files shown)


valeriocos:
This PR should be reviewed together with https://github.com/chaoss/grimoirelab-elk/pull/559

sduenas:
I don't like the idea of adding functionality to clients or backends that is not directly used by Perceval. In this case, this will be only used by ELK. I think it would be better if GELK subclasses Git client and adds this functionality.

Issue #486, opened 2019-02-28
Add ancestors to Confluence backend


maoo:
In order to establish whether a Confluence page (whose activity gets tracked by the Confluence Perceval backend), I've been working on enriching the Confluence backend with `ancestors` (see https://developer.atlassian.com/server/confluence/expansions-in-the-rest-api/).

The ancestors expansion feature should be disabled by default and enabled via optional parameter.

A Pull Request is on the way, with link to this issue, eager to hear your feedback.

PR #487, opened 2019-02-28
Adding page ancestors to confluence backend

maoo:
See issue #486

### DIFF
--- a/perceval/backends/core/confluence.py
+++ b/perceval/backends/core/confluence.py
@@ -18,6 +18,7 @@
 #
 # Authors:
 #     Santiago Dueñas <sduenas@bitergia.com>
+#     Maurizio Pillitu <maoo@apache.org>
 #
 
 import logging
@@ -52,7 +53,7 @@ class Confluence(Backend):
     :param tag: label used to mark the data
     :param archive: archive to store/retrieve items
     """
-    version = '0.9.2'
+    version = '0.10.0'
 
     CATEGORIES = [CATEGORY_HISTORICAL_CONTENT]
 
@@ -86,7 +87,10 @@ class Confluence(Backend):
 
         from_date = datetime_to_utc(from_date)
 
-        kwargs = {'from_date': from_date}
+        kwargs = {
+            'from_date': from_date
+        }
+
         items = super().fetch(category, **kwargs)
 
         return items
@@ -118,6 +122,8 @@ class Confluence(Backend):
 
             for hc in hcs:
                 hc['content_url'] = content_url
+                hc['ancestors'] = content.get('ancestors', [])
+
                 yield hc
                 nhcs += 1
 
@@ -309,6 +315,7 @@ class ConfluenceClient(HttpClient):
     PSTART = 'start'
     PSTATUS = 'status'
     PVERSION = 'version'
+    PANCESTORS = 'ancestors'
 
     # Common values
     VCQL = "lastModified>='%(date)s' order by lastModified"
@@ -340,7 +347,8 @@ class ConfluenceClient(HttpClient):
         # Set parameters
         params = {
             self.PCQL: cql,
-            self.PLIMIT: max_contents
+            self.PLIMIT: max_contents,
+            self.PEXPAND: self.PANCESTORS
         }
 
         if offset:
--- a/tests/data/confluence/confluence_contents.json
+++ b/tests/data/confluence/confluence_contents.json
@@ -1,4 +1,33 @@
 {
+    "ancestors": [
+        {
+            "id": "128548867",
+            "type": "page",
+            "status": "current",
+            "title": "Title 1",
+            "_links" : {
+                "webui" : "/spaces/TEST/title1"
+            }
+        },
+        {
+            "id": "167848895",
+            "type": "page",
+            "status": "current",
+            "title": "Title 2",
+            "_links" : {
+                "webui" : "/spaces/TEST/title2"
+            }
+        },
+        {
+            "id": "128548921",
+            "type": "page",
+            "status": "current",
+            "title": "Title 3",
+            "_links" : {
+                "webui" : "/spaces/TEST/title3"
+            }
+        }
+    ],
     "_links": {
         "base": "http://example.com",
         "context": "",
--- a/tests/test_confluence.py
+++ b/tests/test_confluence.py
@@ -190,7 +190,8 @@ class TestConfluenceBackend(unittest.TestCase):
         expected = [
             {
                 'cql': ["lastModified>='1970-01-01 00:00' order by lastModified"],
-                'limit': ['200']
+                'limit': ['200'],
+                'expand': ['ancestors']
             },
             {
                 'cql': ["lastModified>='1970-01-01 00:00' order by lastModified"],
@@ -264,7 +265,8 @@ class TestConfluenceBackend(unittest.TestCase):
         expected = [
             {
                 'cql': ["lastModified>='2016-06-16 00:00' order by lastModified"],
-                'limit': ['200']
+                'limit': ['200'],
+                'expand': ['ancestors']
             },
             {
                 # Hardcoded in JSON dataset
@@ -337,7 +339,8 @@ class TestConfluenceBackend(unittest.TestCase):
         expected = [
             {
                 'cql': ["lastModified>='1970-01-01 00:00' order by lastModified"],
-                'limit': ['200']
+                'limit': ['200'],
+                'expand': ['ancestors']
             },
             {
                 'cql': ["lastModified>='1970-01-01 00:00' order by lastModified"],
@@ -377,7 +380,8 @@ class TestConfluenceBackend(unittest.TestCase):
         # Check requests
         expected = {
             'cql': ["lastModified>='2016-07-08 00:00' order by lastModified"],
-            'limit': ['200']
+            'limit': ['200'],
+            'expand': ['ancestors']
         }
 
         self.assertEqual(len(http_requests), 1)
@@ -519,7 +523,8 @@ class TestConfluenceClient(unittest.TestCase):
         expected = {
             'cql': ["lastModified>='2016-07-08 00:00' order by lastModified"],
             'start': ['10'],
-            'limit': ['2']
+            'limit': ['2'],
+            'expand': ['ancestors']
         }
 
         self.assertEqual(len(http_requests), 1)
@@ -541,7 +546,8 @@ class TestConfluenceClient(unittest.TestCase):
         expected = [
             {
                 'cql': ["lastModified>='1970-01-01 00:00' order by lastModified"],
-                'limit': ['2']
+                'limit': ['2'],
+                'expand': ['ancestors']
             },
             {
                 'cql': ["lastModified>='1970-01-01 00:00' order by lastModified"],


perceval/backends/core/confluence.py
```diff
@@ -91,6 +92,16 @@ def fetch(self, category=CATEGORY_HISTORICAL_CONTENT, from_date=DEFAULT_DATETIME
 
         return items
 
+    @staticmethod
```
valeriocos: Why static?

perceval/backends/core/confluence.py
```diff
@@ -56,11 +56,12 @@ class Confluence(Backend):
 
     CATEGORIES = [CATEGORY_HISTORICAL_CONTENT]
 
-    def __init__(self, url, tag=None, archive=None):
+    def __init__(self, url, add_ancestors, tag=None, archive=None):
         origin = url
 
         super().__init__(origin, tag=tag, archive=archive)
         self.url = url
+        self.add_ancestors = add_ancestors
```
valeriocos: The backend params generally include global settings (url, api token, username, etc.), specific things should go to the fetch params (e.g., `latest_items` in the git backend https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/backends/core/git.py#L78)

perceval/backends/core/confluence.py
```diff
@@ -215,7 +227,7 @@ def parse_historical_content(raw_json):
     def _init_client(self, from_archive=False):
         """Init client"""
 
-        return ConfluenceClient(self.url, archive=self.archive, from_archive=from_archive)
+        return ConfluenceClient(self.url, self.add_ancestors, archive=self.archive, from_archive=from_archive)
```
valeriocos: in order to ensure that the data returned by Perceval is consistent, it's better to include the ancestors info as something that is always returned by Perceval. What do you think about adding a method (`ancestors`) to the client to retrieve the ancestor pages ?

perceval/backends/core/confluence.py
```diff
@@ -279,6 +291,12 @@ def setup_cmd_parser():
         parser.parser.add_argument('url',
                                    help="URL of the Confluence server")
 
+        # Optional arguments
```
valeriocos: this code is not needed if the ancestor info is included by default in the data returned by the backend

perceval/backends/core/confluence.py
```diff
@@ -279,6 +291,12 @@ def setup_cmd_parser():
...
+        group = parser.parser.add_argument_group('Confluence page options')
+        group.add_argument('--add_ancestors', dest='add_ancestors',
+                           nargs='+', type=bool, default=False,
```
valeriocos: why `nargs='+'` ?

perceval/backends/core/confluence.py
```diff
@@ -332,6 +352,8 @@ def contents(self, from_date=DEFAULT_DATETIME,
         :param limit: maximum number of contents to fetch per request
         """
         resource = self.RCONTENTS + '/' + self.MSEARCH
+        if self.add_ancestors:
```
valeriocos: I would move this code to another method (e.g., `ancestors`) to improve the readbility of the code. What do you think ?

perceval/backends/core/confluence.py
```diff
@@ -332,6 +352,8 @@ def contents(self, from_date=DEFAULT_DATETIME,
...
+            resource = resource + "?expand=ancestors"
```
valeriocos: instead of hard-coding `?expand=ancestors`, you could leverage on variables and reference them in the code (as similarly done in other cases such as `RCONTENTS`, `PSTATUS` and `PVERSION`)

perceval/backends/core/confluence.py
```diff
@@ -91,6 +92,16 @@ def fetch(self, category=CATEGORY_HISTORICAL_CONTENT, from_date=DEFAULT_DATETIME
 
         return items
 
+    @staticmethod
+    def add_ancestors(url_prefix, item, content):
+        ancestors = []
+        if ('ancestors' in content) and (content['ancestors']):
+            for ancestor in content['ancestors']:
+                ancestors.append(url_prefix + ancestor['_links']['webui'])
```
valeriocos: generally Perceval tries to have a one-to-one correspondence with the data in upstream, thus we avoid manipulating the data (something that is done by ELK during the data loading or enrichment)

perceval/backends/core/confluence.py
```diff
@@ -91,6 +92,16 @@ def fetch(self, category=CATEGORY_HISTORICAL_CONTENT, from_date=DEFAULT_DATETIME
 
         return items
 
+    @staticmethod
+    def add_ancestors(url_prefix, item, content):
```
valeriocos: the logic of `add_ancestor` could be simplified. For instance, at first you could check if the ancestors don't exist and return an empty list. Then, the next statements could tackle the case when the ancestors do exist.

perceval/backends/core/confluence.py
maoo: No particular reason, other than I didn't need anything from `self` object. If you prefer to have it non-static I can change it.

perceval/backends/core/confluence.py
maoo: Yep, makes sense, I'll change that

perceval/backends/core/confluence.py
maoo: No idea? :-) I thought that was what made the argument optional. I understand from the comment above that this block will be removed, right?

valeriocos: ehehe
yes, it guess it's better to remove it.

perceval/backends/core/confluence.py
maoo: As per comments above, if we want to always enable ancestors enrichment, then I'd simply change line 354 to `resource = self.RCONTENTS + '/' + self.MSEARCH + '?' + self.PEXPAND '=' + self.PANCESTORS` and remove the other part of the change. WDYT?

perceval/backends/core/confluence.py
maoo: So you're basically proposing to remove `url_prefix +` from line 100 ?

perceval/backends/core/confluence.py
valeriocos: I would put it in a different method to:
- be sure that the paginated ancestors are returned (I'm not sure whether the code you proposed that care of this)
- simplify the testing

perceval/backends/core/confluence.py
```diff
@@ -86,11 +86,22 @@ def fetch(self, category=CATEGORY_HISTORICAL_CONTENT, from_date=DEFAULT_DATETIME
...
+        kwargs = {
+            'from_date': from_date
+        }
+
         items = super().fetch(category, **kwargs)
 
         return items
 
+    @staticmethod
+    def ancestors(item, content):
```
valeriocos: `ancestors` should not be static and probably it's better if it's a private method like `__fetch_contents_summary`

perceval/backends/core/confluence.py
```diff
@@ -86,11 +86,22 @@ def fetch(self, category=CATEGORY_HISTORICAL_CONTENT, from_date=DEFAULT_DATETIME
...
+        ancestors = []
+        if ('ancestors' in content):
```
valeriocos: tests coverage went done due to this code, a test is needed when no ancestors are returned.
I would put something:
```
if 'ancestors' not in content:
 return []

...
```

perceval/backends/core/confluence.py
```diff
@@ -86,11 +86,22 @@ def fetch(self, category=CATEGORY_HISTORICAL_CONTENT, from_date=DEFAULT_DATETIME
...
+            for ancestor in content['ancestors']:
```
valeriocos: to check the test coverage, you could use the package https://coverage.readthedocs.io/en/v4.5.x/

perceval/backends/core/confluence.py
maoo: I really don't see the difference between the 2 syntaxes, since the guard is exactly the same.

I believe this code is less readable, but may be personal preference:

```
    def __fetch_ancestors(self, item, content):
        if ('ancestors' not in content):
            return []
        else:
            ancestors = []
            for ancestor in content['ancestors']:
                ancestors.append(ancestor['_links']['webui'])
            return ancestors
```

If this is not what you meant, can you please share your own version? Maybe I'm not understanding your point here.

valeriocos: yes, sure! the idea in Python is to reduce the number of indentation as much as possible (https://stackoverflow.com/questions/10959683/preferred-maximum-indentation-in-python

In the code you provide above, the else statement is not needed. If the if-condition is hit, the program will return after executing the nested statement. Conversely, if the if-condition is not true, the program will ignore that part and go directly to the next statement.

```
def __fetch_ancestors(self, item, content):
        ancestors = []
        if 'ancestors' not in content: # parenthesis are not needed
            return ancestors
       
       # the code below should go to the client, to avoid mixing logic of different level of abstraction (in a nutshell, the backend should not know the low level-details of the data returned by the API)
       for ancestor in content['ancestors']:
             ancestors.append(ancestor['_links']['webui'])
       return ancestors
```

maoo:
@valeriocos , I should have addressed all issues, except one: testing the private `__fetch_ancestors` function. I'm trying to do something like the following, but doesn't like it.

```
    def test_fetch_ancestors(self):
        """Test if ancestors are properly fetched"""

        http_requests = setup_http_server()

        confluence = Confluence(CONFLUENCE_URL)
        from_date = datetime.datetime(2016, 6, 16, 0, 0, 0)
        content = confluence.fetch(from_date=from_date)

        ancestors = confluence.__fetch_ancestors(content[0])

        # TODO - replace with asserts
        print(ancestors)
```

This fails with

```
ERROR: test_fetch_ancestors (__main__.TestConfluenceBackend)
Test if ancestors are properly fetched
  File "tests/test_confluence.py", line 419, in test_fetch_ancestors
    ancestors = confluence.__fetch_ancestors(content[0])
AttributeError: 'Confluence' object has no attribute '_TestConfluenceBackend__fetch_ancestors'
```

Could you please help me figuring this out? Is there anything else left? TY!

perceval/backends/core/confluence.py
```diff
@@ -94,13 +94,14 @@ def fetch(self, category=CATEGORY_HISTORICAL_CONTENT, from_date=DEFAULT_DATETIME
 
         return items
 
-    @staticmethod
-    def ancestors(item, content):
-        ancestors = []
-        if ('ancestors' in content):
+    def __fetch_ancestors(self, content):
```
valeriocos: Since Perceval tries to manipulate as less as possible the data obtained from a data source, I would add all the data returned in `ancestors`

perceval/backends/core/confluence.py
```diff
@@ -129,7 +130,7 @@ def fetch_items(self, category, **kwargs):
 
             for hc in hcs:
                 hc['content_url'] = content_url
-                hc['ancestors'] = Confluence.ancestors(hc, content)
+                hc['ancestors'] = self.__fetch_ancestors(content)
```
valeriocos: according to the comment below, this line could be transformed into:
`hc['ancestors'] = content['ancestors'] if 'ancestors' in content else []`

valeriocos: as minor comment, I would add a space after this statement to improve the readability (feel free to address this comment if you want)

maoo:
All done @valeriocos ! Thanks for your patience and happy weekend/

valeriocos:
I tested the PR and it works fine. Overall LGTM, maybe there are a couple of minor things to improve:
- improve the commit msg (the description looks OK)
- add @maoo info here: https://github.com/maoo/grimoirelab-perceval/blob/confluence-ancestors/perceval/backends/core/confluence.py#L20
- include in the test data an example of the ancestors data

@sduenas feel free to make your pass on this PR.

maoo:
@valeriocos , thank you! Can you please check now?

perceval/backends/core/confluence.py
```diff
@@ -331,7 +338,7 @@ def contents(self, from_date=DEFAULT_DATETIME,
         :param offset: fetch the contents starting from this offset
         :param limit: maximum number of contents to fetch per request
         """
-        resource = self.RCONTENTS + '/' + self.MSEARCH
+        resource = resource = self.RCONTENTS + '/' + self.MSEARCH
```
valeriocos: ?

valeriocos:
Great @maoo ! thank you a lot.
@sduenas please have a look

perceval/backends/core/confluence.py
maoo: uh, weird, sorry about that. just pushed the fix

maoo:
Thanks @sduenas for the quick review and for the suggestions! I bumped the version to 0.10.0, changed the fallback logic on `hc['ancestors']` and tried to define a better commit message; could you please let me know if the new message is good enough?

sduenas:
Thanks @maoo, the commit message is great. I forgot to mention that we add a tag to the summary of the commit to locate the commits better if we have to search for them. You already have it in your commit before. Just put `[confluence] Extend Perceval Confluence backend to index ancestor pages` as the first line of the commit. Btw, you have a typo, it's Perceval and not Parceval :D

Thanks for the contribution!

maoo:
@sduenas , apologies for the typo(!!), just fixed; also added the `[confluence]` tag at the beginning.

sduenas:
Merged.

@maoo  I had to edit the commit message because lines were longer that 70 characters. Anyway, I pushed to your branch to merge this PR. Hope you don't mind.

Maybe @valeriocos can tell us more about this but I think the next step is to include the `ancestors` values into the enriched items.

maoo:
Awesome, thank you @sduenas and @valeriocos !

valeriocos:
yes, sure @sduenas . I guess with modifying this method: https://github.com/chaoss/grimoirelab-elk/blob/master/grimoire_elk/enriched/confluence.py#L97 by copying the content of the ancestors field should be enough (the csv should also be updated).

maoo:
@valeriocos , please let me know if you want me to send you a PR for this too!

valeriocos:
thank you @maoo , feel free to open an issue in ELK, we can discuss there

maoo:
@valeriocos , done, see https://github.com/chaoss/grimoirelab-elk/issues/571

PR #496, opened 2019-03-25
[git] Fix `rev-list` call with `--max-count`

valeriocos:
This code fixes the call of the git command `rev-list` with parameter `--max-count`, which needs to be prepended with `--branches` or `--tags`.

### DIFF
--- a/perceval/backends/core/git.py
+++ b/perceval/backends/core/git.py
@@ -64,7 +64,7 @@ class Git(Backend):
     :raises RepositoryError: raised when there was an error cloning or
         updating the repository.
     """
-    version = '0.11.0'
+    version = '0.11.1'
 
     CATEGORIES = [CATEGORY_COMMIT]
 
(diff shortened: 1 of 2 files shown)


perceval/backends/core/git.py
```diff
@@ -968,7 +968,7 @@ def rev_list(self, branches=None):
         if branches is None:
             cmd_rev_list.extend(['--branches', '--tags', '--remotes=origin'])
         elif len(branches) == 0:
-            cmd_rev_list.append('--max-count=0')
+            cmd_rev_list.extend(['--branches', '--tags', '--max-count=0'])
```
sduenas: If `branches` is an empty list, it should return an empty list of references, shouldn't it?

valeriocos: I agree with you, for consistency with the codebase I mimicked the code at: https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/backends/core/git.py#L1030. Should that part also be changed?

sduenas: Sorry, I understood wrongly the patch. I get now why that's needed. To run the command. :)

Issue #490, opened 2019-03-10
Filter sensitive data when generating items


sduenas:
Some data such as email or names can be considered as sensitive. Although, Perceval fetches these data, some user might want to filter these data before the items are generated. For that reason, Perceval should provide a way to generate this kind of filtered items.

Fields considered as sensible must be predefined (`hardcoded`) for each backend. When these items are generated, metadata should contain enough information to distinguish these kind of items from normal ones.

sduenas:
My idea to implement this feature consist on:

* Every backend will have a list of **confidential** or **classified** fields. This list will be empty most of the times. By default, this list will be defined inside `Backend` class as empty. It can be reused and it won't be necessary to set it in every subclass.
* `Backend` class will have a parameter in `__init__` method to set if it will be necessary to remove these classified fields before generating any item.
* `Backend` class will implement a method to filter these items. Filtering will be done before generating the metadata of each item.
* Metadata will include a field to indicate if classified items were removed or not. This will allow to identify items with the same UUID but with different fields.
* Users will be able to get items where confidential data is removed

PR #497, opened 2019-03-26
Option to filter classified data

sduenas:
This PR addresses the requirements defined in #490.

Classified fields can be declared using `CLASSIFIED_FIELDS` class attribute. This attribute will contain a list of lists. Each list stores the "path" or nested dicts keys to the field to filter. For example:
`['my', 'classified', 'field']` will remove `field` from `item['data']['my']['classified']` dict.

To filter this fields from the resulting items, `fetch()` method adds a new parameter `filter_classified`.
Classified fields can also be removed from the command line with the option `filter-classified`.

To start with new feature, this PR adds the field `group.topics` to the list of classified fields in Meetup backend. This field is considered sensible for some related groups working with GrimoireLab.

### DIFF
--- a/perceval/backend.py
+++ b/perceval/backend.py
@@ -450,16 +450,17 @@ class BackendCommand:
         """
         backend_args = vars(self.parsed_args)
         category = backend_args.pop('category', None)
+        filter_classified = backend_args.pop('filter_classified', False)
         archived_since = backend_args.pop('archived_since', None)
 
         if self.archive_manager and self.parsed_args.fetch_archive:
-
             items = fetch_from_archive(self.BACKEND, backend_args,
                                        self.archive_manager,
                                        category,
                                        archived_since)
         else:
             items = fetch(self.BACKEND, backend_args, category,
+                          filter_classified=filter_classified,
                           manager=self.archive_manager)
 
         try:
--- a/perceval/backend.py
+++ b/perceval/backend.py
@@ -533,7 +533,8 @@ def uuid(*args):
     return uuid_sha1
 
 
-def fetch(backend_class, backend_args, category, manager=None):
+def fetch(backend_class, backend_args, category, filter_classified=False,
+          manager=None):
     """Fetch items using the given backend.
 
     Generator to get items using the given backend class. When
@@ -548,6 +549,7 @@ def fetch(backend_class, backend_args, category, manager=None):
     :param backend_args: dict of arguments needed to fetch the items
     :param category: category of the items to retrieve.
        If None, it will use the default backend category
+    :param filter_classified: remove classified fields from the resulting items
     :param manager: archive manager needed to store the items
 
     :returns: a generator of items
@@ -561,6 +563,8 @@ def fetch(backend_class, backend_args, category, manager=None):
 
     if category:
         backend_args['category'] = category
+    if filter_classified:
+        backend_args['filter_classified'] = filter_classified
 
     fetch_args = find_signature_parameters(backend.fetch,
                                            backend_args)
--- a/perceval/backend.py
+++ b/perceval/backend.py
@@ -39,6 +39,7 @@ from ._version import __version__
 
 logger = logging.getLogger(__name__)
 
+
 ARCHIVES_DEFAULT_PATH = '~/.perceval/archives/'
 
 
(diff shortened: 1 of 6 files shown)


perceval/backend.py
```diff
@@ -100,32 +110,48 @@ def archive(self, obj):
...
         :returns: a generator of items
+
+        :raises BackendError: either when the category is not valid or
+            'filter_classified' and 'archive' are active at the same time.
         """
         if category not in self.categories:
             cause = "%s category not valid for %s" % (category, self.__class__.__name__)
             raise BackendError(cause=cause)
 
+        if filter_classified and self.archive:
```
valeriocos: The archive stores the url, headers, payload and response of a HTTP request, and the filter is applied later. Why `filter_classified` is disabled when using the archive?

perceval/backend.py
```diff
@@ -161,6 +214,7 @@ def metadata(self, item):
             'origin': self.origin,
             'uuid': uuid(self.origin, self.metadata_id(item)),
             'updated_on': self.metadata_updated_on(item),
+            'classified_fields_filtered': self.classified_fields if filter_classified else None,
```
valeriocos: I'm not sure if this should be included to the item here or later only if `filter_classified` is true. Logically, it makes sense to add it where it is now, but in practice it implies that we should re-create all raw indexes we have (to avoid mixing old items created before this change and new ones with the `classified_fields_filtered` attribute)

If we add the field only if `filter_classified` is true, this would only affect the raw indexes for the customer(s) that asked for this feature.

perceval/backend.py
```diff
@@ -192,6 +246,18 @@ def _init_client(self, from_archive=False):
         raise NotImplementedError
 
 
+def _remove_key_from_nested_dict(nested_dict, path_to_field):
```
valeriocos: Although it could be a corner case, items with a high nesting level could cause problems in terms of memory consumption. If possible, it would be better to use a loop.

tests/test_backend.py
```diff
@@ -310,6 +360,146 @@ def test_fetch_items(self):
...
+                'item': x,
+                'my': {
+                    'classified': {'field': x},
+                    'field': x,
+                },
+            }
+            self.assertDictEqual(item['data'], expected)
+
+    def test_error_archive_and_filter_classified(self):
+        """CHeck if an error is raised when archive and classified fields filtering are both active"""
```
valeriocos: typo -> CHeck

perceval/backend.py
sduenas: Because the fields that are classified come on that response. Therefore, they will be stored in the archive which I think it doesn't make sense if you want to remove them from the result. It would cause security problems in your organization, if you don't remember there's an archive storing them.

valeriocos: Perfect, could you add a coment on top of that condition ?

sduenas: I rather add it to the docstring of that method.

perceval/backend.py
sduenas: I thought of that but after a long discussion with myself, I still think all items generated with a given version should have the same fields, at least in terms of those fields we control. I know it is verbose for the cases where is not in use, but it allows to see nothing was filtered, instead of wondering why this field is available on these items and not in others generated with the same Perceval version.

In terms of storing these items I would also like to have all items with the same data but sometimes it isn't possible.
On the other hand, GrimoireLab uses a schemaless database (Elastic Search) to store items. This change won't affect already created indexes. New items will be added to existing indexes without problems.

Moreover, take into account, this field only works as reference/debugging. To know if data was filtered from the original or not. It shouldn't be used in queries, etc.

valeriocos: Thank you for the clarification, it sounds reasonable.

> On the other hand, GrimoireLab uses a schemaless database (Elastic Search) to store items. This change won't affect already created indexes. New items will be added to existing indexes without problems.

Yes, this change won't affect existing indexes (for the schemaless structure), but we will have items with different meta information. I guess we can live with this discrepancy for a while.

sduenas:
> Overall it looks OK, I left a couple of minor comments. Could you clarify why the filtered items are hard-coded in the backends and cannot be passed as param in the command line ?
> 
> PS fix the flake8 error

Same idea as always. Don't have first and second class items. I'm not saying we shouldn't do that in the future but we don't know yet the implications this feature could cause in the rest of the chain.
What we know for sure is if we allow to do that now, ELK will fail if you filter some of the required fields.

For this iteration, I'd rather stick to this but I'm not saying I didn't think of it. One of the reasons `classified_fields_filtered` in metadata is a list and not a boolean value is this. If in the future we allow to pass a list of fields to filter, the same field will work, storing the fields filtered on those items.

valeriocos:
ok, I was thinking about the case where two customers want to filter different attributes on the same data source. Probably for the first iterations is better to keep it as it is.

tests/test_backend.py
sduenas: I think this is one of the cases where recursion works better. Nested fields shouldn't be longer than 4 or 5 keys? I tested with 100 and 1000 nested keys and it works without problems. If wants to filter something with a deep of 10 keys...it's probably because he's doing something terrible wrong.

perceval/backend.py
sduenas: I think this is one of the cases where recursion works better. Nested fields shouldn't be longer than 4 or 5 keys? I tested with 100 and 1000 nested keys and it works without problems. If wants to filter something with a deep of 10 keys...it's probably because he's doing something terrible wrong.

PR #522, opened 2019-05-08
[gitlab] Add support for Gitlab Oauth tokens

jjmerchante:
GitLab's backend in Perceval doesn't support Oauth tokens. While in GitHub (or many other backends) you can use the same the Header for Personal Tokens and Oauth Tokens, in GitLab is different ([GitLab Personal tokens](https://docs.gitlab.com/ee/api/#personal-access-tokens), [GitLab Oauth Tokens](https://docs.gitlab.com/ee/api/#oauth2-tokens)).

This contribution includes that feature for Gitlab backend. The user can introduce a token and indicate that is OAuth with the parameter `--is-oauth-token`.

Some tests are included.

Signed-off-by: JJMerchante <jj.merchante@gmail.com>

### DIFF
--- a/perceval/backends/core/gitlab.py
+++ b/perceval/backends/core/gitlab.py
@@ -37,6 +37,7 @@ from ...backend import (Backend,
                         BackendCommandArgumentParser)
 from ...client import HttpClient, RateLimitHandler
 from ...utils import DEFAULT_DATETIME
+from ...errors import BackendError
 
 CATEGORY_ISSUE = "issue"
 CATEGORY_MERGE_REQUEST = "merge_request"
@@ -68,6 +69,7 @@ class GitLab(Backend):
     :param owner: GitLab owner
     :param repository: GitLab repository from the owner
     :param api_token: GitLab auth token to access the API
+    :param is_oauth_token: True if the token is OAuth (default False)
     :param base_url: GitLab URL in enterprise edition case;
         when no value is set the backend will be fetch the data
         from the GitLab public site.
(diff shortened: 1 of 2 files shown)


tests/test_gitlab.py
```diff
@@ -369,6 +369,22 @@ def test_initialization(self):
...
+        self.assertEqual(gitlab.tag, 'test')
+        self.assertIsNone(gitlab.client)
+        self.assertIsNone(gitlab.blacklist_ids)
+        self.assertEqual(gitlab.max_retries, MAX_RETRIES)
+        self.assertEqual(gitlab.sleep_time, DEFAULT_SLEEP_TIME)
+        self.assertEqual(gitlab.is_oauth_token, True)
+
+        with self.assertRaises(BackendError):
+            gitlab = GitLab('fdroid', 'fdroiddata', is_oauth_token=True, tag='test')
+
```
Polaris000: Hi @jjmerchante. The unit tests passed but the flake8 test did n't. I think you can remove those extra lines (they seem to be the problem) so that all the tests pass. Thanks!

valeriocos: Maybe it's better to move these two tests to a separate method. Furthermore if the variable `gitlab` is not used you can replace it with `_`, for instance:

```
with self.assertRaises(BackendError):
            _ = GitLab('fdroid', 'fdroiddata', is_oauth_token=True, tag='test')
```

A similar example can be found at: https://github.com/chaoss/grimoirelab-perceval/blob/master/tests/test_github.py#L1678

tests/test_gitlab.py
```diff
@@ -894,13 +910,15 @@ def test_initialization(self):
...
         client = GitLabClient("fdroid", "fdroiddata", "your-token", sleep_for_rate=True,
-                              min_rate_to_sleep=100, max_retries=10, sleep_time=100)
+                              min_rate_to_sleep=100, max_retries=10, sleep_time=100, is_oauth_token=True)
 
         self.assertEqual(client.sleep_for_rate, True)
         self.assertEqual(client.base_url, GITLAB_API_URL)
         self.assertEqual(client.min_rate_to_sleep, 100)
         self.assertEqual(client.sleep_time, 100)
         self.assertEqual(client.max_retries, 10)
+        self.assertEqual(client.is_oauth_token, True)
```
valeriocos: As you similarly did for `TestGitLabBackend.test_initialization`, you should add assertions around line 910 to check that (i) `is_oauth_token` is False and (ii) the client headers are properly set.

tests/test_gitlab.py
```diff
@@ -1323,6 +1341,17 @@ def test_sanitize_for_archive(self):
         self.assertEqual(headers, s_headers)
         self.assertEqual(payload, s_payload)
 
+        url = "http://example.com"
```
valeriocos: This method could be splitted in two: `test_sanitize_for_archive_private_token` and `test_sanitize_for_archive_oauth_token`. The former is the old `test_sanitize_for_archive` while the new one contains your code.

What do you think ?

perceval/backends/core/gitlab.py
```diff
@@ -373,6 +379,7 @@ class GitLabClient(HttpClient, RateLimitHandler):
          before raising a RetryError exception
     :param archive: an archive to store/read fetched data
     :param from_archive: it tells whether to write/read the archive
+    :param is_oauth_token: True if the token is OAuth and not private/personal
```
valeriocos: Depending on the outcome of the previous comment, this param should be moved accordingly.

jjmerchante:
Thank you for the review, I made some changes according to your comments

valeriocos:
@sduenas please have a look at the PR when you have time, thanks!

sduenas:
@jjmerchante there are conflicts in your PR, please fix them before we accept this PR

jjmerchante:
@sduenas  It should be resolved

Issue #447, opened 2018-10-20
Windows - problems cloning a github repo


gaurav5430:
On windows 8, this is how i made perceval work within a Windows installation:

**Installing perceval**
while trying to install with pip i got into problems with dulwich on windows trying to build the optional C extensions which probably requires a C compiler on path , so did a separate install of dulwich with --global-option="--pure" . This required installing dulwich dependencies, and then i had to install perceval with --global-option="--pure" .

```
pip3 install certifi urllib3
pip3 install "dulwich==0.18.6" --global-option="--pure"
pip3 install perceval
```
Finally got it to work within a virtual environment (named 'gl') like this:

`cmd > gl\Scripts\python gl\Scripts\perceval git https://github.com/chaoss/grimoirelab-perceval.git `, but then it tries to clone this repo to a local path which has the full repo url in the name, and so windows refuses to create it and throws an error like : `git command - fatal: could not create leading directories of 'C:\Users\...\https://github.com/chaoss/grimoirelab-perceval.git-git` . This does not work because the name that perceval uses for the cloned repo is not in a correct format. To get over it, I tried to use the --git-path flag, which leads to another issue:  Couldn't resolve host 'github.com' . I think it might be similar to https://github.com/chaoss/grimoirelab-perceval/issues/399

Please note that i was able to make it work with local git repo and see the commits.

sduenas:
We don't support Windows. I'm closing this ticket.

Issue #399, opened 2018-07-03
Git backend [Windows 7/10] issue


muhmaz3:
For us GitHub backend  works, but the Git backend is not working.  Git is already Install on our test machine, we can clone the repo by using Git command line. We used Perceval example code, but we got an error.  Looking into the  git backend, **if we set `evn = None`, line 1268 then it works.** But  we could  not figure out which environment variable we are missing or why it works  when `env= None`. Any guidance on this will be really helpful. 

`repo_url = 'https://github.com/chaoss/grimoirelab-perceval'`
`repo_dir = 'chaooos2'`
`repo = Git(uri=repo_url, gitpath=repo_dir)`

`# fetch all commits as an iteratoir, and iterate it printing each`
`for commit in repo.fetch():`
   `print(commit['data']['commit'])`

> **git command - fatal: unable to access  'https://github.com/chaoss/grimoirelab-perceval/': Could not resolve host: github.com**

sduenas:
Sorry for the late response. I'm not sure about how environment variables work in Windows but `HOME` variable is needed because it's where git configuration file is stored. I'm afraid `os.getenv('HOME', '')` doesn't find this variable and sets its value to `''` which would make your code fail. I'm not sure about this, though.

Would you try to run just this snippet so we can figure out whether `HOME` exists or not?

```
import os
env = os.getenv('HOME', '')
print("HOME:", env)
```

More info about Git environment variables can be found in [Git Internals Environment Variables](https://git-scm.com/book/gr/v2/Git-Internals-Environment-Variables) and [Getting Started with Git Setup](https://git-scm.com/book/en/v2/Getting-Started-First-Time-Git-Setup)

jgbarah:
Maybe there is no HOME in Windows: http://www.itprotoday.com/management-mobility/what-environment-variables-are-available-windows

What do you think about checking for HOME, and if that fails, check for HOMEPATH? (or check for Windows, and have a different code in that case, directly reading HOMEPATH....)

muhmaz3:
@sduenas I have tried your suggestion and as @jgbarah mentioned, there is no HOME on windows and I also got returned default value. Thereafter, I tried  HOMEDRIVE, and HOMEPATH, USERPROFILE, non of them worked.

jgbarah:
[...]

Issue #331, opened 2018-02-21
successfully installed perceval 0.9.10 on a Windows machine, but perceval  is not recognized as a command on the command prompt


kkc-al:
when i try to run "perceval  --version" or "perceval --help" in the virtual environment, I get "'perceval' is not recognized as an internal or external command, operable program or batch file."

kk-gndu:
even "Mordred --help" on the command prompt is not working. I am following  the tutorial 
Installing GrimoireLab Python modules available at https://grimoirelab.gitbooks.io/tutorial/content/before-you-start/installing-grimoirelab.html

kk-gndu:
the list of modules installed in the virtual environment:
(grimoirelab) E:\venvs\grimoirelab\Scripts>pip list
DEPRECATION: The default format will switch to columns in the future. You can use --format=(legacy|columns) (or define a format=(legacy|columns) in your pip.conf under the [list] section) to disable this warning.
beautifulsoup4 (4.6.0)
brewer2mpl (1.4.1)
certifi (2018.1.18)
chardet (3.0.4)
cheroot (5.8.3)
CherryPy (11.0.0)
click (6.7)
cycler (0.10.0)
Cython (0.27.3)
dulwich (0.18.6)
elasticsearch (6.1.1)
elasticsearch-dsl (6.1.0)
feedparser (5.2.1)
grimoire-elk (0.30.23)
grimoire-kidash (0.30.9)
grimoire-mordred (0.1.15)
grimoire-reports (0.1.5)
grimoirelab-panels (0.0.8)
grimoirelab-toolkit (0.1.4)
idna (2.6)
ipaddress (1.0.19)
Jinja2 (2.10)
kidash (0.4.2)
kingarthur (0.1.3)
MarkupSafe (1.0)
matplotlib (2.1.2)
numpy (1.14.0)
pandas (0.18.1)
perceval (0.9.10)
perceval-mozilla (0.1.5)
perceval-opnfv (0.1.4)
pip (9.0.1)
portend (2.2)
prettyplotlib (0.1.7)
PyMySQL (0.8.0)
pyparsing (2.2.0)
pypiwin32 (220)
python-dateutil (2.6.1)
pytz (2018.3)
PyYAML (3.12)
redis (2.10.6)
requests (2.18.4)
rq (0.10.0)
setuptools (38.5.1)
six (1.11.0)
sortinghat (0.5.2)
SQLAlchemy (1.2.2)
tempora (1.10)
[...]

PR #526, opened 2019-05-10
windows 10 changed #!/usr/bin/env python3 to #!/usr/bin/env python

dereklowlind:
-changed #!/usr/bin/env python3 to #!/usr/bin/env python so that windows will not error "/usr/bin/env: ‘python3’: No such file or directory"
-may cause issues if a user has more than one version of python installed
-could make a windows specific copy "pip3 install perceval_windows"

### DIFF
--- a/bin/perceval
+++ b/bin/perceval
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
 # -*- coding: utf-8 -*-
 #
 # Copyright (C) 2015-2019 Bitergia
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
 # -*- coding: utf-8 -*-
 #
 # Perceval documentation build configuration file, created by
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
 # -*- coding: utf-8 -*-
 #
 # Copyright (C) 2015-2019 Bitergia
--- a/tests/base.py
+++ b/tests/base.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
 # -*- coding: utf-8 -*-
 #
 # Copyright (C) 2015-2019 Bitergia
--- a/tests/run_tests.py
+++ b/tests/run_tests.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
 # -*- coding: utf-8 -*-
 #
 # Copyright (C) 2015-2019 Bitergia
--- a/tests/test_archive.py
+++ b/tests/test_archive.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
 # -*- coding: utf-8 -*-
 #
 # Copyright (C) 2015-2019 Bitergia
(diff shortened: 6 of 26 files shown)


valeriocos:
Thank you @dereklowlind for the PR. I've tested the PR and it works fine, however I'm not sure about the change you propose since `#!/usr/bin/env python3` and `#!/usr/bin/env python` should have different meanings in Linux, while in Windows (as reported [here](https://docs.python.org/3.7/using/windows.html#shebang-lines)) only `/usr/bin/env python` is supported.

I like the idea of having a perceval for Windows, and it seems reasonable since we have already some issues reported (i.e., https://github.com/chaoss/grimoirelab-perceval/issues/399, https://github.com/chaoss/grimoirelab-perceval/issues/331, https://github.com/chaoss/grimoirelab-perceval/issues/447). However, we have also to keep in mind that perceval is developed on Linux, thus we don't have proper envs (windows 7, 10, ...) to test it and thus we cannot guarantee that `perceval_windows` will be as good as the Linux version.

@sduenas @jgbarah please share your thoughts, thanks!

dereklowlind:
I have packaged my windows "port" and uploaded it to the pip test site and it works installing from the pip command line tool. I added this message to the top of the description "Warning: This is a port to windows in which some features are broken. Please use only to test out the concept and then setup linux and run normal Perceval". Would anyone object to me putting perceval-windows onto the main pip site?

valeriocos:
It's ok for me @dereklowlind , but we really need the feedback of @sduenas and @jgbarah.

Btw, where will be located this version of perceval? shouldn't perceval-windows be a repo under chaoss org?

dereklowlind:
I did this first port in a forked repo in my personal account (for ease of submitting changes). If the community would like to make it more official then my side project, I would be very happy for it to become a repo of chaoss.org!

valeriocos:
Great @dereklowlind! :) It would be really nice if you can join the community and help to maintain perceval-windows.

Please @sduenas, @jgbarah share your thoughts

dereklowlind:
Timeline for moving forward on perceval-windows?

valeriocos:
ping @sduenas @jgbarah

jgbarah:
I find it very interesting to have Perceval running in Windows. And as a temporary solution, using perceval-windows as a pip package, could be a good idea. However, a couple of comments:

* I'm not sure the change from `python3` to `python` is a good idea for non-Windows systems. Probably it is going to become a minor issues as python3 becomes the only python, but I see problems meanwhile, specially when people have both systems installed. I know with virtual environments this is a non-issue, but still... Is there no way of fixing this in Windows in some other way?
* With respect to the `perceval-windows` package in pip, which differences would it have? Because right now, `perceval` is mostly a source package.

sduenas:
I agree it would be great to have Perceval running on Windows. I don't see the solution you propose in this PR the best one, though. It will break how it works in Linux systems now.

One of the possible solutions that might work would be to install a different script depending the OS. A BAT file (or similar) for Windows and the Python file for Unix systems.

Some parts of the code should be adapted to support multi-platform. We can also have tests for both environments in Travis.

dereklowlind:
How about using the unspecified python version, checking if it is acceptable, then if not finding the highest version available on the system? I have not checked if the code below works.

https://unix.stackexchange.com/questions/66210/how-can-i-have-more-than-one-possibility-in-a-scripts-shebang-line

#!/usr/bin/env python
import os, sys

pythons = [ '/usr/bin/python2.3','/usr/bin/python2.4', '/usr/bin/python2.5', '/usr/bin/python2.6', '/usr/bin/python2.7' ]
py = list(filter( os.path.isfile, pythons ))
if py:
  py = py.pop()
  thepy = int( py[-3:-2] + py[-1:] )
  mypy  = int( ''.join( map(str, sys.version_info[0:2]) ) )
  if thepy > mypy:
    print("moving versions to "+py)
    args = sys.argv
    args.insert( 0, sys.argv[0] )
    os.execv( py, args )

print("do normal stuff")

PR #530, opened 2019-05-16
[slack] Handle timestamps with more than 6 decimals

valeriocos:
This code handles timestamps having more than 6 decimals by rounding their values to the 6th decimal. This change is needed to avoid failures when fetching channels.history data using latest and/or oldest params (which cannot exceed 6 decimals).

### DIFF
--- a/perceval/backends/core/slack.py
+++ b/perceval/backends/core/slack.py
@@ -37,6 +37,7 @@ CATEGORY_MESSAGE = "message"
 
 SLACK_URL = 'https://slack.com/'
 MAX_ITEMS = 1000
+FLOAT_FORMAT = '{:.6f}'
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


perceval/backends/core/slack.py
```diff
@@ -118,7 +119,7 @@ def fetch_items(self, category, **kwargs):
 
         # Minimum value supported by Slack is 0 not 0.0
         if oldest == 0.0:
-            oldest = 0
+            oldest = 0.000000
 
         # Slack does not include on its result the lower limit
```
sduenas: Does it make sense to keep the code below? If dates are rounded to 6 decimals later, this part wouldn't be needed anymore.

perceval/backends/core/slack.py
```diff
@@ -118,7 +119,7 @@ def fetch_items(self, category, **kwargs):
 
         # Minimum value supported by Slack is 0 not 0.0
         if oldest == 0.0:
-            oldest = 0
```
sduenas: Check the comment above. There must be a way to handle this case.

PR #533, opened 2019-05-29
[dependencies] Pin urllib3 and requests dependencies

valeriocos:
This code fixes the urllib3 version used by perceval to 1.22, since in the later versions urllib3.exceptions.SSLError exceptions may be thrown due to CERTIFICATE_VERIFY_FAILED. Furthermore, it also sets the version of requests to 2.18.2, which is compatible with urllib3 1.22.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 python-dateutil>=2.6.0
-requests>=2.7.0
+requests==2.18.2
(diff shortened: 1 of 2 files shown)


valeriocos:
After investigation, this PR doesn't seem to be needed since `requests` will just output a warning instead of throwing an exception as done by elasticsearch 6.3.1

PR #506, opened 2019-04-01
username,password attributes for jenkins data collection

nitishvu:
as discussed with @valeriocos added user credentials support for "perceval jenkins"
discussion link: []()https://github.com/chaoss/grimoirelab-sirmordred/issues/287

current scenario:

1) help page:

              $ perceval jenkins

                         [2019-04-01 15:28:24,642] - Sir Perceval is on his quest.

                          usage: perceval [-h] [--category CATEGORY] [--tag TAG] 

                          [--filter-classified]

                          [--archive-path ARCHIVE_PATH] [--no-archive] [--fetch-archive]

                          [--archived-since ARCHIVED_SINCE] [-o OUTFILE] [--json-line]

                          [--blacklist-jobs [BLACKLIST_JOBS [BLACKLIST_JOBS ...]]]

                          [--detail-depth DETAIL_DEPTH] [--sleep-time SLEEP_TIME]

                          [--username USERNAME] 

                          [--password PASSWORD]
 
                          url
`
2) sample query:

`
        perceval jenkins --username jenkinsuser@org.com --password jenkinspassword 
            http://devops.jenkins.com:8080`

PR #433, opened 2018-09-10
WIP: init jenkins authentication feature

MalloZup:
# What change this PR:

Usually people/org use Jenkins with Authentification.

This pr enable the jenkins backend to auth against a jenkins server

This fix https://github.com/chaoss/grimoirelab-perceval/issues/432

# what need to be done:

- [ ]  Backend implementation

- [ ] Tests

- [ ] Doc

Issue #432, opened 2018-09-06
[jenkins] [feature] jenkins backend cannot authenticate


MalloZup:
Some jenkins clients need to authenticate.

I think that we need top pass user and password to the jenkins server and implement this feature on the jenkins backend.

I could have a look on this

sduenas:
Thanks for working on this.

PR #534, opened 2019-06-07
[jenkins] Support authenticated access

valeriocos:
This PR allows to fetch builds and jobs available via a Jenkins server, which requires authenticated access. The authentication supported are (user, password) and (user, API token)*.

This PR also extends the Perceval client, the `fetch` method now includes a new param `auth`, which allows to issue requests using basic authentication.

This PR replaces/addresses:
- https://github.com/chaoss/grimoirelab-perceval/issues/432
- https://github.com/chaoss/grimoirelab-perceval/pull/506
- https://github.com/chaoss/grimoirelab-perceval/pull/433
- https://github.com/chaoss/grimoirelab-sirmordred/issues/287

Information about how to retrieve a Jenkins API token is available at:
https://medium.com/@samratshaw/programmatically-retrieve-jenkins-rest-api-token-f2c3f0d69483

### DIFF
--- a/perceval/backends/core/jenkins.py
+++ b/perceval/backends/core/jenkins.py
@@ -47,6 +47,8 @@ class Jenkins(Backend):
     The `url` will be set as the origin of the data.
 
     :param url: Jenkins url
+    :param user: Jenkins user
+    :param password: Jenkins password
     :param tag: label used to mark the data
     :param archive: archive to store/retrieve items
     :param blacklist_jobs: exclude the jobs of this list while fetching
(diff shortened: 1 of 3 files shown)


valeriocos:
I'm not sure it's possible. To tackle this problem the PR provides support also to use tokens (instead of 
a password). As reported by the [doc](https://wiki.jenkins.io/display/JENKINS/Authenticating+scripted+clients), this should guarantee more safeness
```
API token is new since 1.426

Earlier versions of Jenkins require you to specify your real password, 
and it is only available when your security realm is password-based (for example, OpenID, 
Crowd and CAS plugins authenticate you without a password, so you simply 
don't have any password!) Specifying the real password is still supported after 1.426, 
but it is not recommended because the risk of revealing password, 
and the human tendency to reuse the same password in different places.
```

sduenas:
> I'm not sure it's possible. To tackle this problem the PR provides support also to use tokens (instead of
> a password). As reported by the [doc](https://wiki.jenkins.io/display/JENKINS/Authenticating+scripted+clients), this should guarantee more safeness
> 
> ```
> API token is new since 1.426
> 
> Earlier versions of Jenkins require you to specify your real password, 
> and it is only available when your security realm is password-based (for example, OpenID, 
> Crowd and CAS plugins authenticate you without a password, so you simply 
> don't have any password!) Specifying the real password is still supported after 1.426, 
> but it is not recommended because the risk of revealing password, 
> and the human tendency to reuse the same password in different places.
> ```

If that's the case, I suggest to only support tokens and avoid the user/password authentification.

valeriocos:
As you prefer, I opened another PR here: https://github.com/chaoss/grimoirelab-perceval/pull/549

Issue #538, opened 2019-06-18
[Question] trying to use groups backend


lukaszgryglicki:
Hi, I'm trying to use `Groups.io` Perceval backend.
The documentation says that I need `GitHub OAuth token` but groups.io doesn't have an option to login via GitHub. I've created an account on groups.io and it can only allow logging in via standard email/password pair or via Google or via Facebook. I've checked all groups.io setting and there is no way to create API token there and their documentation says: `Authentication to the API is performed via HTTP Basic Auth, using a token received from the login API call` [here](https://groups.io/api#authentication-1). SO this is totally unclear what token should I provide.
Next group name, then documentation only says `perceval: error: the following arguments are required: group_name`, what is this group name? I want to get data from zephyr project, what should I pass there?
I've tried:
```
https://lists.zephyrproject.org/g/main/subgroups
https://lists.zephyrproject.org/g/main
https://lists.zephyrproject.org/g
https://lists.zephyrproject.org
lists.zephyrproject.org
zephyrproject
```
None works and the error message is always the same:
`perceval -g groupsio -t `cat /etc/github/oauth` lists.zephyrproject.org --mboxes-path zephyr_groupsio`:
```
[2019-06-18 08:01:44,323 - root - INFO] - Sir Perceval is on his quest.
[2019-06-18 08:01:44,326 - perceval.backends.core.groupsio - INFO] - Looking for messages from 'https://groups.io/g/lists.zephyrproject.org' since 1970-01-01 00:00:00+00:00
[2019-06-18 08:01:44,327 - perceval.backends.core.groupsio - INFO] - Downloading mboxes from 'https://groups.io/g/lists.zephyrproject.org'
[2019-06-18 08:01:44,327 - perceval.backends.core.groupsio - DEBUG] - Storing mboxes in 'zephyr_groupsio'
[2019-06-18 08:01:44,330 - urllib3.connectionpool - DEBUG] - Starting new HTTPS connection (1): api.groups.io
[2019-06-18 08:01:45,076 - urllib3.connectionpool - DEBUG] - https://api.groups.io:443 "GET /v1/downloadarchives?group_name=lists.zephyrproject.org HTTP/1.1" 400 58
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/dist-packages/perceval/backend.py", line 385, in run
    for item in items:
  File "/usr/local/lib/python3.6/dist-packages/perceval/backend.py", line 498, in fetch
    raise e
[...]

PR #539, opened 2019-06-18
[groupsio] Fix description `api_token`

valeriocos:
This code fixes the docstring of the `api_token` variable. Backend version has been changed accordingly: 0.1.4


valeriocos:
This PR is related to https://github.com/chaoss/grimoirelab-perceval/issues/538

valeriocos:
As suggested by @lukaszgryglicki , should we also add some details about how to get the token from groups.io API, @sduenas ?

sduenas:
@valeriocos I think we must, since it's not clear how to use it.

PR #540, opened 2019-06-18
[groupsio] Improve documentation to obtain an API token

valeriocos:
This code enhances the documentation of the groups.io backend by providing the steps to obtain a token from the groups.io API.


valeriocos:
The both places can be good IMHO. Check again the PR, I can remove the first commit if needed.

PR #544, opened 2019-06-19
[meetup] Enable oauth token use

valeriocos:
Due to the deprecation of api key usages, this code allows to use oauth tokens to access the meetup API. The procedure to obtain an oauth token is assisted by the following script:
https://gist.github.com/valeriocos/e16424bc7dc0f2d6dd8bb9295c6f9a4b

Tests have been enhanced accordingly. The backend version is now 0.14.0

### DIFF
--- a/perceval/backends/core/meetup.py
+++ b/perceval/backends/core/meetup.py
@@ -23,6 +23,7 @@
 import json
 import logging
 import requests
+import warnings
 
 from grimoirelab_toolkit.datetime import datetime_to_utc
 from grimoirelab_toolkit.uris import urijoin
@@ -33,6 +34,9 @@ from ...backend import (Backend,
 from ...client import HttpClient, RateLimitHandler
 from ...errors import RepositoryError
 from ...utils import DEFAULT_DATETIME
+
+warnings.simplefilter('always', DeprecationWarning)
+
 
 CATEGORY_EVENT = "event"
 
@@ -59,6 +63,7 @@ class Meetup(Backend):
 
     :param group: name of the group where data will be fetched
     :param api_token: token or key needed to use the API
+    :param is_oauth_token: True if the token is OAuth (default False)
     :param max_items:  maximum number of issues requested on the same query
     :param tag: label used to mark the data
     :param archive: archive to store/retrieve items
@@ -68,7 +73,7 @@ class Meetup(Backend):
     :param sleep_time: time (in seconds) to sleep in case
         of connection problems
     """
-    version = '0.13.0'
+    version = '0.14.0'
 
     CATEGORIES = [CATEGORY_EVENT]
     CLASSIFIED_FIELDS = [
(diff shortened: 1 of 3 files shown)


valeriocos:
Ok, sure @sduenas

valeriocos:
PR updated @sduenas. Most tests are similar to the ones using an `api_key`, probably the test here: https://github.com/chaoss/grimoirelab-perceval/pull/544/files#diff-98a21980071862f1f75ce6c96e041dfcR1125 could be merged with `test_calculate_time_to_reset`, since the rate limit headers don't differ when using oauth tokens and api keys.

perceval/backends/core/meetup.py
```diff
@@ -88,10 +91,18 @@ def __init__(self, group, api_token, max_items=MAX_ITEMS,
...
         self.api_token = api_token
+        self.is_oauth_token = is_oauth_token
         self.sleep_for_rate = sleep_for_rate
         self.min_rate_to_sleep = min_rate_to_sleep
         self.sleep_time = sleep_time
 
+        if not self.is_oauth_token:
+            warnings.warn(
+                "The use of API keys won't be longer supported, "
+                "use Oauth tokens instead",
```
sduenas: Maybe this message is more clear: `API keys are no longer actively supported by Meetup. Use OAuth2 tokens instead`. It's like the message they have in their page.

perceval/backends/core/meetup.py
```diff
@@ -294,6 +305,9 @@ def setup_cmd_parser(cls):
         group.add_argument('--sleep-time', dest='sleep_time',
                            default=SLEEP_TIME, type=int,
                            help="minimun sleeping time to avoid too many request exception")
+        group.add_argument('--is-oauth-token', dest='is_oauth_token',
+                           action='store_true',
+                           help="True if the token is oauth")
```
sduenas: `True` cannot be passed on the command line so something like `set when using OAuth2 tokens` would work.

perceval/backends/core/meetup.py
```diff
@@ -23,6 +23,8 @@
 import json
 import logging
 import requests
+import warnings
+warnings.simplefilter('always', DeprecationWarning)
```
sduenas: Can this go after the import statements?

tests/test_meetup.py
```diff
@@ -1015,6 +1338,23 @@ def test_sanitize_for_archive(self):
         self.assertEqual(headers, s_headers)
         self.assertEqual(payload, s_payload)
 
+    def test_sanitize_for_archive_oauth_token(self):
+        """Test whether the sanitize method works properly when using an oauth token"""
+
+        url = "http://example.com"
+        headers = "headers-information"
+        payload = {'page': 2,
```
sduenas: This should use the same format we have in the file for dicts.

```
dict = {
   'key': value,
    ...
}
```

perceval/backends/core/meetup.py
valeriocos: A new commit updates also the same description in the gitlab backend

perceval/backends/core/meetup.py
```diff
@@ -58,7 +62,8 @@ class Meetup(Backend):
     for authentication with the parameter `api_key`.
 
     :param group: name of the group where data will be fetched
-    :param api_token: token or key needed to use the API
+    :param api_token: oauth token needed to use the API
```
sduenas: This is not only a OAuth token

valeriocos:
No worries, check it again @sduenas

PR #547, opened 2019-07-02
[mediawiki] Skip page without `pageid` attribute

valeriocos:
This PR skips pages not having the `pageid` attribute. This may be the case for pages of type `log` with `actionhidden` attribute. Tests have been added accordingly.

Backend version is now 0.9.6


valeriocos:
Tests for python 3.4 and 3.5 fixed

PR #548, opened 2019-07-05
Add proxy support to git clone command

electrocucaracha:
This change provides support for running `git clone` perceval task behind a corporate proxy. It takes the environment variables if they are defined.


sduenas:
Merged. Thanks!

PR #549, opened 2019-07-09
[jenkins] Support authenticated access using user/API token

valeriocos:
This PR allows to fetch builds and jobs available via a Jenkins server, which requires authenticated access.
Tests have been added accordingly. The backend version has been set to 0.12.0.

### DIFF
--- a/tests/test_jenkins.py
+++ b/tests/test_jenkins.py
@@ -39,9 +39,12 @@ from perceval.backends.core.jenkins import (logger,
                                             JenkinsCommand,
                                             JenkinsClient,
                                             SLEEP_TIME, DETAIL_DEPTH)
+from perceval.errors import BackendError
 from base import TestCaseBackendArchive
 
 
+JENKINS_USER = 'user01'
+JENKINS_TOKEN = 'token01'
 JENKINS_SERVER_URL = 'http://example.com/ci'
 JENKINS_JOBS_URL = JENKINS_SERVER_URL + '/api/json'
 JENKINS_JOB_BUILDS_1 = 'apex-build-brahmaputra'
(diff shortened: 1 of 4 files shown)


tests/test_jenkins.py
```diff
@@ -311,6 +340,62 @@ def test_fetch_depth_2(self):
         self.assertRegex(req.path, '/ci/job')
         self.assertDictEqual(req.querystring, expected)
 
+    @httpretty.activate
+    def test_fetch_auth_api_token(self):
```
sduenas: I don't see anything checking that the client is really sending the authentication as is expected.

tests/test_jenkins.py
```diff
@@ -507,6 +570,21 @@ def test_get_jobs(self):
 
         self.assertEqual(response, body)
 
+    @httpretty.activate
+    def test_get_jobs_auth_api_token(self):
```
sduenas: I think this should check that the token is sent.

valeriocos:
> Besides my comments I miss tests for client.py and the new parameter in fetch.

There are no tests for the params in fetch (beyond GET and POST), so this is the reason why they weren't added. I'll add the missing tests in a different commit.

tests/test_jenkins.py
valeriocos: The token is sent inside a basic authentication, maybe an assertion can be added to check whether in the list of headers there is `Authorization` one (which is must be `Basic`)

sduenas:
> > Besides my comments I miss tests for client.py and the new parameter in fetch.
> 
> There are no tests for the params in fetch (beyond GET and POST), so this is the reason why they weren't added. I'll add the missing tests in a different commit.

It can be integrated in one of the test we have now. Checking the header as you said before should be enough.

valeriocos:
check it again @sduenas

dai-mk:
@valeriocos @sduenas 
How do you use authenticated access? I couldn't find it in the documentation. https://github.com/chaoss/grimoirelab-sirmordred#jenkins-

PR #553, opened 2019-07-15
[gitlab] Recalculate outdated list of merge requests

sduenas:
If during the fetching process, a merge request was modified, the initial list of merge requests is outdated. Due to it, returned items will not be ordered, which can cause several problems when
an incremental retrieval is performed.

To avoid this problem, when the backend detects that an MR has been updated, the process will not be cancelled but it will request a new list, starting on the 'updated_on' field of the last item returned.

Backend version updated to 0.8.0.


tests/test_gitlab.py
```diff
@@ -836,6 +906,46 @@ def test_fetch_merges_enterprise(self):
...
 
+    @httpretty.activate
+    def test_fetch_merges_outdated_list(self):
+        """Test if a new MR list is requested when the original is outdated"""
+
+        setup_gitlab_outdated_mrs_server(GITLAB_URL_PROJECT, GITLAB_MERGES_URL)
+
+        gitlab = GitLab("fdroid", "fdroiddata", "your-token")
+
+        merges = [merges for merges in gitlab.fetch(category=CATEGORY_MERGE_REQUEST)]
```
valeriocos: we could check that a log message has been written:
```
        with self.assertLogs(logger, level='DEBUG') as cm:
            merges = [merges for merges in gitlab.fetch(category=CATEGORY_MERGE_REQUEST)]

            self.assertEqual(cm.output[1], ....)
```

sduenas:
Comments addressed @valeriocos, please check it again.

PR #554, opened 2019-07-19
[backend] Add search terms

valeriocos:
This PR proposes to extend the metadata attributes by adding a new field (`search_terms`) with the purpose of simplifying query operations and avoding inspecting the data attribute. 

The search terms are included in a dict with the following shape:
```
        {
            'term-1': 'value-1',
            'term-2': 'value-2',
            'term-3': 'value-3',
        }
```
The search terms are added to the metadata information of each item in `search_terms` attribute. If search terms are not defined, `search_terms` will be set to `None`.

This new feature is showcased with the Jira backend.

### DIFF
--- a/perceval/backend.py
+++ b/perceval/backend.py
@@ -74,6 +74,19 @@ class Backend:
     Classified data filtering and archiving are not compatible to prevent
     data leaks or security issues.
 
+    Each backend can also provides a set of search terms to simplify query
+    operations (avoiding the manual inspection of the items). The search
+    terms are included in a dict with the following shape:
+        {
+            'term-1': 'value-1',
+            'term-2': 'value-2',
+            'term-3': 'value-3',
+        }
+
+    The search terms are added to the metadata information of each
+    Perceval item in `search_terms` attributes. If `search_terms` is not set,
+    it will be set to `None` .
+
     :param origin: identifier of the repository
     :param tag: tag items using this label
     :param archive: archive to store/retrieve data
@@ -81,7 +94,7 @@ class Backend:
     :raises ValueError: raised when `archive` is not an instance of
         `Archive` class
     """
-    version = '0.8.0'
+    version = '0.9.0'
 
     CATEGORIES = []
     CLASSIFIED_FIELDS = []
@@ -227,6 +240,7 @@ class Backend:
             'updated_on': self.metadata_updated_on(item),
             'classified_fields_filtered': self.classified_fields if filter_classified else None,
             'category': self.metadata_category(item),
+            'search_terms': self.search_terms(item),
             'tag': self.tag,
             'data': item,
         }
@@ -252,6 +266,10 @@ class Backend:
     @staticmethod
     def metadata_category(item):
         raise NotImplementedError
+
+    @staticmethod
+    def search_terms(item):
+        return None
 
     def _init_client(self, from_archive=False):
         raise NotImplementedError
--- a/perceval/backends/core/jira.py
+++ b/perceval/backends/core/jira.py
@@ -96,7 +96,7 @@ class Jira(Backend):
     :param tag: label used to mark the data
     :param archive: archive to store/retrieve items
     """
-    version = '0.12.0'
+    version = '0.13.0'
 
     CATEGORIES = [CATEGORY_ISSUE]
 
@@ -215,6 +215,22 @@ class Jira(Backend):
         'issue'.
         """
         return CATEGORY_ISSUE
+
+    @staticmethod
+    def search_terms(item):
+        """Set the search terms of a Jira items."""
+
+        project_id = item['fields']['project']['id']
+        project_key = item['fields']['project']['key']
+        project_name = item['fields']['project']['name']
+
+        terms = {
+            'project_id': project_id,
+            'project_key': project_key,
+            'project_name': project_name
+        }
+
+        return terms
 
     @staticmethod
     def parse_issues(raw_page):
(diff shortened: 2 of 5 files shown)


valeriocos:
An initial evaluation of the new feature has been done to measure the impact on the raw indexes in GrimoireLab. The evaluation consisted in running the raw collection on a Jira with >10000 issues using 
the old and search terms approach, and then measure the size of the two indexes. The table below (derived using `_cat/indices`)  shows that around 32% of disk storage would be saved with the search terms approach.

| index        | docs.count | size |
| ------------- |:-------------:| -----:| 
| jira_search | 10796 | 73.8mb |
| jira_old | 10796 | 109.1mb |

Furthermore, the additional benefits of the search terms are:
- it's not needed anymore to index all information coming from a data source (but just the one at metadata level). See below an example of new mapping:
```
{
     "dynamic":true,
     "properties": {
         "data": {
             "dynamic":false,
             "properties": {}
            }
      }
}
```
- the raw index mapping in GrimoireLab would be homogeneous across all data sources (see below), thus opening the doors to new ways of storing raw data.
- the codebase of ELK would be reduced (mappings and `_fix_item` method would not be needed anymore).
- errors such immense term exceptions, too many attribute to index, mapping attribute inconsistencies would disappear, thus making the platform more robust.

valeriocos:
To use the search_terms in the filter raw a minor change is needed in the code [here](https://github.com/chaoss/grimoirelab-elk/blob/master/grimoire_elk/elastic_items.py#L35) (`FILTER_DATA_ATTR = 'search_terms.'`). Beyond this, the configurations are pretty the same:

```
"jira": [
            "https://xxx.jira.com --filter-raw=search_terms.project_key:ACME"
        ]
```

```
"jira": [
            "https://xxx.jira.com --filter-raw=data.fields.project.key:ACME"
        ]
```

valeriocos:
The PR is ELK is available at: https://github.com/chaoss/grimoirelab-elk/pull/665

perceval/backend.py
```diff
@@ -74,14 +74,27 @@ class Backend:
     Classified data filtering and archiving are not compatible to prevent
     data leaks or security issues.
 
+    Each backend can also provides a set of search terms to simplify query
```
sduenas: I would changed this with:

```suggestion
    Each backend might provide a set of search terms to simplify query
```

perceval/backend.py
```diff
@@ -74,14 +74,27 @@ class Backend:
...
+    operations (avoiding the manual inspection of the items). The search
+    terms are included in a dict with the following shape:
+        {
+            'term-1': 'value-1',
+            'term-2': 'value-2',
+            'term-3': 'value-3',
+        }
+
+    The search terms are added to the metadata information of each
```
sduenas:
```suggestion
    These terms are added to the metadata information of each
```

perceval/backend.py
```diff
@@ -74,14 +74,27 @@ class Backend:
...
+    Perceval item in `search_terms` attributes. If `search_terms` is not set,
```
sduenas:
```suggestion
    item in `search_terms` attribute. If `search_terms` is not set,
```

perceval/backend.py
```diff
@@ -253,6 +267,10 @@ def metadata_updated_on(item):
     def metadata_category(item):
         raise NotImplementedError
 
+    @staticmethod
+    def search_terms(item):
```
sduenas: Wouldn't it be better to use the same behaviour we have for `classified_fields`?.

We have a common method in `Backend` class that removes those fields. They are defined in `CLASSIFIED_FIELDS` class variable. I think it will be better to have something similar, so we just need a piece of code to copy these fields and then each backend would define which fields are going to be included.

perceval/backends/core/jira.py
```diff
@@ -216,6 +216,22 @@ def metadata_category(item):
...
+    @staticmethod
+    def search_terms(item):
+        """Set the search terms of a Jira items."""
+
+        project_id = item['fields']['project']['id']
+        project_key = item['fields']['project']['key']
+        project_name = item['fields']['project']['name']
+
+        terms = {
+            'project_id': project_id,
```
sduenas: Does make sense to include the `id` of the item too? The one returned by `metadata_id()`.

tests/test_jira.py
```diff
@@ -328,6 +328,86 @@ def request_callback(method, uri, headers):
...
+
+        httpretty.register_uri(httpretty.GET,
+                               JIRA_FIELDS_URL,
+                               body=body, status=200)
+
+        jira = Jira(JIRA_SERVER_URL)
+        issues = [issue for issue in jira.fetch()]
+
+        issue = issues[0]
+        self.assertEqual(issue['uuid'], '6a7ba2a01aee56603b9d8a5f6b40c843fc089b2f')
```
sduenas: To align this section with the other ones you should also add this line:

```
self.assertEqual(issue['origin'], 'http://example.com')
```

perceval/backend.py
sduenas: Would it be better to use `search_fields` instead of `search_terms`? I'm saying this because we have `classified_fields`, so maybe it makes more sense to align it with that concept too.

perceval/backend.py
valeriocos: I'm not sure about the use of `might` (which expresses permission) over `can` (which expresses ability/capability) in this sentence.
I'll fix the typo (`provides`)

perceval/backend.py
valeriocos: I will give it a try

valeriocos:
I have prepared another PR addressing your comments: https://github.com/chaoss/grimoirelab-perceval/pull/560, so we can better evaluate the two solutions.

valeriocos:
Closing this PR since PR #560 seems better

PR #555, opened 2019-07-21
[backend] Include summary of fetch executions

valeriocos:
This PR allows to calculate and return the summary of fetch executions. By default, the summary (defined as a class) models information about the total number of items fetched, the min and max date of the items fetched and the last UUID. The summary can be tweaked by each backend to include the number of skipped items and some extras information specific for the backend (e.g., offset, new branches added in Git).

Using the param `summary` (default `False`), the summary can be outputted as JSON, as the last item
of a fetch execution.


valeriocos:
Tests fail due to a wrong metadata_updated_on value in the test data (fixed with: https://github.com/chaoss/grimoirelab-perceval/pull/556)

perceval/backend.py
```diff
@@ -510,6 +539,157 @@ def setup_cmd_parser(cls):
...
+        self._items_skipped = 0
+        self._min_updated_on = None
+        self._max_updated_on = None
+        self._last_updated_on = None
+        self._last_uuid = None
+        self._min_offset = None
+        self._max_offset = None
+        self._extras = None
+
+    @property
```
sduenas: I think we can remove all this setters and getters. Since the main method to update the summary is `update` and it encapsulates all the logic, I don't think we need all this code.

What it would be useful is to have a property like `total` which returns the sum of `fetched` and `skipped`.

perceval/backend.py
```diff
@@ -510,6 +539,157 @@ def setup_cmd_parser(cls):
...
+        self.min_updated_on = updated_on if not self.min_updated_on else min(self.min_updated_on, updated_on)
+        self.max_updated_on = updated_on if not self.max_updated_on else max(self.max_updated_on, updated_on)
+        self.last_updated_on = updated_on
+
+        offset = item.get('offset', None)
+        if offset is not None:
+            self.min_offset = offset if self.min_offset is None else min(self.min_offset, offset)
+            self.max_offset = offset if self.max_offset is None else max(self.max_offset, offset)
+
+    def to_json(self):
```
sduenas: This should be better named to `to_dict`.

perceval/backend.py
```diff
@@ -589,7 +778,7 @@ def fetch(backend_class, backend_args, category, filter_classified=False,
 
 
 def fetch_from_archive(backend_class, backend_args, manager,
-                       category, archived_after):
+                       category, archived_after, summary=False):
```
sduenas: I think we should always produce this output when running the command.

perceval/backend.py
```diff
@@ -581,6 +765,11 @@ def fetch(backend_class, backend_args, category, filter_classified=False,
     try:
         for item in items:
             yield item
+
+        if summary:
+            summary_json = backend.summary.to_json()
+            summary_str = json.dumps(summary_json, indent=4, sort_keys=True)
+            sys.stderr.write(summary_str)
```
sduenas: I would like to have a better formatted output instead of having a JSON.

perceval/backend.py
```diff
@@ -510,6 +539,157 @@ def setup_cmd_parser(cls):
         raise NotImplementedError
 
 
+class Summary:
```
sduenas: You can work in the documentation of this method when you can.

valeriocos:
please @sduenas check it again

perceval/backend.py
```diff
@@ -581,6 +701,9 @@ def fetch(backend_class, backend_args, category, filter_classified=False,
     try:
         for item in items:
             yield item
+
+        summary = repr(backend.summary)
+        sys.stderr.write(summary)
```
sduenas: I don't really like this. We need to find a way to provide access  to the summary object from this function. Others will find this useful. For example, in Arthur, this function is called by workers. The workers generate a result that is quite similar to this summary object. Instead of generating a new object, it will be great if the object is the same.

I think we should create a new function for this purpose, something that can return a generator of items and the result. Not sure if with a callable we can do it.

valeriocos:
Yes, I like the idea. 
Should the functions `fetch_from_archive` and `fetch` be removed and converted to private methods of `BackendItemsGenerator`?

sduenas:
> Yes, I like the idea.
> Should the functions `fetch_from_archive` and `fetch` be removed and converted to private methods of `BackendItemsGenerator`?

I'd rather not. Let's keep these functions as they are now, adding a deprecation message. You can add similar private methods to this new backend to do the similar process.

valeriocos:
To live-test `BackendItemsGenerator`, the function `run` should be modified in the following way:
```
def run(self):
    """Fetch and write items.

    This method runs the backend to fetch the items from the given
    origin. Items are converted to JSON objects and written to the
    defined output.

    If `fetch-archive` parameter was given as an argument during
    the inizialization of the instance, the items will be retrieved
    using the archive manager.
    """
    backend_args = vars(self.parsed_args)
    category = backend_args.pop('category', None)
    filter_classified = backend_args.pop('filter_classified', False)
    archived_since = backend_args.pop('archived_since', None)

    fetch_archive = False
    if self.archive_manager and self.parsed_args.fetch_archive:
        fetch_archive = True

    big = BackendItemsGenerator(self.BACKEND, backend_args, category, filter_classified,
                                self.archive_manager, archived_since, fetch_archive)

    try:
        for item in big.items():
            if self.json_line:
                obj = json.dumps(item, separators=(',', ':'), sort_keys=True)
            else:
                obj = json.dumps(item, indent=4, sort_keys=True)
            self.outfile.write(obj)
            self.outfile.write('\n')
            
        res = big.results()
    except IOError as e:
        raise RuntimeError(str(e))
    except Exception as e:
        raise RuntimeError(str(e))
```

perceval/backend.py
```diff
@@ -533,6 +533,111 @@ def setup_cmd_parser(cls):
         raise NotImplementedError
 
 
+class BackendItemsGenerator:
+    def __init__(self, backend_class, backend_args, category, filter_classified=False,
+                 manager=None, archived_after=None, fetch_archive=False):
+        self.backend_class = backend_class
```
sduenas: We don't need to store most of these attributes. They'll be part of the backend. Most of the code of `init_backend` should be here.

perceval/backend.py
```diff
@@ -533,6 +533,111 @@ def setup_cmd_parser(cls):
         raise NotImplementedError
 
 
+class BackendItemsGenerator:
```
sduenas: This class should be a context manager, so it should contain the required methods, `__enter__` and `__exit__`.

perceval/backend.py
```diff
@@ -533,6 +533,111 @@ def setup_cmd_parser(cls):
...
+    def items(self):
+        if not self.fetch_archive:
+            items = self.__fetch(self.backend_args, self.category,
+                                 self.filter_classified, self.manager)
+        else:
+            items = self.__fetch_from_archive(self.manager, self.category, self.archived_after)
+
+        return items
+
+    def results(self):
```
sduenas: Let's rename this to summary.

perceval/backend.py
sduenas: The iterator can also created here if we consider it makes sense.

valeriocos:
To live-test `BackendItemsGenerator`, the function `run` should be modified in the following way:

```
def run(self):
    """Fetch and write items.

    This method runs the backend to fetch the items from the given
    origin. Items are converted to JSON objects and written to the
    defined output.

    If `fetch-archive` parameter was given as an argument during
    the inizialization of the instance, the items will be retrieved
    using the archive manager.
    """
    backend_args = vars(self.parsed_args)
    category = backend_args.pop('category', None)
    filter_classified = backend_args.pop('filter_classified', False)
    archived_since = backend_args.pop('archived_since', None)

    fetch_archive = False
    if self.archive_manager and self.parsed_args.fetch_archive:
        fetch_archive = True

    with BackendItemsGenerator(self.BACKEND, backend_args, category, self.archive_manager,
                               fetch_archive, filter_classified, archived_since) as big:
        try:
            for item in big.items:
                if self.json_line:
                    obj = json.dumps(item, separators=(',', ':'), sort_keys=True)
                else:
                    obj = json.dumps(item, indent=4, sort_keys=True)
                self.outfile.write(obj)
                self.outfile.write('\n')
        except IOError as e:
            raise RuntimeError(str(e))
        except Exception as e:
            raise RuntimeError(str(e))
```

valeriocos:
Tests have been added, please @sduenas review it.

perceval/backend.py
```diff
@@ -510,6 +533,206 @@ def setup_cmd_parser(cls):
...
+            self.backend = backend_class(**init_args)
+            items = self.__fetch_from_archive(manager, category, archived_after)
+
+        self.items = items
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        pass
```
sduenas: Does it make sense to set the summary and the backend to `None`?

perceval/backend.py
```diff
@@ -510,6 +533,206 @@ def setup_cmd_parser(cls):
...
+
+    def summary(self):
```
sduenas: I think this should be a `property` so one can overwrite it.

perceval/backend.py
```diff
@@ -510,6 +533,206 @@ def setup_cmd_parser(cls):
...
+        """Return the summary object in the backend"""
```
sduenas: I wouldn't make a reference to the backend. Say something like this is the summary of the last execution or similar.

perceval/backend.py
```diff
@@ -510,6 +533,206 @@ def setup_cmd_parser(cls):
...
+        return self.backend.summary
+
+    def __fetch(self, backend_args, category, filter_classified=False,
+                manager=None):
+        """Fetch items using the given backend.
+
+        Generator to get items using the given backend class. When
```
sduenas: Part of this documentation should go to the description of the class too.

On this docstring itself, you shouldn't mention to the backend_class nor to the given backend. That was part of the old function doc.

perceval/backend.py
```diff
@@ -510,6 +533,206 @@ def setup_cmd_parser(cls):
...
+        except Exception as e:
+            if manager:
+                archive_path = self.backend.archive.archive_path
+                manager.remove_archive(archive_path)
+            raise e
+
+    def __fetch_from_archive(self, manager, category, archived_after):
+        """Fetch items from an archive manager.
+
+        Generator to get the items of a category (previously fetched
```
sduenas: Same comments as above.

perceval/backend.py
```diff
@@ -510,6 +533,206 @@ def setup_cmd_parser(cls):
...
+    This class models the summary of a fetch execution. It includes
+    the last UUID, number of items fetched, skipped and their sum,
+    plus the min, max and last updated_on times. Furthermore, for
+    backends using offsets, the corresponding summary contains the
+    min and max offsets retrieved. Finally, the summary also includes
+    some extra fields, which can be used by any backend to include
+    fetch-specific information.
+    """
+    def __init__(self):
+        self.fetched = 0
```
sduenas: `fetched` should be renamed to something else. The backend might fetch a lot of items but for many reasons they wouldn't be generated.

perceval/backend.py
```diff
@@ -510,6 +533,206 @@ def setup_cmd_parser(cls):
...
+                'last_uuid': self.last_uuid,
+                'min_offset': self.min_offset,
+                'max_offset': self.max_offset,
+                'extras': self.extras
+            }
+        }
+
+        return summary
+
+    def __repr__(self):
```
sduenas: I don't really like to have this here. `__repr__` should write something less complex. I don't think we need it on this case.

perceval/backend.py
```diff
@@ -510,6 +533,206 @@ def setup_cmd_parser(cls):
...
+        self.min_updated_on = updated_on if not self.min_updated_on else min(self.min_updated_on, updated_on)
+        self.max_updated_on = updated_on if not self.max_updated_on else max(self.max_updated_on, updated_on)
+        self.last_updated_on = updated_on
+
+        offset = item.get('offset', None)
+        if offset is not None:
+            self.min_offset = offset if self.min_offset is None else min(self.min_offset, offset)
+            self.max_offset = offset if self.max_offset is None else max(self.max_offset, offset)
+
+    def to_dict(self):
```
sduenas: Not sure if we need this either.

perceval/backend.py
sduenas: With the changes you made this wouldn't be needed anymore here. The code that will write the summary will be either in the command or in the `perceval` script. Probably the summary should be written in the log.

perceval/backend.py
```diff
@@ -623,6 +853,9 @@ def fetch_from_archive(backend_class, backend_args, manager,
         try:
             for item in items:
                 yield item
+
+            summary = repr(backend.summary)
```
sduenas: Same as above.

tests/test_backend.py
```diff
@@ -1248,6 +1760,35 @@ def test_items(self):
             self.assertEqual(item['tag'], 'test')
             self.assertEqual(item['classified_fields_filtered'], None)
 
+    @unittest.mock.patch('perceval.backend.sys.stderr')
```
sduenas: These kind of tests can be removed. What I miss is a test for `TestBackend` to check that the summary object contains the data.

perceval/backend.py
```diff
@@ -510,6 +533,206 @@ def setup_cmd_parser(cls):
...
+            'summary': {
+                'total': self.total,
+                'fetched': self.fetched,
+                'skipped': self.skipped,
+                'min_updated_on': self.min_updated_on.isoformat() if self.min_updated_on else None,
+                'max_updated_on': self.max_updated_on.isoformat() if self.max_updated_on else None,
+                'last_updated_on': self.last_updated_on.isoformat() if self.last_updated_on else None,
+                'last_uuid': self.last_uuid,
+                'min_offset': self.min_offset,
+                'max_offset': self.max_offset,
```
sduenas: Maybe `last_offset` does make sense. We can find contradictions in data having that.

perceval/backend.py
valeriocos: I don't see this change, `fetched` seems a reasonable definition.

PR #559, opened 2019-07-22
[backend] Add generic blacklist param

valeriocos:
This PR extends the BackendCommandArgumentParser by including the param `blacklist`, which allows
to define a list of items IDs (parsed as `str`) to be filtered out.

Tests and the GitLab backend have been modified accordingly. The backend version is set to 0.9.0.

### DIFF
--- a/perceval/backend.py
+++ b/perceval/backend.py
@@ -81,7 +81,7 @@ class Backend:
     :raises ValueError: raised when `archive` is not an instance of
         `Archive` class
     """
-    version = '0.8.0'
+    version = '0.9.0'
 
     CATEGORIES = []
     CLASSIFIED_FIELDS = []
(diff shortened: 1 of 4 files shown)


perceval/backend.py
```diff
@@ -327,6 +328,11 @@ def __init__(self, categories, from_date=False, to_date=False, offset=False,
                                type=int, default=0,
                                help="offset to start fetching items")
 
+        if blacklist:
+            group.add_argument('--blacklist-ids', dest='blacklist_ids',
+                               nargs='*', type=str,
```
sduenas: Does this make sense to be a list of strings?

valeriocos: I guess so, since it is more inclusive, however it may not be applicable for all backends (gitlab, github which IDs are numerics). An example of non-numeric IDs are commit hashes. Maybe we can re/define the `--blacklist-ids` for each backend.

sduenas: There are no many backends with a blacklist: jenkins, gerrit and gitlab so far. So, it's possible to use the same parameter. My main concern is to make clear what `ids` mean for each case.

valeriocos: A new proposal is available at: https://github.com/chaoss/grimoirelab-perceval/pull/575. The idea is to compare the approaches and decide how to proceed.

PR #560, opened 2019-07-25
[backend] Add search fields in metadata items

valeriocos:
This code enhances the metadata information with a set of search fields, which simplify query operations and avoid the manual inspections of the Perceval items.
    
These fields are added to the item metadata information in the `search_fields` attribute. By default, `search_fields` contains the id of the item ('item_id': item_id_value), obtained via the `metadata_id` method). However each backend can set extra search fields using the dict EXTRA_SEARCH_FIELDS. An example of EXTRA_SEARCH_FIELDS is provided below:
  
```
       {
          'project_id': ['fields', 'project', 'id'],
          'project_key': ['fields', 'project', 'key'],
          'project_name': ['fields', 'project', 'name']
       }
```
Each key in the dict is a search field to be included in the item metadata information, while the corresponding value is a list that stores the "path" of the search field value within the item.
    
 The backend version is set to 0.9.0.


perceval/backend.py
```diff
@@ -207,6 +236,42 @@ def filter_classified_data(self, item):
...
+        }
+
+        logger.debug("Adding extra search fields to item %s", item_uuid)
+        for sf in self.EXTRA_SEARCH_FIELDS:
+            try:
+                search_field = self.EXTRA_SEARCH_FIELDS[sf]
+                field_value = _find_value_from_nested_dict(item, search_field)
+                search_fields[sf] = field_value
+            except KeyError:
+                logger.debug("Extra search field '%s' not found for item %s; field ignored",
```
sduenas: I guess this cases should be not common. I think this should be a warning.

valeriocos: Thanks @sduenas! 
Sure! Should I updated also this line: https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/backend.py#L203 ?

sduenas: No I think that case is different. From my point of view, I think all the fields included in the `search_fields` should be mandatory and available on all items. In the case of `filtered_classified` those fields may appear or not, cause depending on how you get the data (for example, using authentication in bugzilla you get the email addresses of the users) you can have those values or not.

valeriocos: PR updated

valeriocos:
The PR proposes an initial set of search fields for every backend. Please @sduenas @alpgarcia have a look when you have time.

valeriocos:
I agree with your comments, we can go backend by backend and decide what to add as search terms. Feel free to make a pass on the current search terms proposals and leave feedback. 

As reference, examples of Perceval data is available at: https://github.com/chaoss/grimoirelab-elk/tree/master/tests/data

sduenas:
> I agree with your comments, we can go backend by backend and decide what to add as search terms. Feel free to make a pass on the current search terms proposals and leave feedback.
> 
> As reference, examples of Perceval data is available at: https://github.com/chaoss/grimoirelab-elk/tree/master/tests/data

Ok. Let's do that.

valeriocos:
ok, great. I wait for your feedback then

Issue #563, opened 2019-08-01
[feature-request] Fetch events for issues


Polaris000:
Apart from the information fetched for issues (`<api_url>/issues/<issue_num>`), it may be a good idea to also fetch event information, or atleast the last re-open event for issues (`<api_url>/issues/<issue_num>/events`). 
Apart from the `created_at`, `closed_at` and `updated_at` fields, there could be an `events` field added to the issue data Perceval fetches.

Though I couldn't find any call to the `events` url in the github backend module, please let me know if this feature already exists.

Currently, when reopened issues are fetched by Perceval, the `status` field is `open` but there is a `closed_by` field as well. That could be one way to tell which issues were reopened, but how do you find the date of reopening?

Thanks in advance @valeriocos , @sduenas!

valeriocos:
Hi @Polaris000 , a PR addressing your request has been just made available.
I noticed some inconsistency in the documentation at: https://developer.github.com/v3/activity/events/#list-public-events. In particular, it is reported that:
```
Events support pagination, however the per_page option is unsupported. The fixed 
page size is 30 items. Fetching up to ten pages is supported, for a total of 300 events.

Only events created within the past 90 days will be included in timelines. Events older 
than 90 days will not be included (even if the total number of events in the timeline is 
less than 300).
``` 

However from the tests I did (fetching events from grimoirelab-perceval repo), I was able to fetch events that happened before 90 days ago. Since you are going to use the code in the PR, could you provide some feedback?

Thanks

Polaris000:
I'll have a look, thanks @valeriocos.

sduenas:
As I said in #565, we'll keep this on hold until there's a solution that doesn't break Perceval.

Issue #160, opened 2017-09-19
Support GitHub v4 API


jgbarah:
There is a [new GitHub API](https://githubengineering.com/the-github-graphql-api/), based on GraphQL ([documentation](https://developer.github.com/v4/)). For some of the stuff Perceval does, it could be more appropriate than v3 (which is the one we're currentĺy using).

Besides, there is at least one Python package, [graphene](http://graphene-python.org/), which could help a lot with the implementation.

It would be nice if at least we explore to which extent this API could be useful.

valeriocos:
Graphene seems to be used more on the server side (build/test APIs with GraphQL). Conversely, calls to GitHub v4 API are made using common HTTP requests which contain in their payload GraphQL code.

I've played a bit with the new API and pushed some examples at [github-v4-playground](https://github.com/valeriocos/perceval/blob/github-v4-explore/tests/test_github_v4.py)

As a first impression, the advantages to use the new API are:
- Live test of GraphQL queries (see [explorer](https://developer.github.com/v4/explorer/))
- Get nested data in one single query (ex.: get author and assignees for an issue)

The limitations are: 
- GraphQL requires to be explicit about the fields returned by a query (there is nothing similar to SELECT * in SQL), so in some cases we could have pretty long queries.
- It doesn't seem possible to retrieve nested paginated data in a single query. For instance, issues are generally paginated and some of them can have paginated comments. In this case, two queries should be written to deal with the pagination on issues and comments (as similarly done with REST API calls).

acs:
Closing this ticket after the evaluation of GraphQL. We don't have plans to use it in the near future.

PR #565, opened 2019-08-13
[github] Add events to fetched issues

valeriocos:
This PR addresses https://github.com/chaoss/grimoirelab-perceval/issues/563. It includes the events related to a given issue, they are stored in the attribute `events_data`.
Backend version is set to 0.23.0 and tests have been updated accordingly.

This PR aims at unblocking certain tasks planned for the GSoC program (more details from @Polaris000), however its inclusion to the code base is not mandatory and can be discussed.


Polaris000:
Thanks a lot @valeriocos!

valeriocos:
please @sduenas share your thoughts on this PR when you have time

Polaris000:
@valeriocos , maybe we could specify this as an optional parameter (which a note saying that this would slow down the fetch..?) ? WDYT?
This was suggested by @GeorgLink .

GeorgLink:
We had a conversation about this on the CHAOSS mailing list.

The idea is to maintain the event data collection as a regular part of
Perceval and deactivate it by default.
Managing it as a separate branch increases the overhead for keeping it in
sync with the overall development.
Keeping it separate also has the danger that the feature will not be
maintained at all and all the work was for nothing.
the compromise is to include it in the master branch development and simply
deactivate it by default, but anyone who wants to use it can do so.

valeriocos:
Is this the chaoss mailing list: chaoss@lists.linuxfoundation.org? I have missed this conversation.

One of the principle of Perceval is to fetch always the same data from a given data source, so I'm not sure about the flag option to deactivate/activate events info. I guess this PR will be merged to the code base, but the final decision is in @sduenas hand.

valeriocos:
Thanks for the pointers @GeorgLink , let's move the conversation to just one place.

Georg
```
>>> What are your thoughts on adding issue events to Perceval?
>>> Will this be always in a separate branch or be merged back into master?
>>> Pros/cons?
```

```
>> Thanks Georg for the reply!
>> Regarding the issue events, I agree with Valerio, in that, they will make
>> fetching data inefficient. Hence, I don't see this change getting merged
>> with grimoirelab-perceval's master branch. The sheer number of events per
>> issue as well as the extra API calls become exponentially more time
>> consuming for larger repositories. The obvious advantage for smaller
>> repositories would be access to the extra events information. Maybe if
>> GitHub finds some way to add more information to the main issue url
>> (<api_url>/issues/<issue_num>), that could be useful. For now though, issue
>> metrics would probably require working on the separate branch :( Open for
>> any good ideas though!

**Comment by valeriocos** [comment]
```
>>> On Sun, Aug 18, 2019 at 10:45 AM Aniruddha Karajgi <akarajgi0 at gmail.com>
>>> wrote:
>>>
>>>> This is a weekly update with the meeting log and corresponding blog
>>>> post for the meet we had on #grimoirelab channel (freenode IRC) on the 13th
>>>> of August.
>>>>
>>>> Also, regular post updates will be added as comments to the following
>>>> issue in governance repository:
>>>> https://github.com/chaoss/governance/issues/135
>>>>
>>>> Other links are below:
>>>> - post:
>>>> https://polaris000.github.io/blog/week_9
>>>>
>>>> - meeting log:
>>>> https://github.com/Polaris000/GSoC_19_Perceval_Implementations/blob/master/IRC_chats/aug_13_week9.txt
>>>>
>>>> All GSoC related posts can be found here:
>>>> https://polaris000.github.io/tags/GSoC
>>>>
>>>> Please upvote if you liked the post!
>>>> Thanks!
```

```
>> On Mon, Aug 19, 2019 at 10:43 PM Georg Link <linkgeorg at gmail.com> wrote:
>>
>>> Aniruddha,
>>>
>>> Thanks for sharing. You are almost at the finish line.
>>>
>>> What are your thoughts on adding issue events to Perceval?
>>> Will this be always in a separate branch or be merged back into master?
>>> Pros/cons?
>>>
>>> Best,
>>> Georg
```

```
> On Mon, Aug 19, 2019 at 10:58 AM Aniruddha Karajgi <akarajgi0 at gmail.com>
> wrote:
>
>> Thanks Georg for the reply!
>> Regarding the issue events, I agree with Valerio, in that, they will make
>> fetching data inefficient. Hence, I don't see this change getting merged
>> with grimoirelab-perceval's master branch. The sheer number of events per
>> issue as well as the extra API calls become exponentially more time
>> consuming for larger repositories. The obvious advantage for smaller
>> repositories would be access to the extra events information. Maybe if
>> GitHub finds some way to add more information to the main issue url
>> (<api_url>/issues/<issue_num>), that could be useful. For now though, issue
>> metrics would probably require working on the separate branch :( Open for
>> any good ideas though!
```

```
On Tue, Aug 20, 2019 at 7:31 PM Georg Link <linkgeorg at gmail.com> wrote:

> Aniruddha,
>
> Thank you for the reasoned thoughts on the subject. I need a bit more
> information to form my own opinion.
>
> 1) "sheer number of events per issue" - how many are we talking about on
> average or what is the median?
>
> 2) "extra API calls" - how many more calls are needed? One extra call per
> event; per issue; per ___?
>
> 3) "API calls become exponentially more time consuming for larger
> repositories" - I assume you have done some tests. What did you find for
> small, medium, and large repository?
>
> 4) "obvious advantage ... would be to access the extra events information"
> - What information is this? I'm not familiar with the API, so I don't know
> what specifically we gain.
>
> 5) "if GitHub finds some way to add more information" - Again, I'm not an
> expert on the GitHub API. Have you explored the GraphQL interface for
> retrieving this information?
>
> I believe if we get data for the above questions, then we can make an
> informed decision on whether it is worth integrating this feature or not.
>
> My biggest concern with having a separate branch is that it will be out of
> date very soon and unless we upstream the feature into the master branch,
> it will become useless and your work will be lost to us.
> What is your opinion on including the "extra events" as an optional
> feature that users can activate when needed but by default does not slow
> down data collection as it is now.
>
> Best,
> Georg
```

```
Hi Georg. Thanks for your queries.

Though I don't think I have the answer to all of your questions, I'll try
my best!

1. I don't have an exact number, but events include everything from
re-opening and closing to referencing, tagging, assigning, etc, and this
can be a very large number for active projects. More info can be found here
<https://developer.github.com/v3/issues/events/>.
2. I'm not too sure about this one, but for each issue, you have to
generally fetch the following (maybe Valerio can help you out with the
specifics):
- the general information for that issue
- the events
- comments information
- reactions
- comment reactions
- assignees data
3. No, I have not run any tests. Fetching so much information would add up
to a lot more API calls for large repositories
4. When you had asked this question, I thought you were asking about the
advantage of fetching events. I meant that the extra events information can
help a lot, since it gives a better picture of how often issues are
updated, rather than the updated_at field, which just gives the date of the
last update.
5. So, Perceval tries to fetch as much information as possible in the least
number of API calls. Most useful information is present on the url
<api_url>/issues/<issue_num>. If GitHub adds events information to this, it
would be better, since it would reduce the number of calls made.

My biggest concern with having a separate branch is that it will be out of
date very soon and unless we upstream the feature into the master branch,
it will become useless and your work will be lost to us.
Yes, that is a concern.

What is your opinion on including the "extra events" as an optional feature
that users can activate when needed but by default does not slow down data
collection as it is now.
That seems to be a good idea. May be we can ask Valerio if he would like
this to be added to Perceval as an option.
```

valeriocos:
Before including the events as optional parameter, we may want to explore other alternatives @sduenas . They aim at keeping coherence with the other backends (i.e., fetching always the same data from a given data source).

The endpoint [events](https://developer.github.com/v3/activity/events/#events) is really basic and it doesn't seem to include parameters to retrieve a subset of events types or after/since a given date. Thus, we could contact the GitHub support and ask to include these parameters. If this request is accepted, we could include a limited set of events when fetching GItHub issues (taking into account @Polaris000 needs) .

As mentioned by @Polaris000, fetching issue events is highly inefficient and would affect the overall fetching process due to token consumption. Furthermore, for every issue updated, Perceval will retrieve the same events over and over. Under these conditions, it would be probably better to create a new category (events) and fetch [all issues events for a repository](https://developer.github.com/v3/activity/events/#list-issue-events-for-a-repository) 

WDYT @sduenas ?

GeorgLink:
I don't know what all you have tried. Would it be more efficient to ask for the data using the GraphQL interface, instead of the REST API?

valeriocos:
We evaluated graphql time ago (https://github.com/chaoss/grimoirelab-perceval/issues/160), but at the end we discard it.
It could be a nice experiment, maybe as a specific backend.

GeorgLink:
Thanks for the background. Any particular reason why GraphQL was dismissed?

valeriocos:
Probably it was too early to integrate it. The limitations found were the following:

- GraphQL requires to be explicit about the fields returned by a query (there is nothing similar to SELECT * in SQL), so in some cases we could have pretty long queries.

- It doesn't seem possible to retrieve nested paginated data in a single query. For instance, issues are generally paginated and some of them can have paginated comments. In this case, two queries should be written to deal with the pagination on issues and comments (as similarly done with REST API calls).

GeorgLink:
Interesting, thank you.

>

sduenas:
I don't like this proposal due to the limitations of GitHub. As @valeriocos said it's not possible to get all the events so that will break the common behaviour backends have.

I understand there's important information there but I'd rather place this on hold until we find something that doesn't break other components.

Thanks for your PR, anyway. We can use it as a reference.

Btw, this discussion should have been done in the issue ticket, not here.

Issue #567, opened 2019-08-29
[meetup] The token is visible in the logs


jjmerchante:
When you try to analyze a group that does not exist, an error is generated and the token that was being used is shown, for example for `events` group (does not exist):

`Error: Not Found for url: https://api.meetup.com/events/events?fields=event_hosts,featured,group_topics,plain_text_description,rsvpable,series&status=cancelled,upcoming,past,proposed,suggested&page=200&order=updated&access_token=xxxxxxxxxxxxxxxxxxxxxxxxxxxxx&scroll=since%3A1970-01-01T00%3A00%3A00.000Z`

It would be better not to show it to prevent anyone reading the logs from obtaining private information.

valeriocos:
Thank you @jjmerchante for reporting this issue, I'll have a look at it ASAP

valeriocos:
The referenced PR should solve the problem. The output for a failing meetup group is now the following:
```
[2019-08-29 14:15:32,233] - Sir Perceval is on his quest.
[2019-08-29 14:15:32,235] - Fetching events of 'events' group from 2019-08-01 00:00:00+00:00 to --
Traceback (most recent call last):
  File "/home/slimbook/Escritorio/sources/perceval/perceval/backend.py", line 471, in run
    for item in items:
  File "/home/slimbook/Escritorio/sources/perceval/perceval/backend.py", line 588, in fetch
    raise e
  File "/home/slimbook/Escritorio/sources/perceval/perceval/backend.py", line 582, in fetch
    for item in items:
  File "/home/slimbook/Escritorio/sources/perceval/perceval/backend.py", line 161, in fetch
    for item in self.fetch_items(category, **kwargs):
  File "/home/slimbook/Escritorio/sources/perceval/perceval/backends/core/meetup.py", line 149, in fetch_items
    for evp in ev_pages:
  File "/home/slimbook/Escritorio/sources/perceval/perceval/backends/core/meetup.py", line 395, in events
    raise error
  File "/home/slimbook/Escritorio/sources/perceval/perceval/backends/core/meetup.py", line 388, in events
    for page in self._fetch(resource, params):
  File "/home/slimbook/Escritorio/sources/perceval/perceval/backends/core/meetup.py", line 466, in _fetch
    r = self.fetch(url, payload=params, headers=headers)
  File "/home/slimbook/Escritorio/sources/perceval/perceval/client.py", line 132, in fetch
[...]

PR #568, opened 2019-08-29
[meetup] Pass oauth2 token via headers

valeriocos:
This PR tackles https://github.com/chaoss/grimoirelab-perceval/issues/567.

The oauth2 token was passed as payload of the requests when issuing petitions to the Meetup API. This was causing a potential security leak for failing petitions, since the URL with the corresponding payload was fully exposed.

Now the token is passed through the headers, which are then sanitized before storing them to the archive.

Since API tokens have been decommissioned in favor of oauth2 token, API token support has been removed from the backend.

Tests have been changed accordingly.
Backend version is now 0.15.0.

### DIFF
--- a/perceval/backends/core/meetup.py
+++ b/perceval/backends/core/meetup.py
@@ -22,7 +22,6 @@
 import json
 import logging
 import requests
-import warnings
 
 from grimoirelab_toolkit.datetime import datetime_to_utc
 from grimoirelab_toolkit.uris import urijoin
@@ -33,8 +32,6 @@ from ...backend import (Backend,
 from ...client import HttpClient, RateLimitHandler
 from ...errors import RepositoryError
 from ...utils import DEFAULT_DATETIME
-
-warnings.simplefilter('always', DeprecationWarning)
 
 
 CATEGORY_EVENT = "event"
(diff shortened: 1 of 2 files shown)


valeriocos:
Please @sduenas @jjmerchante have a look

valeriocos:
Coverage has decreased due to the removal of LOC. If running locally `coverage`, the coverage for meetup is still 100%:
`/home/slimbook/Escritorio/sources/perceval/perceval/backends/core/meetup.py                                                                        179      0   100%`

jjmerchante:
The way it was resolved seems good to me. Using the Token in a header is the recommended way

perceval/backends/core/meetup.py
```diff
@@ -92,17 +91,10 @@ def __init__(self, group, api_token, is_oauth_token=False,
         self.group = group
         self.max_items = max_items
         self.api_token = api_token
-        self.is_oauth_token = is_oauth_token
         self.sleep_for_rate = sleep_for_rate
         self.min_rate_to_sleep = min_rate_to_sleep
         self.sleep_time = sleep_time
 
-        if not self.is_oauth_token:
-            warnings.warn(
```
sduenas: There's also some references to `warning` on the top of the file that you should remove.

perceval/backends/core/meetup.py
```diff
@@ -82,7 +81,7 @@ class Meetup(Backend):
         ['venue']
     ]
 
-    def __init__(self, group, api_token, is_oauth_token=False,
+    def __init__(self, group, api_token,
```
sduenas: Can you update the documentation of this class?  It should say it uses OAuth2 tokens. There's also a reference to `api_key` when the parameter is `api_token` actually.

Issue #569, opened 2019-08-31
[groupsio] update endpoints to fetch archives


valeriocos:
Due to recent breaking changes on groupsio, the endpoints used by the corresponding backend should be updated.

Note that existing Basic Auth tokens will continue to work until the end of September, or when they expire, whichever comes first. More details at: https://beta.groups.io/g/api/message/115

PR #570, opened 2019-09-01
[groupsio] Replace HTTP basic auth with cookies based auth

valeriocos:
This PR addresses #569. It replaces the previous authentication method with a cookies based one. This change is due to recent changes in the Groupsio API that will be effective at the end of September.
More info at: https://beta.groups.io/g/api/message/115

Now the backend accepts email and password (no the api_token anymore), they are used to call the login endpoint, which will set one or more cookies, and enable the other endpoints to work.
All requests performed by the backend are now handled via a session object, which automatically manages cookies.

Backend version is now 0.2.0

The script to list subscriptions has been updated, the new version is available at: https://gist.github.com/valeriocos/2e2231e17fd3052800303bf99bd0c7c4


tests/data/groupsio/login
```diff
@@ -0,0 +1,44 @@
...
+        "time_pref": "standard_time",
+        "timezone": "Europe/Berlin",
+        "topic_sort_dir": "sort_asc",
+        "topics_sort_dir": "sort_desc",
+        "two_factor_enabled": false,
+        "updated": "2019-06-18T02:24:50.203775583-07:00",
+        "user_name": "jsmith",
+        "website": ""
+    }
+}
```
sduenas: Please, all files should contain a black line at the end of the file.

tests/test_groupsio.py
```diff
@@ -252,11 +258,16 @@ def test_fetch(self):
...
-        client = GroupsioClient('beta+api', self.tmp_path, 'aaaaa', verify=False)
+        client = GroupsioClient('beta+api', self.tmp_path, 'jsmith@example.com', 'aaaaa', verify=False)
         success = client.fetch()
 
         # Check requests
         expected = [
+            {
+                'email': ['jsmith@example.com'],
+                'password': ['aaaaa'],
+                'api_key': ['nonce']
```
sduenas: What's `nonce`? According to the post where these breaking changes are described, `api_key` is not needed anymore.

valeriocos: I found it here: https://groups.io/api#login-2, but it can be removed since optional

PR #577, opened 2019-09-20
[test_backend] Provide test for find_backends

valeriocos:
This code increases the test coverage of the backend file by including tests for the `find_backends`
function.


valeriocos:
please @sduenas check it again, thanks

PR #579, opened 2019-10-01
[gitlab] Set extra HTTP retry status from backend

valeriocos:
This code allows to set from the backend/command line which HTTP status should be retried before raising a HTTP error. By default such status are 500 and 502, they complete the ones already defined in the HttpClient class (i.e., 413, 429, 503).

An example of how to set the retry status from command line is provided below:
```
perceval gitlab coldnight ci-test
--category merge_request
-t xxx
--sleep-for-rate
--extra-retry-status 500 502 503
```

Backend version is set to 0.11.0.

Signed-off-by: Valerio Cosentino <valcos@bitergia.com>

### DIFF
--- a/perceval/backends/core/gitlab.py
+++ b/perceval/backends/core/gitlab.py
@@ -57,6 +57,8 @@ PER_PAGE = 100
 DEFAULT_SLEEP_TIME = 1
 MAX_RETRIES = 5
 
+DEFAULT_RETRY_AFTER_STATUS_CODES = [500, 502]
+
 TARGET_ISSUE_FIELDS = ['user_notes_count', 'award_emoji']
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


perceval/backends/core/gitlab.py
```diff
@@ -95,7 +99,7 @@ def __init__(self, owner=None, repository=None, api_token=None,
                  is_oauth_token=False, base_url=None, tag=None, archive=None,
                  sleep_for_rate=False, min_rate_to_sleep=MIN_RATE_LIMIT,
                  max_retries=MAX_RETRIES, sleep_time=DEFAULT_SLEEP_TIME,
-                 blacklist_ids=None):
+                 blacklist_ids=None, extra_retry_after_status=None):
```
jgbarah: If I understand it correctly, None in this parameter means "default value". If that's the case, code would be cleaner if we have `extra_retry_after_status=[500,502]`, which would also be more explicit, or maybe `extra_retry_after_status=`DEFAULT_RETRY_AFTER_STATUS_CODES` as parameter.

valeriocos: Thank you for the review @jgbarah , we generally avoid passing mutable arguments. I would merge it as it is

PR #580, opened 2019-10-07
[github] Make fetching owners information from GitHub API optional

jjmerchante:
This code makes it optional to obtain user information from the GitHub API. This way you can choose to not collect personal information from users.

### DIFF
--- a/perceval/backend.py
+++ b/perceval/backend.py
@@ -259,7 +259,7 @@ class Backend:
 
         for cf in self.CLASSIFIED_FIELDS:
             try:
-                _remove_key_from_nested_dict(item, cf)
+                _remove_key_from_nested_dictlist(item, cf)
             except KeyError:
                 logger.debug("Classified field '%s' not found for item %s; field ignored",
                              '.'.join(cf), item_uuid)
@@ -379,16 +379,19 @@ def _find_value_from_nested_dict(nested_dict, path_to_field):
         return _find_value_from_nested_dict(nested_dict[key], path_to_field[1:])
 
 
-def _remove_key_from_nested_dict(nested_dict, path_to_field):
+def _remove_key_from_nested_dictlist(nested_dictlist, path_to_field):
     if len(path_to_field) == 0:
         return
 
-    key = path_to_field[0]
-
-    if len(path_to_field) == 1:
-        nested_dict.pop(key)
+    if isinstance(nested_dictlist, list):
+        for item in nested_dictlist:
+            _remove_key_from_nested_dictlist(item, path_to_field)
     else:
-        _remove_key_from_nested_dict(nested_dict[key], path_to_field[1:])
+        key = path_to_field[0]
+        if len(path_to_field) == 1:
+            nested_dictlist.pop(key)
+        else:
+            _remove_key_from_nested_dictlist(nested_dictlist[key], path_to_field[1:])
 
 
 class BackendCommandArgumentParser:
--- a/perceval/backends/core/github.py
+++ b/perceval/backends/core/github.py
@@ -86,7 +86,7 @@ class GitHub(Backend):
     :param tag: label used to mark the data
     :param archive: archive to store/retrieve items
     :param sleep_for_rate: sleep until rate limit is reset
-    :param min_rate_to_sleep: minimun rate needed to sleep until
+    :param min_rate_to_sleep: minimum rate needed to sleep until
          it will be reset
     :param max_retries: number of max retries to a data source
         before raising a RetryError exception
@@ -95,9 +95,22 @@ class GitHub(Backend):
     :param sleep_time: time to sleep in case
         of connection problems
     """
-    version = '0.23.1'
+    version = '0.24.0'
 
     CATEGORIES = [CATEGORY_ISSUE, CATEGORY_PULL_REQUEST, CATEGORY_REPO]
+
+    CLASSIFIED_FIELDS = [
+        ['user_data'],
+        ['merged_by_data'],
+        ['assignee_data'],
+        ['assignees_data'],
+        ['requested_reviewers_data'],
+        ['comments_data', 'user_data'],
+        ['comments_data', 'reactions_data', 'user_data'],
+        ['reviews_data', 'user_data'],
+        ['review_comments_data', 'user_data'],
+        ['review_comments_data', 'reactions_data', 'user_data']
+    ]
 
     def __init__(self, owner=None, repository=None,
                  api_token=None, base_url=None,
@@ -124,6 +137,7 @@ class GitHub(Backend):
         self.max_items = max_items
 
         self.client = None
+        self.exclude_user_data = False
         self._users = {}  # internal users cache
 
     def search_fields(self, item):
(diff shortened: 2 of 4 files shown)


GeorgLink:
+1 to GDPR compliance through not capturing user data.

I'm not the engineer, but quickly looking through the changes, I'm wondering whether this code is duplicating what we already have for fetching the information, except without the user data.

valeriocos:
Most of the code you see is about testing. It is needed to make sure that the new feature (you can see the changes in the file perceval/backends/core/github.py) works smoothly.

jjmerchante:
You can make the review @valeriocos.

perceval/backends/core/github.py
```diff
@@ -155,6 +157,9 @@ def fetch(self, category=CATEGORY_ISSUE, from_date=DEFAULT_DATETIME, to_date=DEF
 
         :returns: a generator of issues
         """
+        if self.exclude_user_data:
+            logger.info("Excluding user data. Won't get user data for the login")
```
valeriocos: `Won't get user data for the login` seems too specific, maybe it could be rewritten as:
`Excluding user data. Personal user information won't be collected from the API.`

tests/test_github.py
```diff
@@ -442,6 +524,127 @@ def test_fetch_pulls(self):
         self.assertEqual(len(pull['data']['reviews_data']), 2)
         self.assertEqual(pull['data']['reviews_data'][0]['user_data']['login'], 'zhquan_example')
 
+    @httpretty.activate
+    def test_fetch_pulls_no_user_data(self):
+        """Test whether a list of pull requests is returned"""
```
valeriocos: """Test whether a list of pull requests is returned without user data"""

valeriocos: Maybe the name of the test should be changed since only one issue is returned

tests/test_github.py
```diff
@@ -960,6 +1163,335 @@ def test_fetch_more_pulls(self):
                              'WARNING:perceval.backends.core.github:'
                              'Missing user info for https://api.github.com/repos/zhquan_example/repo/pulls/comments/2')
 
+    @httpretty.activate
+    def test_fetch_more_issues_no_user_data(self):
+        """Test when return two issues"""
```
valeriocos: """Test whether a list of issues is returned without user data"""

tests/test_github.py
```diff
@@ -960,6 +1163,335 @@ def test_fetch_more_pulls(self):
...
+        self.assertEqual(issue['data']['assignees_data'], [])
+        self.assertEqual(len(issue['data']['comments_data']), 1)
+        self.assertEqual(issue['data']['reactions']['total_count'], len(issue['data']['reactions_data']))
+        self.assertIsNone(issue['data']['comments_data'][0]['user_data'])
+        self.assertEqual(issue['data']['comments_data'][0]['reactions']['total_count'],
+                         len(issue['data']['comments_data'][0]['reactions_data']))
+
+    @httpretty.activate
+    def test_fetch_more_pulls_no_user_data(self):
+        """Test when return two pulls"""
```
valeriocos: """Test whether a list of pull requests is returned without user data"""

valeriocos:
Please @sduenas have a look at it when you have time, thanks

tests/test_github.py
```diff
@@ -960,6 +1163,335 @@ def test_fetch_more_pulls(self):
...
+            self.assertEqual(len(pull['data']['review_comments_data'][0]['reactions_data']), 0)
+            self.assertEqual(len(pull['data']['review_comments_data'][1]['reactions_data']), 0)
+            self.assertEqual(len(pull['data']['commits_data']), 1)
+            self.assertEqual(len(pull['data']['reviews_data']), 1)
+            self.assertIsNone(pull['data']['reviews_data'][0]['user_data'])
+            self.assertIsNone(pull['data']['review_comments_data'][0]['user_data'])
+            self.assertIsNone(pull['data']['review_comments_data'][2]['user_data'])
+
+            self.assertEqual(cm.output[0],
+                             "INFO:perceval.backends.core.github:Excluding user data. Won't get user data for the login")
```
valeriocos: a minor typo on the log message, that has changed from the initial version of the PR

perceval/backends/core/github.py
```diff
@@ -85,8 +85,9 @@ class GitHub(Backend):
         from the GitHub public site.
     :param tag: label used to mark the data
     :param archive: archive to store/retrieve items
+    :param exclude_user_data: Don't fetch personal user information from the API
```
sduenas: Please start the sentence with lower case and do not use contractions. (`do not fetch`

perceval/backends/core/github.py
```diff
@@ -963,6 +966,9 @@ def setup_cmd_parser(cls):
         group.add_argument('--min-rate-to-sleep', dest='min_rate_to_sleep',
                            default=MIN_RATE_LIMIT, type=int,
                            help="sleep until reset when the rate limit reaches this value")
+        group.add_argument('--exclude-user-data', dest='exclude_user_data',
+                           action='store_true',
+                           help="Don't fetch personal user information from the API")
```
sduenas: Same as in the previous comment.

valeriocos:
what about having this info reflected in the category names? If the exclude-user flag is true, the category name could include something like '_no_user_data'

sduenas:
> what about having this info reflected in the category names? If the exclude-user flag is true, the category name could include something like '_no_user_data'

That's a possible solution but it has a big problem. If for any reason we keep adding things like this, let's say, something that removes the comments, we will have to add that to `category` field making it unusable.

Not sure if it makes more sense to add it to `filter_classified`.

jjmerchante:
I've updated the pull request with the new implementation and I have modified the tests

perceval/backend.py
```diff
@@ -383,12 +383,15 @@ def _remove_key_from_nested_dict(nested_dict, path_to_field):
...
+    if isinstance(nested_dict, list):
+        for dict_item in nested_dict:
+            _remove_key_from_nested_dict(dict_item, path_to_field)
     else:
-        _remove_key_from_nested_dict(nested_dict[key], path_to_field[1:])
+        key = path_to_field[0]
+        if len(path_to_field) == 1:
+            nested_dict.pop(key)
+        else:
+            _remove_key_from_nested_dict(nested_dict[key], path_to_field[1:])
```
sduenas: The name `nested_dict` for something that might not be a `dict` is misleading. Do you mind to rename it to `nested_dictlist`. The same with the name of the function.

You can also rename `dict_item` to `item`.

tests/test_github.py
```diff
@@ -255,6 +255,82 @@ def test_fetch_issues(self):
                          issue['data']['comments_data'][0]['reactions']['total_count'])
         self.assertEqual(issue['data']['comments_data'][0]['reactions_data'][0]['user_data']['login'], 'zhquan_example')
 
+    @httpretty.activate
```
sduenas: You should also add a more generic test to check `filter_classified` items.

Take as example this PR: https://github.com/chaoss/grimoirelab-perceval/pull/497/files

jjmerchante: I have modified `test_backend.py`. I have included an item in `ClassifiedFieldsBackend` which is a list that contains a dictionary with a field to filter. I have extended the list CLASSIFIED_FIELDS and modified the tests according to these changes. Let me know if that is enough or if I have to add something more sophisticated

jjmerchante:
Ok I missunderstanded, @sduenas check the latest changes, I have included a generic test that checks if the classified fields are filtered

PR #581, opened 2019-10-16
[github] Replace logging with logger call

valeriocos:
This code replaces the logging call with the equivalent of logger when getting the info for a given GitHub user. This change is needed since in ELK, the log message appears as root and not part of perceval.

Backend version is set to 0.23.1

### DIFF
--- a/perceval/backends/core/github.py
+++ b/perceval/backends/core/github.py
@@ -95,7 +95,7 @@ class GitHub(Backend):
     :param sleep_time: time to sleep in case
         of connection problems
     """
-    version = '0.23.0'
+    version = '0.23.1'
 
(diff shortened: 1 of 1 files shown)


perceval/backends/core/github.py
```diff
@@ -751,7 +751,7 @@ def user(self, login):
 
         url_user = urijoin(self.base_url, 'users', login)
 
-        logging.info("Getting info for %s" % (url_user))
+        logger.info("Getting info for %s" % url_user)
```
sduenas: Can you change the level of this entry too? I think this a debug message, not an info one.

valeriocos:
done! please check again @sduenas, thanks!

Issue #591, opened 2019-11-14
Perceval jenkins datasource fails (internal python error)


lukaszgryglicki:
`p2o.py --enrich --index (redacted) --index-enrich (redacted) -e (redacted) --db-host mariadb.mariadb --db-sortinghat sortinghat --db-user sortinghat --db-password (redacted) jenkins https://jenkins.zowe.org --no-archive --category build`:
```
2019-11-14 10:08:32,615 Looking for projects at url 'https://jenkins.zowe.org'
2019-11-14 10:08:33,352 Error feeding raw from jenkins (https://jenkins.zowe.org): 'builds'
Traceback (most recent call last):
  File "/repos/grimoirelab-elk/grimoire_elk/elk.py", line 228, in feed_backend
    ocean_backend.feed(**params)
  File "/repos/grimoirelab-elk/grimoire_elk/raw/elastic.py", line 230, in feed
    self.feed_items(items)
  File "/repos/grimoirelab-elk/grimoire_elk/raw/elastic.py", line 246, in feed_items
    for item in items:
  File "/repos/grimoirelab-perceval/perceval/backend.py", line 215, in fetch
    for item in self.fetch_items(category, **kwargs):
  File "/repos/grimoirelab-perceval/perceval/backends/core/jenkins.py", line 148, in fetch_items
    builds = builds['builds']
KeyError: 'builds'
2019-11-14 10:08:33,354 [jenkins] Done collection for https://jenkins.zowe.org
2019-11-14 10:08:33,355 Backend feed completed
2019-11-14 10:08:33,441 Total items enriched 0 
2019-11-14 10:08:33,442 [jenkins] Done enrichment for https://jenkins.zowe.org
2019-11-14 10:08:33,442 Enrich backend completed
2019-11-14 10:08:33,442 Finished in 0.01 min
```

PR #595, opened 2019-11-18
[jenkins] Handle missing job builds

valeriocos:
This PR addresse #591, thus it handles jobs not having builds associated.

Backend version is now 0.14.2


sduenas:
Not sure if this is the best solution, specially after the findings described in #596 . The problem is not that there aren't builds, it's because the type of job is different so builds are accessed in a different way.

valeriocos:
Replaced by https://github.com/chaoss/grimoirelab-perceval/pull/597

PR #597, opened 2019-11-20
[jenkins] Handle workflow multibranch job builds

valeriocos:
This code allows to process workflow multibranches job, which include nested jobs and their corresponding builds.

Since the API doesn't offer an endpoint to retrieve workflow jobs, the backend leverages on the class type of the job to process it accordingly. Thus, if the class type is `WorkflowMultiBranchProject`, the nested jobs are retrieved one by one and their builds obtained. Otherwise the logic to process the job is left as it was.

Backend version is now 0.15.0


valeriocos:
please check it again

PR #598, opened 2019-11-23
[discourse] Pass credentials via headers

valeriocos:
This code changes the way the credentials are passed due to [a recent change](https://meta.discourse.org/t/discourse-api-documentation/22706) in the Discourse API. Now, the credentials, defined as `api_token` and `api_username` are defined in the headers (and not as request params).

Tests have been updated accordingly.

Backend version is now 0.12.0.

### DIFF
--- a/perceval/backends/core/discourse.py
+++ b/perceval/backends/core/discourse.py
@@ -19,6 +19,7 @@
 #    Santiago Dueñas <sduenas@bitergia.com>
 #    J. Manrique López de la Fuente <jsmanrique@bitergia.com>
 #    Alvaro del Castillo San Felix <acs@bitergia.com>
+#    Valerio Cosentino <valcos@bitergia.com>
 #
 
 import json
@@ -31,6 +32,7 @@ from ...backend import (Backend,
                         BackendCommand,
                         BackendCommandArgumentParser)
 from ...client import HttpClient
+from ...errors import BackendError, HttpClientError
 from ...utils import DEFAULT_DATETIME
 
 
@@ -50,6 +52,7 @@ class Discourse(Backend):
     will be set as the origin of the data.
 
     :param url: Discourse URL
+    :param api_username: Discourse API username
     :param api_token: Discourse API access token
     :param tag: label used to mark the data
     :param archive: archive to store/retrieve items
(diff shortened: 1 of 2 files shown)


perceval/backends/core/discourse.py
```diff
@@ -58,20 +61,25 @@ class Discourse(Backend):
...
                  max_retries=MAX_RETRIES, sleep_time=DEFAULT_SLEEP_TIME):
         origin = url
 
         super().__init__(origin, tag=tag, archive=archive)
         self.url = url
+        self.api_username = api_username
         self.api_token = api_token
+
+        if (self.api_username and not self.api_token) or (self.api_token and not self.api_username):
+            raise BackendError(cause="Api token and username must be defined together")
```
sduenas: Can you add this also to the client? It would be useful to have an exception in case someone wants to use the client only.

perceval/backends/core/discourse.py
```diff
@@ -299,27 +308,31 @@ class DiscourseClient(HttpClient):
...
+
     # Params
-    PKEY = 'api_key'
     PPAGE = 'page'
 
     # Data type
     TJSON = '.json'
 
-    def __init__(self, base_url, api_key=None, sleep_time=DEFAULT_SLEEP_TIME, max_retries=MAX_RETRIES,
+    def __init__(self, base_url, api_username=None, api_key=None, sleep_time=DEFAULT_SLEEP_TIME, max_retries=MAX_RETRIES,
```
sduenas: Would you mind to make this line shorter? Before the change was also lager than it should, in my opinion.

perceval/backends/core/discourse.py
valeriocos: should this change be applied also to GitLab (https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/backends/core/gitlab.py#L106)?

valeriocos:
@sduenas please check the code again. Note that the changed asked at https://github.com/chaoss/grimoirelab-perceval/pull/598#pullrequestreview-322207512 has been applied also to the GitLab client for consistency reasons.

If needed, I can prepare a different PR for the change on GitLab.

valeriocos:
Running coveralls locally doesn't report that the test coverage for discourse and gitlab has decreased. They are 100% and 99%   (missing lines 407-412)

perceval/backends/core/discourse.py
sduenas: Yes!

sduenas:
> @sduenas please check the code again. Note that the changed asked at [#598 (review)](https://github.com/chaoss/grimoirelab-perceval/pull/598#pullrequestreview-322207512) has been applied also to the GitLab client for consistency reasons.
> 
> If needed, I can prepare a different PR for the change on GitLab.

It would be better to have it in different PRs. The rest of the code looks good.

valeriocos:
splitted: https://github.com/chaoss/grimoirelab-perceval/pull/599, please have a look when you have time.

Issue #607, opened 2020-01-16
[groups.io] support rate limiting and start_msg_num parameter for /downloadarchives API


lukaszgryglicki:
Feature request:
groups.io is now rate limiting calls to `/downloadarchives` API

- Please add support for `--sleep-for-rate`, something similar to GitHub perceval backend.

Additionally, groups.io now allows "resuming" data processing. This is the informations we gotfrom groups.io developers:
*I've added a start_msg_num parameter to the `/downloadarchives` API. Also, in the message headers of the archives downloaded, I've added a `X-Groupsio-MsgNum` header with the message number. Download all the archives initially, then scan for the highest MsgNum in what you've downloaded. Use that number for start_msg_num next time you call `/downloadarchives` and you'll only get newer messages. Repeat the process each time.*

- Please add support for groups.io resuming.

lukaszgryglicki:
OK, starting work on this.

valeriocos:
@lukaszgryglicki are you with this issue? if needed I can take it over. Let me know, thanks

lukaszgryglicki:
I need to work on something else at least one day, so you can take it.

valeriocos:
Hi @lukaszgryglicki, 

I'm not sure the sleep for rate can be useful in this case, since the remaining time before issuing another request to the API isn't returned in the response (or at least, I didn't find it). In token-based backends (e.g., github and gitlab), that time is included in the response, thus it's pretty easy to stop and resume the fetching process. It would be great if that info would be included in the API response.
In the meanwhile, you can pull the data from groups.io every 2-3 days (we are following this approach ATM)

I'm not sure about the usefulness of `start_msg_num`, as a final user it would make more sense to get the messages after a given date. In this sense, I dropped a message here: https://beta.groups.io/g/api/topic/feature_request_add/69933215?p=,,,20,0,0,0::recentpostdate%2Fsticky,,,20,2,0,69933215

Best,
Valerio

valeriocos:
@lukaszgryglicki the discussion on the groupsio link above is ongoing, feel free to jump on it if you see it useful.

lukaszgryglicki:
I've seen the discussion, is there any problem to implement resuming using last message ID instead of date? 
And "sleep for rate" needs some work on groups.io side, right? Like returning the number of seconds before rate limit expires, right?

valeriocos:
> I've seen the discussion, is there any problem to implement resuming using last message ID instead of date?

[...]

PR #609, opened 2020-02-03
[groupsio] Fetch messages after a given date

valeriocos:
This PR addresses #607. It allows to collect a list of messages sent after a given date (included) from the groupsio API. The backend `from_date` parameter is propagated to the client, which sends it to the API endpoint to download archives. This change is made possible after a recent change in the upstream API, and allows to improve the performance of the fetching operation.

Backend version is now 0.3.2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,14 +238,10 @@ $ perceval googlehits "bitergia grimoirelab"
 
 ### Groups.io
 ```
-$ perceval groupsio 'updates' -t 'abcdefghijk' --from-date '2016-01-01'
+$ perceval groupsio 'updates' -e '<me@example.com>' -p 'my-password' --from-date '2016-01-01'
 ```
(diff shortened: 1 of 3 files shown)


valeriocos:
Documentation updated @sduenas

lukaszgryglicki:
Will `p2o.py` make use of this automatically? I mean run p2o.py once on some group (so it stores raw/enriched data in the ES index) and then run again after for example 2 days - will it only fetch data for the last 2 days using `date-from` parameter?

valeriocos:
yes @lukaszgryglicki , it should be automatically since p2o relies on [feed_backend](https://github.com/chaoss/grimoirelab-elk/blob/master/grimoire_elk/elk.py#L105) and [enrich_backend](https://github.com/chaoss/grimoirelab-elk/blob/master/grimoire_elk/elk.py#L473), which are also used by mordred.

lukaszgryglicki:
That's exactly what I was expecting. Thanks!

Issue #610, opened 2020-02-04
[perceval] Evaluate support for --no-verify param


valeriocos:
In some cases, a certificate is needed to execute Perceval against a remote data source (e.g, mattermost, gitlab). The approach of non verifying the certificates has been already applied to different backends such as [groupsio](https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/backends/core/groupsio.py#L64), [pipermail](https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/backends/core/pipermail.py#L62) and [jira](https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/backends/core/jira.py#L97). However, it would useful to have it for all backends (as similarly done to disable archives https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/backend.py#L530)

The goal of this issue is to evaluate and implement a generic approach to not verifying certificates.

valeriocos:
The param has been merged to the perceval code base. Before closing this issue, we need to update the perceval-based components (perceval for puppet, mozilla, etc.)

valeriocos:
Fixed wiht #612. The change has been applied also to the other percevals:
- https://github.com/chaoss/grimoirelab-perceval-mozilla
- https://github.com/chaoss/grimoirelab-perceval-opnfv
- https://github.com/chaoss/grimoirelab-perceval-puppet
- https://github.com/Bitergia/grimoirelab-perceval-finos

PR #612, opened 2020-02-06
[perceva] Allow to disable SSL verification

valeriocos:
This PR addresses https://github.com/chaoss/grimoirelab-perceval/issues/610, thus it proposes to include the parameter `verify` in the generic Perceval backend to disable SSL verification.

The BackendCommandArgumentParser has been modified accordingly to support this feature. Thus, the user can disable the SSL verification by typing `--no-verify` in the command line together with the other Perceval parameters.

The HttpClient has been updated. Some backends have been updated to showcase this feature and simplifying the review process. In the next days, the other backends will be extended.

### DIFF
--- a/perceval/backend.py
+++ b/perceval/backend.py
@@ -117,23 +117,25 @@ class Backend:
     :param origin: identifier of the repository
     :param tag: tag items using this label
     :param archive: archive to store/retrieve data
+    :param ssl_verify: enable/disable SSL verification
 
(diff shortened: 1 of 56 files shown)


valeriocos:
in Git you can disable the ssl verification with `git -c http.sslVerify=false clone https://e..`. What do you think about `no-ssl-verify` or `disable-ssl-verify`? The corresponding variable could be `ssl-verify`. WDYT?

Issue #543, opened 2019-06-19
[Feature request] Add support for Rocket.Chat


lukaszgryglicki:
Hi, any chance to add backend support for Rocket.Chat, for example, to be able to get data from https://chat.hyperledger.org.
Thanks, best regards.

sduenas:
I don't think this is under our plan yet but PRs are always welcome.

lukaszgryglicki:
Ok, so I'll ask my team if they want me to implement this.

PR #629, opened 2020-03-18
Add support for rocket chat 

Syn3rman:
Hey @valeriocos,
As per the discussions in [chaoss/grimoirelab#257](https://github.com/chaoss/grimoirelab/issues/257) and related issue #543, this commit aims at adding support to fetch messages from a RocketChat server.


valeriocos:
Thanks @Syn3rman , please don't hesitate to write when you need feedback on the implementation. Is there a public instance of rocketchat I can use to test the PR?

Syn3rman:
Hey @valeriocos,
There's https://open.rocket.chat/channel/general for which you will need to create a personal access token under the My Account > Personal Access Tokens Section (after creating an account).

If you choose to use this, you can use the following room-id to pass it to the CLI, which is for the sandbox room: Drjw54ftqGa4antMW. (Since I was not able to get the id of the room from the browser, I had to make an api call to get it, because the api needs the room-id and not the room name - at least from what I could gather). I could try adding a CLI flag for getting all the rooms available for a given server, if needed.

Also, the timestamp for the messages was in the format 2020-03-18T05:50:51.181Z and I couldn't figure out how to convert it to a UNIX ts, so I have just sliced the first 10 characters (yyyy-mm-dd) and used them to get a timestamp. 

And the last question I have in mind is - is there a need to change the format in which the CLI takes the arguments from the user? The current way works but I am not sure if it conforms to standards that might exist.

Syn3rman:
Hey @valeriocos,

I think that the pr is ready for review, could you have a look at it?

The only problem I am running into is setting the from_date because I can't figure out the format required (and the way to convert a datetime object to it).

valeriocos:
Thank you @Syn3rman , before having a look at the PR, please check if the function [str_to_datetime](https://github.com/chaoss/grimoirelab-toolkit/blob/master/grimoirelab_toolkit/datetime.py#L97) is useful for your case. See example at https://github.com/chaoss/grimoirelab-perceval/blob/d273525c6609ab9bcb8f06a2cb0d78b4c70fa11c/perceval/backends/core/mbox.py#L159.

Please make also sure that the docstrings of the different methods follow the same format of the other backends.

Thanks

Syn3rman:
Hey @valeriocos,

I spent quite some time figuring how to get the from_date param to work before realizing I should probably try sending a get request to the API. I did so and was not able to get the messages from a given timestamp to now, and after googling around a bit I came across the following reported issue - https://github.com/RocketChat/Rocket.Chat/issues/16664.

This makes me believe that there is a problem with the API itself, and the community does not seem to be too active, so it can take a while to get it fixed. 

So, should I drop the --from-date flag entirely or complete the code for it too, since there anyways won't be a way to test it?
(I'll fix the docstrings and update the pr as soon as I get a go ahead on this.)

animeshk08:
Hi @Syn3rman. I had a look at rocket chats API documentation. I believe you are using the following endpoint: https://rocket.chat/docs/developer-guides/rest-api/livechat/message/index.html#load-livechat-messages-history

Do you think this endpoint would me more suitable for retrieving the messages:
https://rocket.chat/docs/developer-guides/rest-api/channels/messages/

Here you can just provide the room name as well. There is no need for room id. The `access token` and `userid` are added to the header. Also, a from-date parameter can be mapped by creating a query as shown here: https://rocket.chat/docs/developer-guides/rest-api/query-and-fields-info/

Note: This endpoint can be used in case of private channels: https://rocket.chat/docs/developer-guides/rest-api/groups/messages/

WDYT?

perceval/backends/core/rocketchat.py
```diff
@@ -0,0 +1,338 @@
...
+
+        resource = self.ROOM_HISTORY
+
+        params = {
+            self.ROOM_ID: roomId,
+            self.PCOUNT: self.max_items,
+            self.POLDEST: "oldest",
+        }
+
+        response = self._fetch(resource, params)
```
valeriocos: How many old messages can be fetched from a channel? In Slack, the amount of messages to be retrieved is limited (that API doesn't store all historical data), do we have the same behavior in RocketChat?

perceval/backends/core/rocketchat.py
```diff
@@ -0,0 +1,338 @@
...
+    """RocketChat API client.
+    Client for fetching information from the RocketChat server
+    using its REST API.
+    :param api_token: key needed to use the API
+    :param max_items: maximum number of items per request
+    :param archive: an archive to store/read fetched data
+    :param from_archive: it tells whether to write/read the archive
+    :param ssl_verify: enable/disable SSL verification
+    """
+    ROOM_HISTORY = 'channels.history'
```
valeriocos: Following the pointer of @animeshk08 , I checked the two endpoints:
- https://rocket.chat/docs/developer-guides/rest-api/channels/history/
- https://rocket.chat/docs/developer-guides/rest-api/channels/messages/

From the description of the two endpoints I don't understand the difference. I see that the latter was added in a later release and has some additional filtering. Beyond that the data returned seems the same.

Is there a specific reason for selecting the `channels.history`?

perceval/backends/core/rocketchat.py
```diff
@@ -0,0 +1,338 @@
...
+        logger.debug("RocketChat client requests: %s params: %s",
+                     resource, str(params))
+
+        r = self.fetch(url, payload=params, headers=headers)
+
+        # Check for possible API errors
+        result = r.json()
+
+        if r.status_code != 200:
+            raise RocketChatClientError(error=result['error'])
```
valeriocos: In which cases is the exception triggered?

perceval/backends/core/rocketchat.py
```diff
@@ -0,0 +1,338 @@
...
+                    roomId, str(from_date))
+
+        raw_info = self.client.room_info(self.roomId)
+        room_info = self.parse_room_info(raw_info)
+
+        oldest = datetime_to_utc(from_date).timestamp()
+
+        raw_history = self.client.history(roomId,
+                                          oldest=oldest)
+        messages = self.parse_history(raw_history)
```
valeriocos: As similarly done for [slack](https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/backends/core/slack.py#L128), it would probably make sense to fetch channels members (https://rocket.chat/docs/developer-guides/rest-api/channels/members/), WDYT?

perceval/backends/core/rocketchat.py
```diff
@@ -0,0 +1,338 @@
...
+        user = {
+            "username": username,
+            "name": name,
+        }
+
+        self._users[user_id] = user
+        return user
+
+
+class RocketChatClient(HttpClient):
```
valeriocos: This class should also use the `RateLimitHandler` (example at https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/backends/core/github.py#L571). This class is used to handle rate limit information, which are returned in the headers of the response (see figure below):

captura_269

perceval/backends/core/rocketchat.py
```diff
@@ -0,0 +1,338 @@
...
+        self.roomId = roomId
+        self.api_token = api_token
+        self.max_items = max_items
+        self.client = None
+        self.user_id = user_id
+        self._users = {}
+
+    def fetch(self, category=CATEGORY_MESSAGE, from_date=DEFAULT_DATETIME):
+        """Fetch the messages from the room.
+                This method fetches the messages stored on the room that were
```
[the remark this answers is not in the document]
Syn3rman: I had updated the docstrings but had forgotten to push them, so sorry!

perceval/backends/core/rocketchat.py
Syn3rman: The slack backend seems to only fetch the number of users in the channel, which is already provided in the usersCount field in the channels_info. If there is a need to get all the users in a channel, it could be done though.

perceval/backends/core/rocketchat.py
```diff
@@ -0,0 +1,338 @@
...
+                message['user_data'] = self.__get_or_fetch_user(user_id, username, name)
+
+            message["room_info"] = room_info
+            yield message
+            nmsgs += 1
+
+        logger.info("Fetch process completed: %s message fetched", nmsgs)
+
+    @staticmethod
+    def parse_room_info(raw_room_info):
```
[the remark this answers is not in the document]
Syn3rman: Done!

perceval/backends/core/rocketchat.py
```diff
@@ -0,0 +1,338 @@
...
+    @staticmethod
+    def metadata_updated_on(item):
+        """Extracts the update time from a RocketChat item.
+        The timestamp is extracted from 'ts' field,
+        and sliced to get the year, month and date.
+        It is then converted into a UNIX timestamp.
+        :param item: item generated by the backend
+        :returns: extracted timestamp
+        """
+        ts = item['ts'][:10]
```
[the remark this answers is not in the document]
Syn3rman: I ended up using 
```
ts = str_to_datetime(item['ts'])
ts = time.mktime(ts.timetuple())
```
Is it alright?

perceval/backends/core/rocketchat.py
Syn3rman: What exactly do I need to add in it? 
GitHub supports multiple tokens but I have not yet added the support for multiple API tokens.

perceval/backends/core/rocketchat.py
Syn3rman: I had originally missed the query parameter while going through the documentation, which is why I stuck with channels.history. However, I spent a substantial amount of time the day yesterday trying to get date queries to work using either of the endpoints, and I came to the conclusion that date queries just don't work (at least I wasn't able to get them to work even spending a lot of time on them). 

For channels.history, I tried adding the oldest parameter, in various formats, including the one mentioned at [https://github.com/RocketChat/Rocket.Chat/issues/10871](https://github.com/RocketChat/Rocket.Chat/issues/10871), but the api sent the same response as the one without it specified. I then thought of moving to channels.messages and went over the documentation for the query parameter to try and make channels.messages work, but I ran into a `invalid-query` error for queries that supposedly worked for other people, and was stuck on it before realizing my mistake. However, even after getting it to work and spending quite some time on getting it to work, I still kept getting the following response:

    {
        "messages": [],
        "success": true
    }

By this point, I was convinced that date queries don't work, but in a last effort, tried adding the `latest` parameter to the channels.history endpoint, and this time got the same result as above.

This is why I am pretty convinced that date queries don't work with the rocket chat api.

perceval/backends/core/rocketchat.py
Syn3rman: From what I read over [here](https://github.com/RocketChat/Rocket.Chat/pull/11236#issuecomment-399990958), I believe that the number of messages stored depends on the admin, and can vary for each server. Since the `oldest` parameter did not seem to work, I tried running the backend on [https://open.rocket.chat/](https://open.rocket.chat/) with `--max-items` set to 100000, and it returned 42089 messages, the last one dating to `2019-03-12` (around a year)

perceval/backends/core/rocketchat.py
Syn3rman: One example is `400 Client Error: Bad Request for url`, I'm not sure if others exist. Also, there are cases when the api returns an html page instead of a response, one case being if the server url is incorrect. There are also cases when the response status code is 200, but the api encountered an error and sent it inside the response body, like:

    {'success': False,
     'error': 'Invalid query parameter provided: "{"_updatedAt" : { "$gt" : { "date : 1526847048093}}}" [error-invalid-query]',
     'errorType': 'error-invalid-query',
     'details': {'helperMethod': 'parseJsonQuery'}}

I've added error handling for these too.

perceval/backends/core/rocketchat.py
```diff
@@ -0,0 +1,338 @@
...
+        # Required arguments
+        parser.parser.add_argument('url',
+                                   help="URL of the RocketChat server")
+
+        # Rocket Chat options
+        group = parser.parser.add_argument_group('Rocket.chat arguments')
+        group.add_argument('--max-items', dest='max_items',
+                           type=int, default=MAX_ITEMS,
+                           help="Maximum number of items requested on the same query")
+        group.add_argument('--room-id', dest='roomId',
```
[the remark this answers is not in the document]
Syn3rman: Moved it to required! Also, there was only one argument remaining in the argument group `rocketchat arguments`, and I decided to keep the group instead of removing it even though it has only one argument for the sake of consistency and in the case that other arguments need to be added in the future.

perceval/backends/core/rocketchat.py
valeriocos: Interesting, thanks @Syn3rman for the info. 

Why the param `oldest` doesn't seem to work? I understand that if you run the backend on https://open.rocket.chat/ with from-date `2019-01-01` and max-items `100000`, the last message should have date `2019-01-01`. Am I missing something?

perceval/backends/core/rocketchat.py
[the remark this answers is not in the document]
valeriocos: Can we have a solution without the need of importing the package `time`?

perceval/backends/core/rocketchat.py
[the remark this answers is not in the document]
valeriocos: no worries

perceval/backends/core/rocketchat.py
valeriocos: I made some tests and I see that `max-items/count` takes over the `oldest/from-date`. The reason should be that the client doesn't provide a mechanism to iterate over the paginated results provided by the RocketChat API (https://rocket.chat/docs/developer-guides/rest-api/offset-and-count-and-sort-info/).

An approach could be to mimic what the slack backend does (https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/backends/core/slack.py#L135). Thus, the backend keeps asking for data until there is no message returned by the API. WDYT?

perceval/backends/core/rocketchat.py
valeriocos: Ok, thank you @Syn3rman for reporting this useful information. Once the PR is ready for review, I'll have a look and evaluate possible workarounds (if exist).

perceval/backends/core/rocketchat.py
valeriocos: We can leave the inclusion of all users in a channel as future enhancement :)

perceval/backends/core/rocketchat.py
Syn3rman: > Interesting, thanks @Syn3rman for the info.
> 
> Why the param `oldest` doesn't seem to work? I understand that if you run the backend on https://open.rocket.chat/ with from-date `2019-01-01` and max-items `100000`, the last message should have date `2019-01-01`. Am I missing something?

The api seems to ignore the parameter and return the result that it would if the parameter wasn't passed. I had tried substituting the `max-items` as 100000 just to check how far back the messages were stored.

> An approach could be to mimic what the slack backend does (https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/backends/core/slack.py#L135). Thus, the backend keeps asking for data until there is no message returned by the API. WDYT?

I went through the link stated above and I think that even though the API does not support the from date parameter, we could just fetch results incrementally and stop when the date is matched.
However, I think that this might be a bit slow and exhaust the rate limit, if it is very stringent. 
I tried it the below code as a standalone python script, and it appears to work:
```
fetching = True
offset = 1

while fetching:
    params = {
        "roomName": "general",
        "count": count,
        "offset": offset*count
    }
    r = requests.get(url, headers=headers,params=params)
    res = r.json()
    msg = res["messages"]
    for val in msg:
        msg_date = str_to_datetime(val["_updatedAt"])
        print(msg_date, from_date, msg_date < from_date)
        if msg_date < from_date:
            fetching = False
    offset+=1
```

valeriocos: good news, thanks @Syn3rman. A similar approach has been implemented also in other backends, for instance:
- https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/backends/core/meetup.py#L154
- https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/backends/core/mattermost.py#L146

Please consider that Perceval retrieves data from the oldest to the newest, take this into account when getting the data from the API (ref: https://rocket.chat/docs/developer-guides/rest-api/offset-and-count-and-sort-info/#query-parameter-information). Thanks

Syn3rman: Okay!
I finally figured out how to do it, I will complete this and the rate limit part in a few hours and push it :)

valeriocos:
Please @Syn3rman ping me when the PR is ready for review, thanks!

Syn3rman:
Hey @valeriocos,
I think it's good to go, so could you have a look at it when you're free?

valeriocos:
Thank you @Syn3rman I'll have a look at it tomorrow

perceval/backends/core/rocketchat.py
```diff
@@ -0,0 +1,350 @@
...
+    :param max_items: maximum number of message requested on the same query
+    :param tag: label used to mark the data
+    :param archive: archive to store/retrieve items
+    :param ssl_verify: enable/disable SSL verification
+    """
+    version = '0.1.0'
+
+    CATEGORIES = [CATEGORY_MESSAGE]
+
+    def __init__(self, url, user_id, channel_name, api_token, max_items=MAX_ITEMS,
```
valeriocos: There are a bunch of params that aren't present, please add them:
```
     def __init__(self, url, user_id, channel_name, api_token, max_items=MAX_ITEMS,
+                 sleep_for_rate=False, min_rate_to_sleep=MIN_RATE_LIMIT,
+                 max_retries=MAX_RETRIES, sleep_time=SLEEP_TIME,
                  tag=None, archive=None, ssl_verify=True):
```

perceval/backends/core/rocketchat.py
```diff
@@ -0,0 +1,350 @@
...
+                 tag=None, archive=None, ssl_verify=True):
+
+        self.url = url
+
+        super().__init__(url, tag=tag, archive=archive, ssl_verify=ssl_verify)
+        self.channel_name = channel_name
+        self.api_token = api_token
+        self.max_items = max_items
```
valeriocos: Add a blank line, and later please add:
```
+
+        self.sleep_for_rate = sleep_for_rate
+        self.min_rate_to_sleep = min_rate_to_sleep
+        self.max_retries = max_retries
+        self.sleep_time = sleep_time
+
```

perceval/backends/core/rocketchat.py
```diff
@@ -0,0 +1,350 @@
...
+        :returns: extracted timestamp
+        """
+        ts = str_to_datetime(item['_updatedAt']).timestamp()
+        return ts
+
+    def _init_client(self, from_archive=False):
+        """Init client"""
+
+        return RocketChatClient(self.url, self.user_id, self.api_token, self.max_items, self.archive,
+                                from_archive, self.ssl_verify, channel_name=self.channel_name)
```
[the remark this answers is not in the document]
valeriocos: Please pass the backend params to the client:
```
+        return RocketChatClient(self.url, self.user_id, self.api_token,
+                                self.sleep_for_rate, self.min_rate_to_sleep,
+                                self.sleep_time, self.max_retries, self.max_items, self.archive,
+                                from_archive, self.ssl_verify)
```

perceval/backends/core/rocketchat.py
```diff
@@ -0,0 +1,350 @@
...
+    """
+    CHANNEL_MESSAGES = 'channels.messages'
+    CHANNEL_NAME = 'roomName'
+    CHANNEL_INFO = 'channels.info'
+    AUTH_TOKEN = 'X-Auth-Token'
+    USER_ID = 'X-User-Id'
+    PCOUNT = "count"
+    POLDEST = "oldest"
+
+    def __init__(self, url, user_id, api_token, max_items,
```
valeriocos: Add the missing params

perceval/backends/core/rocketchat.py
```diff
@@ -0,0 +1,350 @@
...
+                 sleep_for_rate=False, min_rate_to_sleep=MIN_RATE_LIMIT, sleep_time=SLEEP_TIME,
+                 archive=None, from_archive=False,
+                 ssl_verify=True, channel_name=None):
+        self.api_token = api_token
+        self.max_items = max_items
+        self.channel_name = channel_name
+        self.user_id = user_id
+
+        super().__init__(url, sleep_time=sleep_time, archive=archive, from_archive=from_archive, ssl_verify=ssl_verify)
```
valeriocos: Please pass the max_retries to the method.

```
-        super().__init__(url, sleep_time=sleep_time, archive=archive, from_archive=from_archive, ssl_verify=ssl_verify)
+        super().__init__(url, sleep_time=sleep_time, max_retries=max_retries,
+                         archive=archive, from_archive=from_archive, ssl_verify=ssl_verify)
```

perceval/backends/core/rocketchat.py
```diff
@@ -0,0 +1,350 @@
...
+
+        r = self.fetch(url, payload=params, headers=headers)
+
+        # Check for possible API errors
+        try:
+            response = r.json()
+        except ValueError:
+            raise RocketChatClientError(error="Check command and try again.")
+
+        if r.status_code != 200:
```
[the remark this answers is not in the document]
Syn3rman: Hey,
I added that because I'd run into a scenario in which the API returned an HTML document, causing r.json() to fail.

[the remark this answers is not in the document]
valeriocos: Interesting, do you have the example? Or do you remember the content of the HTML?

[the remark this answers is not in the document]
Syn3rman: I ran into it while trying out and tuning an example given in the documentation. I'm not sure why or how I ran into it, but the content [looked like](https://pastebin.com/j8LYDp83)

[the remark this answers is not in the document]
valeriocos: I'll try to replicate this issue. If we are not able to replicate it, it is better to remove that part of code. WDYT?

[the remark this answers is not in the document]
Syn3rman: I too wasn't able to replicate it, so I think it will be okay to remove it.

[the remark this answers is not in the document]
valeriocos: ok, perfect!

can you replace the if-elif statement with:
```
        response = r.json()
        if not response["success"]:
            raise RocketChatClientError(error=response['error'])
```
?

perceval/backends/core/rocketchat.py
```diff
@@ -0,0 +1,398 @@
...
+        self.user_id = user_id
+
+        super().__init__(url, sleep_time=sleep_time, max_retries=max_retries,
+                         archive=archive, from_archive=from_archive, ssl_verify=ssl_verify)
+        super().setup_rate_limit_handler(sleep_for_rate=sleep_for_rate, min_rate_to_sleep=min_rate_to_sleep)
+
+    def calculate_time_to_reset(self):
+        """Number of seconds to wait. They are contained in the rate limit reset header"""
+
+        time_to_reset = self.rate_limit_reset_ts - (datetime_utcnow().replace(microsecond=0).timestamp() + 1) * 1000
```
valeriocos: can you move the second expression to a variable?
```
current_epoch = (datetime_utcnow().replace(microsecond=0).timestamp() + 1) * 1000
```

Syn3rman:
Hey @valeriocos,
Would it be possible to connect on IRC for some time if you're free? 
This was my first time writing tests and I was running into some problems so I decided to work on the remaining microtasks and my proposal instead. I've got some questions regarding the proposal too so could I ask those too?

valeriocos:
Hi @Syn3rman ,

Sure, what about at 14h Madrid time (in around 2h20)?

Syn3rman:
Sure, thanks!

valeriocos:
Thank you for the quick reply @Syn3rman ! We can talk on IRC even now if you are available (I'm already connected).

inishchith:
@valeriocos Sorry, I missed out on the mentions 🙈 

> @Syn3rman said: .. And I'd love to hear any additional comments @inishchith might have!

Sure, I'll have a look in the next few hours.

Thanks!

valeriocos:
No worries @inishchith!

valeriocos:
In reference to the comment https://github.com/chaoss/grimoirelab-perceval/pull/629#issuecomment-605418865, @inishchith please consider to share some of your experience to write tests (you did a really good job in writing the tests for Graal, which are similar to the ones for Perceval), thanks!

inishchith:
@Syn3rman I had a closer look at your work. 
It looks to me in great shape as you've addressed most of the questions and suggestions by @valeriocos. Please do have a look at the remaining ones.

Great Job ✨ 


> @valeriocos said: In reference to the comment #629 (comment),...

Sure. 
I would have started off by going through the tests written for [GitHub](https://github.com/chaoss/grimoirelab-perceval/blob/master/tests/test_github.py), being extensive it'd add more clarity. Please do let me know once you go through and what parts you're finding difficult to understand. @valeriocos and I would be more than happy to help.

Also, would like to add a thought. 
As the GSoC student application deadline is right around the corner, I'd prefer you to put tests on hold (as we're not in a hurry :P) and focus on writing a great proposal and completing the microtasks if you're planning to apply. 

LMK what you think :) 

/cc @valeriocos 

Thanks!

Syn3rman:
Thanks for the reply @inishchith!

I am almost ready with my draft proposal, I'll share it with you'll for feedback once I'm done with it!

Syn3rman:
> I am almost ready with my draft proposal, I'll share it with you'll for feedback once I'm done with it!

Could you'll have a look at it when you get the time? @inishchith @valeriocos 

Proposal [here](https://docs.google.com/document/d/1H1nunFjeVjRQhtEF-VmQLaoRcb53G59eN5hIlYnTO_k/edit?usp=sharing)

valeriocos:
Hi @Syn3rman , how is it going? Do you need help to finalize the PR?

Syn3rman:
Hey @valeriocos,
Sorry for the late reply. I was running into a few problems while writing the tests but I've added an initial version for the tests. I also managed to mess up the fork while rebasing the commits :/ so I've reset the fork to the upstream version. Is it okay if I commit the files for rocketchat into a single commit in the updated fork?

valeriocos:
Hi @Syn3rman , no worries!

>  Is it okay if I commit the files for rocketchat into a single commit in the updated fork?

Yes, a single commit would be great! Thanks

valeriocos:
Hi @Syn3rman, do you have any update on this PR? Plese ping me when I can review it, thanks!

valeriocos:
Hi @animeshk08 since this PR seems dead, would you like to take it over? thanks

animeshk08:
Hi @valeriocos, Sure I will complete and then start working on the enricher :)

valeriocos:
thanks @animeshk08 !

Issue #631, opened 2020-03-19
[Feature Request] Add support for Gitter backend


animeshk08:
[Gitter](https://gitter.im/) is a chat and networking platform that helps to manage, grow and connect communities through messaging, content and discovery.

It would be a good idea to include Gitter as a backend for Perceval.

The API documentation of Gitter can be found here: https://developer.gitter.im/docs/rest-api

valeriocos:
I like the idea @animeshk08! While waiting for others to comment on this issue, please can you start having a look at the API and check which data could be extracted (a couple of target Gitter instances can be useful to test the code too)?

Thanks

animeshk08:
Hi, I have had a look at the API documentation of Gitter.
We can fetch the messages from a Gitter room, similar to Slack.
The `List Messages` API found [here](https://developer.gitter.im/docs/messages-resource) can be used for that.

However, it does require the `roomid` endpoint which can be found in a way similar to `groupsio` backend where we list the rooms a user has joined using the `List rooms`API found [here](https://developer.gitter.im/docs/rooms-resource) match the room name and get the `roomid` for a room.

In terms of parameters, Perceval can take the `group name`, `room name` and `access token` as the parameters.

Note:
 1. A room belongs inside a group(community) hence a room name will typically be `{groupname}/{roomname}`.
1. An access token is necessary to authenticate yourself to the APIs


Some Gitter instances which can be used for testing may be:
1. https://gitter.im/jenkinsci/jenkins 
1. https://gitter.im/fossasia/fossasia
1. https://gitter.im/openMF/mifos

Please comment here for any suggestion :)

valeriocos:
Thank you @animeshk08 ! Your proposal looks good!

> In terms of parameters, Perceval can take the group name, room name and access token as the parameters.

+1 

> Some Gitter instances which can be used for testing may be:

+1

WRT the authentication, please consider to write a script to get the token (there is something similar for twitter https://gist.github.com/valeriocos/7d4d28f72f53fbce49f1512ba77ef5f6)

I wasn't able to find info about rate limit at https://developer.gitter.im/docs/. Please let me know if you find something, thanks!

animeshk08:
> WRT the authentication, please consider to write a script to get the token (there is something similar for twitter https://gist.github.com/valeriocos/7d4d28f72f53fbce49f1512ba77ef5f6)

Sure! Thank you for the reference.
 
[...]

PR #632, opened 2020-03-21
[backend] Add Gitter backend

animeshk08:
This commit fixes issue #631 
This commit adds Gitter backend to
Perceval.

Signed-off-by: animesh <animuz111@gmail.com>


animeshk08:
Hi @valeriocos. I have implemented Gitter backend for Perceval. Please try it out using the rooms mentioned here:

Some Gitter instances which can be used for testing may be:

1. https://gitter.im/jenkinsci/jenkins
1. https://gitter.im/fossasia/fossasia
1. https://gitter.im/openMF/mifos

Please note you have to join a room to fetch its messages. Also, an API token is mandatory for this backend. You can Generate one from here: https://developer.gitter.im/docs/welcome,
by Signing in and going to your `profile` or the `Apps` section at the top bar.

Thanks!

perceval/backends/core/gitter.py
```diff
@@ -0,0 +1,325 @@
...
+    def _init_client(self, from_archive=False):
+        """Init client"""
+
+        room = urijoin(self.group, self.room)
+
+        return GitterClient(self.api_token, self.max_items, self.archive,
+                            from_archive, self.ssl_verify, room=room)
+
+
+class GitterClient(HttpClient):
```
valeriocos: This class should inherits from [RateLimitHandler](https://github.com/chaoss/grimoirelab-perceval/blob/198c87e34e1acd7c8eee656471d77c812967b505/perceval/client.py#L223) too, since the API responses return rate limit information (see figue below). Such an info should be used to handle requests throttling/consumption.

captura_277

Example of the use of the RateLimitHandler are:
- https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/backends/core/gitlab.py#L464
- https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/backends/core/meetup.py#L362
- https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/backends/core/github.py#L571

In a nutshell, you have to:
- check if the ratelimit headers are the default ones (https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/client.py#L235). If not, you should redefine them (ex.: https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/backends/core/gitlab.py#L464)

- init the instance
  - https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/backends/core/meetup.py#L362

- redefine calculate_time_to_reset
  - https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/backends/core/meetup.py#L364
  - https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/backends/core/gitlab.py#L612
  - https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/backends/core/github.py#L625

perceval/backends/core/gitter.py
```diff
@@ -0,0 +1,325 @@
...
+        """
+        from_date = kwargs['from_date']
+        logger.debug("Get Gitter message paginated items of room: %s from date: %s", self.room, from_date)
+        from_date = str_to_datetime(from_date).timestamp()
+        fetching = True
+        before_id = None
+        num_msgs = 0
+        self.room_id = self.client.get_room_id()
+        logger.debug("Room Id of room: %s is %s", self.room, self.room_id)
+        page = 0
```
valeriocos: Please move the variable `page` before the call of `logger.debug`, and separate them with a blank line to improve readibility.

perceval/backends/core/gitter.py
```diff
@@ -0,0 +1,325 @@
...
+        """Fetch the messages
+
+        :param category: the category of items to fetch
+        :param kwargs: backend arguments
+
+        :returns: a generator of items
+        """
+        from_date = kwargs['from_date']
+        logger.debug("Get Gitter message paginated items of room: %s from date: %s", self.room, from_date)
+        from_date = str_to_datetime(from_date).timestamp()
```
valeriocos: Please add a blank line before the assignment of `from_date` to improve readibility. Thus what is after the `logger.debug` statement concerns only the initialization of the variables

perceval/backends/core/gitter.py
```diff
@@ -0,0 +1,325 @@
...
+
+    Client for fetching information from the Gitter server
+    using its REST API.
+
+    :param api_token: key needed to use the API
+    :param max_items: maximum number of items per request
+    :param archive: an archive to store/read fetched data
+    :param from_archive: it tells whether to write/read the archive
+    :param ssl_verify: enable/disable SSL verification
+    :param room: room from which the message are to be fetch.
```
valeriocos: This param may be removed if we decide to not pass the room information to the client.

perceval/backends/core/gitter.py
```diff
@@ -0,0 +1,325 @@
...
+        This contains the absolute path of the room i.e. {group}/{room}
+    """
+    AUTHORIZATION_HEADER = 'Authorization'
+    GET_MESSAGES_ENDPOINT = 'chatMessages'
+    GET_ROOMS_ENDPOINT = 'rooms'
+
+    def __init__(self, api_token, max_items=MAX_ITEMS, archive=None,
```
valeriocos: The client should implement the method `sanitize_for_archive`, which is used to remove the token information from the requests before storing them to the archive (ex.: https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/backends/core/gitlab.py#L684).

This method isn't present in other clients (like the ones below).
- https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/backends/core/pagure.py
- https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/backends/core/github.py

 An issue should be opened to check how many clients miss this method and if this is preventing to fetch archived data generated with the corresponding backends. @animeshk08 let me know if you want to open this issue, otherwise I can do it :)

perceval/backends/core/gitter.py
```diff
@@ -0,0 +1,325 @@
...
+        'message'.
+        """
+        return CATEGORY_MESSAGE
+
+    def _init_client(self, from_archive=False):
+        """Init client"""
+
+        room = urijoin(self.group, self.room)
+
+        return GitterClient(self.api_token, self.max_items, self.archive,
```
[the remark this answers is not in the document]
animeshk08: Hi @valeriocos. The only reason I have used roomid to init the client is because, I thought it would be a good idea to include the roomid in the search fields. Since roomid needs a separate API call as is not returned by the chatmessages API.
Let me know if it's not necessary? Or if you prefer any other implementation

perceval/backends/core/gitter.py
animeshk08: Thanks for the reference! I will open the issue:)

perceval/backends/core/gitter.py
```diff
@@ -0,0 +1,325 @@
...
+    :param api_token: key needed to use the API
+    :param max_items: maximum number of items per request
+    :param archive: an archive to store/read fetched data
+    :param from_archive: it tells whether to write/read the archive
+    :param ssl_verify: enable/disable SSL verification
+    :param room: room from which the message are to be fetch.
+        This contains the absolute path of the room i.e. {group}/{room}
+    """
+    AUTHORIZATION_HEADER = 'Authorization'
+    GET_MESSAGES_ENDPOINT = 'chatMessages'
```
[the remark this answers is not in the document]
animeshk08: Sure! Thank you for the info. I will get a PR ready addressing this issue!

perceval/backends/core/gitter.py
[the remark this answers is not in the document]
valeriocos: Sorry for not being clear @animeshk08 . The roomid in the search fields is a really good idea, but this is at the level of the backend. 

I was suggesting to remove the param `room` at https://github.com/chaoss/grimoirelab-perceval/pull/632/files#diff-6459a75c74db880450cc119cd49e3051R235. The info to retrieve the roomid can be passed as params of the method `get_room_id` (https://github.com/chaoss/grimoirelab-perceval/pull/632/files#diff-6459a75c74db880450cc119cd49e3051R278)

valeriocos:
@animeshk08 ping me when the PR is ready for review, thanks!

perceval/backends/core/gitter.py
```diff
@@ -0,0 +1,325 @@
...
+        """Returns the Gitter argument parser."""
+
+        parser = BackendCommandArgumentParser(cls.BACKEND,
+                                              from_date=True,
+                                              token_auth=True,
+                                              archive=True,
+                                              ssl_verify=True)
+
+        # Backend token is required
+        action = parser.parser._option_string_actions['--api-token']
```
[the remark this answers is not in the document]
animeshk08: Hi @valeriocos. I believe these lines are added to specify that an API token is necessary for this backend.
If these lines are removed and no API token is provided the error throw is: `401 Client Error: Unauthorized for url: https://api.gitter.im/v1/rooms`. 
However, if these are included and no API is provide the error thrown is `error: the following arguments are required: -t/--api-token`. 
Let me know if there is another approach fo making the `api-token` mandatory for this backend. :)

perceval/backends/core/gitter.py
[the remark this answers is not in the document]
animeshk08: Thank you for the clarification. I will make the change :)

animeshk08:
Hi, @valeriocos Sorry for the delay! I have made the changes requested :)

Also, I have added the tests and updated the docs :rocket: 

Can you please review the PR? Thank you :)

valeriocos:
Sure @animeshk08 , I'm on it, thanks!

perceval/backends/core/gitter.py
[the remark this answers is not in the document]
valeriocos: Good catch! I thought that the token was required. Thanks!

tests/test_gitter.py
```diff
@@ -0,0 +1,498 @@
...
+
+    def test_has_archiving(self):
+        """Test if it returns True when has_archiving is called"""
+
+        self.assertEqual(Gitter.has_archiving(), True)
+
+    def test_has_resuming(self):
+        """Test if it returns True when has_resuming is called"""
+
+        self.assertEqual(Gitter.has_resuming(), True)
```
valeriocos: This should be changed to `False`

perceval/backends/core/gitter.py
```diff
@@ -0,0 +1,378 @@
...
+        """
+        return True
+
+    @classmethod
+    def has_resuming(cls):
+        """Returns whether it supports to resume the fetch process.
+
+        :returns: this backend does not support items resuming
+        """
+        return True
```
valeriocos: I have been playing with the backend, and the messages returned are from the most recent to the last recent. In this case, the backend doesn't support resuming, because in case of failure the fetching process must start over.

The resuming is supported when the data is retrieved from the oldest to the most recent, because in case of failures the fetching process can start from the last item collected.

perceval/backends/core/gitter.py
```diff
@@ -0,0 +1,378 @@
...
+                            from_archive, self.ssl_verify)
+
+
+class GitterClient(HttpClient, RateLimitHandler):
+    """Gitter API client.
+
+    Client for fetching information from the Gitter server
+    using its REST API.
+
+    :param api_token: key needed to use the API
```
valeriocos: sleep_for_rate and min_rate_to_sleep are missing

perceval/backends/core/gitter.py
```diff
@@ -0,0 +1,378 @@
...
+        self.max_items = max_items
+
+        super().__init__(base_url, archive=archive, from_archive=from_archive,
+                         ssl_verify=ssl_verify)
+        super().setup_rate_limit_handler(sleep_for_rate=sleep_for_rate, min_rate_to_sleep=min_rate_to_sleep)
+
+    def calculate_time_to_reset(self):
+        """Number of seconds to wait. They are contained in the rate limit reset header"""
+
+        time_to_reset = self.rate_limit_reset_ts - (datetime_utcnow().replace(microsecond=0).timestamp() + 1)
```
valeriocos: When the rate limit is over, the time to reset is not logged correctly. Please check if the following code improves the log message:
```
-        time_to_reset = self.rate_limit_reset_ts - (datetime_utcnow().replace(microsecond=0).timestamp() + 1)
+        current_epoch = (datetime_utcnow().replace(microsecond=0).timestamp() + 1) * 1000
+        time_to_reset = (self.rate_limit_reset_ts - current_epoch) / 1000
```

tests/test_gitter.py
```diff
@@ -0,0 +1,498 @@
...
+        from_date = datetime.datetime(2020, 3, 1, 0, 0, tzinfo=dateutil.tz.tzutc())
+
+        args = ['-t', 'aaa',
+                '--tag', 'test',
+                '--max-items', '10',
+                '--no-ssl-verify',
+                '--min-rate-to-sleep', '1',
+                '--from-date', '2020-03-01',
+                'testapicomm', 'community',
+                ]
```
valeriocos: please move the brakets to the previous line, thanks

perceval/backends/core/gitter.py
valeriocos: It seems that it isn't possible to fetch the messages in ASC order. If you find something, please let me know.

animeshk08: Thank you for the info :)
I rechecked the Gitter API doc. The API does not support fetching the messages in ascending order. I will make this change :)

animeshk08:
Hi, @valeriocos. I have made the changes. Please review the PR :)
Thanks

animeshk08:
Thanks, @valeriocos. I will start working on the issues and PRs discussed here :)

PR #636, opened 2020-03-29
[backend] Add predefined paramaters to clients

animeshk08:
This commit removes string values from client backends
and uses predefined values instead inside the methods.

### DIFF
--- a/perceval/backends/core/askbot.py
+++ b/perceval/backends/core/askbot.py
@@ -266,11 +266,28 @@ class AskbotClient(HttpClient):
     """
 
     API_QUESTIONS = 'api/v1/questions/'
-    HTML_QUESTION = 'question/'
-    ORDER_API = 'activity-asc'
-    ORDER_HTML = 'votes'
-    COMMENTS = 's/post_comments'
-    COMMENTS_OLD = 'post_comments'
+
+    # API resources
+    RHTML_QUESTION = 'question/'
+    RCOMMENTS = 's/post_comments'
+    RCOMMENTS_OLD = 'post_comments'
+
+    # API header
+    HREQUEST_WITH = 'X-Requested-With'
+
+    # Resource parameters
+    PPAGE = 'page'
+    PSORT = 'sort'
+    PPOST_ID = 'post_id'
+    PPOST_TYPE = 'post_type'
+    PAVATAR_SIZE = 'avatar_size'
+
+    # Predefined values
+    VORDER_API = 'activity-asc'
+    VORDER_HTML = 'votes'
+    VANSWER = 'answer'
+    VAVATAR_SIZE = 0
+    VHTTP_REQUEST = 'XMLHttpRequest'
 
     def __init__(self, base_url, archive=None, from_archive=False, ssl_verify=True):
         super().__init__(base_url, archive=archive, from_archive=from_archive, ssl_verify=ssl_verify)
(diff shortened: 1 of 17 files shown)


animeshk08:
Hi @valeriocos. Based on discussions [here](https://github.com/chaoss/grimoirelab-perceval/pull/632#discussion_r396074177). I have started a PR. Can have a look at these two backends to check it matches the expected results. I have followed [discourse](https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/backends/core/discourse.py) backend as many backends had different parameter names. However, discourse was consistent.

Thanks :)

valeriocos:
Hi @animeshk08, thank you for opening this PR!

> Can have a look at these two backends to check it matches the expected results.

Yes, the refactor of the 2 backends look good!

animeshk08:
Thanks @valeriocos! I will extend the same to other backends.

animeshk08:
Hi @valeriocos. I am done with the refactor. Let me know if any changes are required :)

animeshk08:
Thank you @valeriocos. Sure that's a great idea.

I had a question.

How do you guys keep a track of the API? If an API changes its format do we make the change in our backend/tests. How often do we do that?

I ask this because recently Pagure changed its API for issues. It came to my attention only because of the errors in the test. They have now references to related PRs of issues that were not there when I implemented it.

valeriocos:
Sorry @animeshk08 for the delay, I missed this notification!

> How do we keep a track of the APIs? If an API response changes its format do we make the change in our backend/tests. How often do we do that?

Good questions! For some APIs (e.g., Slack, Groupsio) we keep an eye on their websites: 
- https://api.slack.com/changelog
- https://beta.groups.io/g/api/topics

As fallback option, we rely on the logs generated by the different GrimoireLab instances running in our servers. Once we detect a change, we provide a fix and we update the tests.

We don't have scheduled activities to review or monitor the API changes, since the APIs we rely on are generally stable and breaking changes are rare. AFAIR, the last changes were needed on groupsio (which has a pretty recent API).

It would be great to have a link for each API used by the Perceval backends, however I'm not sure if all APIs have it. If you want, we can start with a gist and add there the info we found for each API. Once the gist is ready, we can upload it to the Perceval repo, WDYT?

animeshk08:
Thanks for the reply @valeriocos.

 > It would be great to have a link for each API used by the Perceval backends, however I'm not sure if all APIs have it. If you want, we can start with a gist and add there the info we found for each API. Once the gist is ready, we can upload it to the Perceval repo, WDYT?

I like the idea. What do you think we should include in this script? The API endpoint and its response? Or should it be more like a changelog where we update the different version of the API and mention at which point we decided to make a change in the Perceval backend? Maybe both could be implemented or something else can also be done. What are your thoughts?

valeriocos:
I was thinking about something simpler ^^ (a file with the link to the API documentation), but your idea looks much better! Can you elaborate on what to include in the script?

animeshk08:
> I was thinking about something simpler ^^ (a file with the link to the API documentation), but your idea looks much better! Can you elaborate on what to include in the script?

Sure!. I have a few possible implementations in mind. I will explore them a bit and open the issue by tomorrow. 

Also any updates on this PR?

valeriocos:
I guess it can be merged, further improvements can be tackled in another PR.

animeshk08:
Thanks, @valeriocos. I needed this merge for the sanitize archive PR :)

Issue #635, opened 2020-03-29
[Feature Request] Add support for Airtable 


animeshk08:
[Airtable](https://airtable.com/) is a spreadsheet-database hybrid, with the features of a database but applied to a spreadsheet. 

The API documentation for Airtable can be found here: https://airtable.com/api

NOTE: You need to signin and create an workspace to access the API docs.

animeshk08:
The `List records` endpoint can be used for fetching all the records(Sorry cannot provide the link as the API doc is workspace dependent).

An authorization token is **necessary** for Airtable backend and can be passed in the header.

A sample response is:
```
{
    "records": [
    {
      "id": "recMeO3iru1KoQVPI",
      "fields": {
        "uuid": "uuid",
        "project_1": "project_1",
        "project": "project",
        "origin": "origin",
        "body_extract": "body_extract",
        "subject_analysed": "Subject_analyzed"
      },
      "createdTime": "2020-03-19T19:55:15.000Z"
    },
    {
      "id": "recJ7LVnygXh22yCv",
      "fields": {
        "uuid": "7603db8137ef75c3f924cd05de583a4506617062",
        "project_1": "Chaoss",
        "project": "Chaoss",
        "origin": "https://groups.io/g/zabiplane",
        "body_extract": "*Reminder:* Zabiplane R&D and Use-Case Working Groups\n\n*When:* Thursday, 12 March 2020, 8:30am to 9:30am, (GMT-07:00) America/Los Angeles\n\n*Where:* https://meet.google.com/euc-vtop-nny\n\nView Event ( https://zabiplane.groups.io/g/dev/viewevent?eventid=711445 )\n\n*Organizer:* Carlo Contravali and Gary Greenberg\n\n",
        "subject_analysed": "Upcoming Event: Zabiplane R&D and Use-Case Working Groups - Thu, 03/12/2020 8:30am-9:30am #cal-reminder"
      },
      "createdTime": "2020-03-19T19:55:15.000Z"
    },
    {
      "id": "recXduAatagswJFMv",
      "fields": {
        "uuid": "22f001738a33e270deaa09f4ed8bc3bb123de5ce",
        "project_1": "Chaoss",
        "project": "Chaoss",
        "origin": "https://groups.io/g/zabiplane",
[...]

PR #640, opened 2020-04-03
[backend] Add Airtable backend

animeshk08:
Fixes: #635 
This commit adds support to fetch
records from an Airtable table.

### DIFF
--- a/perceval/backends/core/airtable.py
+++ b/perceval/backends/core/airtable.py
@@ -1,0 +1,345 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2015-2020 Bitergia
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+# Authors:
+#     Animesh Kumar <animuz111@gmail.com>
+#
+
+import logging
+from grimoirelab_toolkit.datetime import str_to_datetime, datetime_to_utc
+from grimoirelab_toolkit.uris import urijoin
+from requests.exceptions import HTTPError
+from requests.utils import quote
+
+from ...backend import (Backend,
+                        BackendCommand,
+                        BackendCommandArgumentParser)
+from ...client import HttpClient
+from ...utils import DEFAULT_DATETIME
+
+CATEGORY_RECORD = "record"
+
+AIRTABLE_URL = 'https://airtable.com/'
+AIRTABLE_API_URL = 'https://api.airtable.com/v0'
+DEFAULT_SEARCH_FIELD = 'item_id'
+
+logger = logging.getLogger(__name__)
+
+
+class Airtable(Backend):
+    """Airtable backend.
+
+    This class retrieves the records created in an Airtable table.
+    To access the server an API token is required.
+
+    The origin of the data will be set to the `AIRTABLE_URL` plus the
+    identifier of the table; i.e 'https://airtable.com/{table}'.
+
+    :param workspace_id: ID of the workspace containing the table
+    :param table: identifier of the table from which the records are to be fetched
+    :param api_token: token or key needed to use the API
+    :param max_items: maximum number of records requested on the same query
+    :param tag: label used to mark the data
+    :param archive: archive to store/retrieve items
+    :param ssl_verify: enable/disable SSL verification
+    """
+    version = '0.1.0'
+
+    CATEGORIES = [CATEGORY_RECORD]
+
+    def __init__(self, workspace_id=None, table=None, api_token=None, max_items=None,
+                 tag=None, archive=None, ssl_verify=True):
+
+        # Table name may have non URL format (e.g. include spaces)
+        table = quote(table)
+        origin = urijoin(AIRTABLE_URL, table)
+
+        super().__init__(origin, tag=tag, archive=archive, ssl_verify=ssl_verify)
+        self.workspace_id = workspace_id
+        self.table = table
+        self.api_token = api_token
+        self.max_items = max_items
+        self.client = None
+
+    def search_fields(self, item):
+        """Add search fields to an item.
+
+        It adds the values of `metadata_id` and `table`
+
+        :param item: the item to extract the search fields values
+
+        :returns: a dict of search fields
+        """
+        search_fields = {
+            DEFAULT_SEARCH_FIELD: self.metadata_id(item),
+            'table': self.table,
+        }
+
+        return search_fields
+
+    def fetch(self, category=CATEGORY_RECORD, from_date=DEFAULT_DATETIME):
+        """Fetch the records from the table.
+
+        This method fetches the records that were created in the table
+        since the given date.
+
+        :param category: the category of items to fetch
+        :param from_date: date from which records are to be fetched
+
+        :returns: a generator of records
+        """
+        if not from_date:
+            from_date = DEFAULT_DATETIME
+
+        from_date = datetime_to_utc(from_date).isoformat()
+        kwargs = {
+            'from_date': from_date,
+        }
+
+        items = super().fetch(category, **kwargs)
+
+        return items
+
+    def fetch_items(self, category, **kwargs):
+        """Fetch the records.
+
+        :param category: the category of items to fetch
+        :param kwargs: backend arguments
+
+        :returns: a generator of items
+        """
+        from_date = kwargs['from_date']
+
+        logger.debug("Get Airtable records paginated items of table: %s from date: %s", self.table, from_date)
+
+        from_date = str(from_date)
+        fetching = True
+        offset = None
+        num_records = 0
+        page = 0
+
+        while fetching:
+            page += 1
+            logger.debug("Page: %i" % page)
+            raw_record_group = self.client.record_page(self.workspace_id, self.table, offset, from_date)
+            raw_record_group = raw_record_group.json()
+
+            if "offset" in raw_record_group:
+                offset = raw_record_group["offset"]
+            else:
+                fetching = False
+
+            record_group = raw_record_group["records"]
+
+            if not record_group:
+                fetching = False
+                continue
+
+            for raw_record in record_group:
+                num_records += 1
+                yield raw_record
+
+        logger.debug("Fetch process completed: %s records fetched", num_records)
+
+    @classmethod
+    def has_archiving(cls):
+        """Returns whether it supports archiving items on the fetch process.
+
+        :returns: this backend supports items archive
+        """
+        return True
+
+    @classmethod
+    def has_resuming(cls):
+        """Returns whether it supports to resume the fetch process.
+
+        :returns: this backend does not support items resuming
+        """
+        return False
+
+    @staticmethod
+    def metadata_id(item):
+        """Extracts the identifier from an Airtable item."""
+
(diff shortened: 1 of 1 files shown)


animeshk08:
Some issues with this implementation:

* The workspace name cannot be retrieved from any API. Also, it is not needed to fetch the records either. Maybe we can include it as an optional param. Since workspace name is not available the current  `origin` is set to https://airtable.com/{table}

* If a record is missing some fields those fields are not returned by the API. Let me know if this needs special handling. if yes, Should it be handled here or at ELK?

* Cannot fetch records sorted by date. Records are returned sorted **alphabetically** by the record id.

**How to test:**
Airtable is not a public data source. However, the testing can still be done either on the sample workspaces or by importing a spreadsheet.

The workspace I have used for testing can be found here: https://airtable.com/invite/l?inviteId=inv1ZJweDnNNSByAm&inviteToken=4f745c93726f184fd0a4d22d9459ff496619b9ee52a8cf516af11d55abfa2583

**Sample usage:**

* Return 3 records.
```
airtable -t xxx --from-date "2020-03-29 13:00:00" workspace_id "Imported table"
```

* Return 701 records(paginated items).
```
airtable -t xxx workspace_id "Table 2"
```

* Returns 1 item.
```
airtable -t xxx --from-date 2020-04-01 workspace_id "Table 2"
```

> NOTE: An API token is needed for this backend it can be generated by creating an Airtable user account and generating a key here: https://airtable.com/account
The workspace(base) ID is found inside the API documentation here: https://airtable.com/api

animeshk08:
Hi @valeriocos. I have implemented the Airtable backend. It may take some efforts to create test workspace for this backend(sorry about that :see_no_evil:) 

Please review the PR and let me know thw changes :)

sduenas:
Hi @animeshk08, first of all, thanks for your backend. 

I think we shouldn't include this backend as part of the core of perceval. The idea of perceval is to analyze data from datasources/repositories used by software developers. I don't think Airflow can be considered as part of that kind of repositories. I understand people might use it but it's not a common thing. Also, Airflow doesn't have a data schema in the way of Bugzilla, GitHub and others have.

However, we can consider to open a new repository to include these other kind of backends, that might be useful to someone but that won't be part of the core of GrimoireLab.

animeshk08:
Hi @sduenas. Thank you for going through this PR.

> I think we shouldn't include this backend as part of the core of perceval. The idea of perceval is to analyze data from datasources/repositories used by software developers. I don't think Airflow can be considered as part of that kind of repositories. I understand people might use it but it's not a common thing. Also, Airflow doesn't have a data schema in the way of Bugzilla, GitHub and others have.

I completely agree with you. I myself was a bit reluctant about this backend, hence I thought of mentioning all the issues with the AIrtable backend. Even though the data returned is rich, however, it's not organized and doesn't match our use case. Based on the discussions I have read on Airtable forums there won't be any changes in the API anytime soon are the community is more or less in support of the current format(however, they do want a `last_modified` field in the API).

> However, we can consider to open a new repository to include these other kind of backends, that might be useful to someone but that won't be part of the core of GrimoireLab.

Yes, that is a good idea. Maybe we can do that once we have a bunch of such backends in mind. Right now I cannot think of anything other than Airtable and Google Sheets. I will keep this implementation in a separate branch of mine and also it will be present in this PR(closed). If the use case arise it would be a good reference/starting point.

Also, the implementation may be helpful to anyone who is working on the SCMS idea for GSoC(https://github.com/chaoss/grimoirelab/issues/288). Hence I have mentioned it here :)

\cc @valeriocos

valeriocos:
Hi @animeshk08 , thank you for the PR, and for triggering an interesting discussion.

By having a look at the current Perceval backends, it seems that there is no clear rule to decide which datasources/repositories are used by software developers. For instance, the backends for Rss, Telegram, Twitter and Googlehits seem to be on the border (or outside this definition). At the same time, other backends available in Perceval-related repos[*] could be promoted to the main Perceval repo since useful for some software developers communities (e.g., functest, puppetforge).

It would be probably convenient to define some rules (e.g., relation with software development, popularity of the datasource) to decide what goes into Perceval and what does not. Once we have clear rules, we could divide the backends between Perceval main repo and another Perceval (e.g., `perceval-extensions`).

[*] Perceval-related backends:
- https://github.com/chaoss/grimoirelab-perceval-mozilla
- https://github.com/chaoss/grimoirelab-perceval-opnfv
- https://github.com/chaoss/grimoirelab-perceval-puppet
- https://github.com/Bitergia/grimoirelab-perceval-finos

animeshk08:
Hi @valeriocos. That's a good way to move forward. If possible I would also like to join the discussion regarding the definition of these rules. :)

valeriocos:
> If possible I would also like to join the discussion regarding the definition of these rules

Sure, that's a good idea! Thanks

animeshk08:
The point raised in this https://github.com/chaoss/grimoirelab-perceval/pull/645#issuecomment-610832196 is valid. Continous addition of backends without the actual need will make it hard to maintain them. 
As mentioned by @valeriocos a set of rules may be helpful to decide the new backends. However, defining such a list may be hard as the backends are so diverse. 
 
I think an alternative would be to prepare a list of questions that should be included with the feature request issue. 

Some of these questions could be:

* Approx user base.
* Average traffic if available.
* Approx open source user base.
* List popular projects using the platform.
* Similar platforms.

This will help in analyzing whether the backend is actually needed and the decision could be communicated inside the issue itself.

WDYT? @valeriocos @sduenas

valeriocos:
Hi @animeshk08 , thank you for moving forward this discussion. 

By looking at the proposals to integrate Airtable and Gitee (https://github.com/chaoss/grimoirelab-perceval/pull/645), I understand that even if we have a set of questions to decide where a backend goes, the final decision is up to the maintainers.

To simplify the contributions of new backends, it is probably easier to say that all new backends go to the other Perceval repo (mentioned at https://github.com/chaoss/grimoirelab-perceval/pull/640#issuecomment-609037285), which would act as an incubator. Then, it is up to the maintainers to decide when/if the backend will be promoted to the main Perceval. WDYT?

animeshk08:
> To simplify the contributions of new backends, it is probably easier to say that by default all new backends go to the other Perceval repo (mentioned at #640 (comment)), which would act as an incubator. Then, it is up to the maintainers to decide when/if the backend will be promoted to the main Perceval repo. This decision can happen directly when the contributor opens an issue to propose a new backend.

I am in support of this. I just have a few questions :)

* Will all the future backends follow this flow?
* Should these backends be maintained/checked for a possible shift to main repo regularly? If yes how often. Maybe after every major version upgrade of Perceval?
* Having a dedicated branch in the main repo or having a separate repo, which do you think will be better? I raise this as GrimoireLab already has so many repositories.

Also, my suggestion for the set of questions was inspired by the comments:  https://github.com/chaoss/grimoirelab-perceval/pull/645#issuecomment-610262313 and https://github.com/chaoss/grimoirelab-perceval/pull/629#issuecomment-600542687

I think we can create a template and reference it for anyone creating an issue/ PR for a new backend.

valeriocos:
> Will all the future backends follow this flow?

I'd say yes. The advantage of this flow is that a contributor has the default option to submit the backend to the incubator repo. Thus, all discussions that could slow down the contribution of a new backend are postponed and may involved only the maintainers.

> Should these backends be maintained/checked for a possible shift to main repo regularly? If yes how often.

I'd say yes. Once or twice per year should be OK, since an API may take time to become stable.

> Maybe after every major version upgrade of Perceval?

We don't have major version upgrade of Perceval and there is no plan for the release 1.0.0.

> Having a dedicated branch in the main repo or having a separate repo, which do you think will be better? I raise this as GrimoireLab already has so many repositories.

I'd say a separate repo, since a branch won't be really visible. If we merge the different Perceval repos into two, we will reduce the global number of repos.

> I think we can create a template and reference it for anyone creating an issue/ PR for a new backend.

This is a really interesting point. I'd love to use templates for issues and pull requests (not just related to new backends), this would probably simplify the contribution process and open the doors for automation. We can move this point to another issue (maybe better in the GrimoireLab main repo) and see what the community thinks about it. However, if you see it convenient, we can first focus on the problem around the Perceval repos and then move to the templates.

animeshk08:
> I'd say yes. Once or twice per year should be OK, since an API may take time to become stab

This also sounds great.

> I'd say a separate repo, since a branch won't be really visible. If we merge the different Perceval repos into two, we will reduce the global number of repos.

If we are planning to merge the Perceval repository, then we shiftig to a new repository sounds better.

> However, if you see it convenient, we can first focus on the problem around the Perceval repos and then move to the templates.

Yes, the idea for creating templates is good. Let's address this after restructuring the Perceval repositories

WillemJiang:
> Hi @animeshk08 , thank you for moving forward this discussion.
> 
> By looking at the proposals to integrate Airtable and Gitee (#645), I understand that even if we have a set of questions to decide where a backend goes, the final decision is up to the maintainers.
> 
> To simplify the contributions of new backends, it is probably easier to say that by default all new backends go to the other Perceval repo (mentioned at [#640 (comment)](https://github.com/chaoss/grimoirelab-perceval/pull/640#issuecomment-609037285)), which would act as an incubator. Then, it is up to the maintainers to decide when/if the backend will be promoted to the main Perceval repo. This decision can happen directly when the contributor opens an issue to propose a new backend.
> 
> WDYT?

I fully support this proposal. In this way we don't make a high the bar for the contributor, and the community can keep contributing to the upstream  repo, finally the maintainers still have the last call for the backend. BTW, it could save us lot of time when the core APIs are changed if we have unit tests for the backend.

valeriocos:
Thank you @WillemJiang for joining the discussion! Let's wait for @sduenas 's opinion.

sduenas:
I've been thinking about this for a few days. I think the best option for these cases won't be to have an incubator repository because in the end we will have to maintain it.

I think a better approach is people should have their own repositories where they can store their own backends. We can link those repositories in our README file as third-party backends. At some point, if we see that it would be a good option to add them to the main repository, we will ask that to the owner or maintainer of that backend to integrate with the core.

What do you think? Opinions?

WillemJiang:
+1 to links to README for the third party backend.  
We may also need to provide an easy way to integrate the third party backend with the latest core, otherwise it could introduce some trouble when the user wants to try the third party backend.

valeriocos:
I understand that your proposal applies only to some backends (e.g., `Gitee` and `Airtable`) for which you aren't sure. Other backends, more used and popular, such as Rocketchat and Zulip would skip this process, right?

Said that, I understand that linking some backends and having a incubator repository for other backends are compatible solutions.

The incubator repository could be maintained by the same backend's contributors. This would also help defining a path to promote those contributors to GrimoireLab maintainers. ATM there is no option to nourish and grow our community.

> I think a better approach is people should have their own repositories where they can store their own backends.

I understand that this is the current scenario we have. People may have in their forks some Perceval backends, which are not shared with this repo.

> We can link those repositories in our README file as third-party backends. 

Which are the rules to link a backend? Should the backend pass some code review process before listing it in the README?

> We may also need to provide an easy way to integrate the third party backend with the latest core, otherwise it could introduce some trouble when the user wants to try the third party backend.

I agree with the comment of @WillemJiang. This also may also impact on the other components (ELK, Mordred and Sigils).

animeshk08:
I am more in support with @valeriocos approach. However, if we decide to not have an incubator repository the best way would be to have a discussion inside the issue itself regarding whether a said backend should be implemented.(Based on how aged the APIs are, number of users etc.)

vchrombie:
Adhering to the [contributing guidelines (#incubating-repositories)](https://github.com/chaoss/grimoirelab/blob/master/CONTRIBUTING.md#incubating-repositories) and after communicating with @animeshk08, this work is moved to a separate repository [perceval-backends/grimoirelab-perceval-airtable](https://github.com/perceval-backends/grimoirelab-perceval-airtable) and will be maintained over there for some time. We can open this PR again and update it when we want to merge this backend into the main module.

I'll be opening a PR adding the repo to the readme. The pull request and the issue will be closed soon if no one has any objection.

Best,
Venu

vchrombie:
Closing the PR in favour of https://github.com/chaoss/grimoirelab-perceval/pull/640#issuecomment-915882067.

Issue #644, opened 2020-04-07
[Feature Request] Add Gitee backend support


WillemJiang:
[Gitee](https://gitee.com/) provides similar service as Github does.
Here are the [API document](https://gitee.com/api/v5/swagger) of Gitee, I'm sorry there is only Chinese version. 

Here are some notes for the Gitee backend support
* Gitee doesn't have the ratelimit check, but I think we can still reuse the rat-limit check as Github client does.
* The issue API of Gitee is quite similar with Github issue API, it doesn't provide  issue and comment reaction.
*  Gitee PR API is quite different from Github PR API,  we can get PR ,comments and commits related information, but I cannot find out the merged by information from PR request. We may need to revisit the code when doing some work on grimoirelab-elk support.

WillemJiang:
I did some work by porting github backend to gitee, I will submit a PR (working in process) for it.

valeriocos:
Thank you @WillemJiang for opening the issue and kickstarting the PR (https://github.com/chaoss/grimoirelab-perceval/pull/645)!

PR #645, opened 2020-04-07
[backend] Added Gitee backend support

WillemJiang:
This PR provides [Gitee](https://gitee.com/) backend support.


valeriocos:
Can you provide a couple of repos to test the backend, and also the instructions to access the API (if needed)? Thanks!

WillemJiang:
@valeriocos  sure, I will create a test repo in gitee and add some comment on how to access the API.

valeriocos:
Thanks @WillemJiang

WillemJiang:
@valeriocos I just update the issue #644 with the detail API calls with curl.

sduenas:
I'm also not sure if we want to integrate this into the core of perceval. I understand somebody would be interested on it but not sure how many. The main problem is to maintain this backend (and others through time), if we add it, that means that any other change we make will be supported by this backend. Also, if nobody uses it, it might get deprecated fast.

I'd be happy to explore the idea of having a program or similar to integrate new datasources but I'm not sure how.

We can think of having a separate repository for these backends (but someone else will have to maintain them at least when Perceval changes). Althought, I think the best idea would be to help people to have their own repositories where they maintain their own backends. At some point they might decide to contribute with that backend to the core and we can evaluate if is mature and stable enough to support it.

WillemJiang:
@sduenas Thanks for you comment. As Gitee provides a similar service as Github does, and it's quite popular in China. I think it's could a good addon feature for grimoirelab.  

 I think upstream first is a good practise in open source to save the cost of maintenance, and it's benefit for the whole ecosystem.  

From my work experience of [Apache Camel](https://github.com/apache/camel) which has 300+ community contribute components, it doesn't add much work on the mentainer, as we have unit and integration tests to help us to spot the changes of camel-core which could have side effect on the camel components.  And these components make Apache Camel more popular, as the users can easily find a handy component for their daily work.

I agree we need to give some time for the new bankend to be mature enough.   Maybe we can add a contribution repo in grimorelab to host this new added backend first.  If the backend is mature enough (we need to define a rule for it), we could consider to move it into the core module. At the meantime we can still encourage community contributor to contribute the new backend to grimoirelab.

valeriocos:
Hi @WillemJiang,

> I agree we need to give some time for the new bankend to be mature enough. Maybe we can add a contribution repo in grimorelab to host this new added backend first. If the backend is mature enough (we need to define a rule for it), we could consider to move it into the core module. At the meantime we can still encourage community contributor to contribute the new backend to grimoirelab.

The discussion around this issue and the proposal to have a contribution repo seem to be aligned with the discussion around another issue (https://github.com/chaoss/grimoirelab-perceval/pull/640). Please have a look and join the discussion if you want, thanks!

valeriocos:
@sduenas to avoid being stucked on the decision where to include this backend, please consider that the backend can be added it to the Perceval main repo, and then the maintainers will decide where to move it (if needed).

vchrombie:
Closing this PR as per https://github.com/chaoss/grimoirelab-perceval/issues/644#issuecomment-907635708

PR #648, opened 2020-04-08
[github] Fetch issue/pull request events

valeriocos:
This code adds a new category to fetch the following event types from issues and pull
requests:
- ADDED_TO_PROJECT_EVENT
- MOVED_COLUMNS_IN_PROJECT_EVENT
- REMOVED_FROM_PROJECT_EVENT
- CROSS_REFERENCED_EVENT
- LABELED_EVENT
- UNLABELED_EVENT
- CLOSED_EVENT

Due to the limitation of not fetching events after a given date from the endpoint `timeline`
of GitHub v3, the events are fetched via the GitHub v4 (based on GraphQL).

The events are fetched in the following way. The issues of a given tracker are retrieved in
ascending order based on the last time they were updated. For each issue, its events (optionally
from/to a given date) are collected using a GraphQL call. Each event is returned by Perceval together with the corresponding issue (available in `data.issue`).

No user information beyond the `login` is included in data returned by Perceval for this category. Thus, the filter classified support has no effect when fetching events.

The backend can be executed in the following way:
```
perceval github chaoss grimoirelab-toolkit
--category event -t xxx
--sleep-for-rate
--no-archive
--from-date 2019-01-01
```

Backend version is now 0.26.0

### DIFF
--- a/perceval/backends/core/github.py
+++ b/perceval/backends/core/github.py
@@ -48,6 +48,7 @@ from ...utils import DEFAULT_DATETIME, DEFAULT_LAST_DATETIME
 CATEGORY_ISSUE = "issue"
 CATEGORY_PULL_REQUEST = "pull_request"
 CATEGORY_REPO = 'repository'
+CATEGORY_EVENT = "event"
 
 GITHUB_URL = "https://github.com/"
 GITHUB_API_URL = "https://api.github.com"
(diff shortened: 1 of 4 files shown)


valeriocos:
> As it retrieves events from one issue/PR and then it moves to the next one, events will be returned with no orther. It means we can not resume data and I'm not sure how these data are gonna be processed in later stages.

This is a limitation. It doesn't seem possible (or at least I didn't find how) to collect timeline events from a repository after a given date.

> Why did you decide to retrieve these events like a different category and not inside of each issue/PR?

The incremental fetching of issues/pull requests is based on the field `updated_at`, however this field isn't updated when events occur on them. Please note that in some cases the field is updated as a side-effect of the event, for instance closing an issue will (i) create a closedevent and (ii) cause the `updated_at` to be modified. This doesn't happen for events concerning the movements in the issue tracker kanban or crossreference events.

> Is there anyway to split the GraphQL query in smaller strings? It's ~150 lines long.

Yes, that's possible. I left it as a unique block since it is a query tailored to a specific need, and ATM I'm not sure that parts of it could be reused by other queries.

valeriocos:
After discussing with @sduenas , the goal is to move the category to a new backend.

valeriocos:
Related to #652

valeriocos:
Replaced by #652

PR #652, opened 2020-04-22
[perceval] Add githubql backend

valeriocos:
This class allows the fetch the issue events of a GitHub repository. Note that the events retrieved included also the ones of pull requests, since in GitHub, every pull request is an issue, but an issue may not be a pull request. Pull requests can be identified by the attribute `pull_request` included in `data.issue`.

The following event types from issues and pull requests:
- ADDED_TO_PROJECT_EVENT
- MOVED_COLUMNS_IN_PROJECT_EVENT
- REMOVED_FROM_PROJECT_EVENT
- CROSS_REFERENCED_EVENT
- LABELED_EVENT
- UNLABELED_EVENT
- CLOSED_EVENT

Due to the limitation of not fetching events after a given date from the endpoint `timeline` of GitHub v3, the events are fetched via the GitHub v4 (based on GraphQL).

All issues of a given tracker are retrieved in ascending order based on the last time they were updated. For each issue, its events (optionally from/until a given date) are collected using a GraphQL call. Each event is returned by Perceval together with the corresponding issue (available in data.issue).

Since the events are collected issue by issue, the incremental fetching is not supported. This limitation is due to the fact that events that occur on an issue may not update the issue attributes. Since there is no way to identify new events from the attributes of an issue, all issues must be fetched for every execution.

No user information beyond the login is included in data returned by this backend. Thus, the backend doesn't require filter classified support.

The backend can be executed in the following way:
```
perceval githubql chaoss grimoirelab-toolkit
--category event -t xxx
--sleep-for-rate
--no-archive
--from-date 2019-01-01
```

Backend version is 0.1.0

### DIFF
--- a/tests/test_githubql.py
+++ b/tests/test_githubql.py
@@ -1,0 +1,664 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2015-2020 Bitergia
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+# Authors:
+#     Valerio Cosentino <valcos@bitergia.com>
+#
+
+import datetime
+import os
+import unittest.mock
+
+import httpretty
+import pkg_resources
+
+pkg_resources.declare_namespace('perceval.backends')
+
+from perceval.client import RateLimitHandler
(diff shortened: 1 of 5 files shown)


tests/data/github/github_events
```diff
@@ -0,0 +1,52 @@
+{
```
sduenas, line 1: Do you think it would be better to create a new directory `githubql` to store this data?

tests/test_githubql.py
```diff
@@ -0,0 +1,528 @@
...
+GITHUB_ENTERPRISE_ISSUES_URL = GITHUB_ENTREPRISE_REPO_URL + "/issues"
+
+
+def read_file(filename, mode='r'):
+    with open(os.path.join(os.path.dirname(os.path.abspath(__file__)), filename), mode) as f:
+        content = f.read()
+    return content
+
+
+class TestGitHubQLBackend(unittest.TestCase):
```
sduenas, line 62: I only miss a test checking the pagination.

tests/data/github/github_events
valeriocos: I understand that the test data is part of GitHub, so it can make sense to leave in the same folder. However if you find it convenient, we can move it to a separate folder.

tests/test_githubql.py
valeriocos: Added, please check again

Issue #367, opened 2018-03-31
Support for Bitbucket, Sonar and Squash 


Avfdo1994:
Is it possible to use perceval with bitbucket, sonar and squash?

acs:
The current supported data sources are described at:

https://github.com/chaoss/grimoirelab-perceval/blob/master/README.md

There is no support for those data sources.

acs:
http://www.squashtest.org/en squash data source is this one?

Avfdo1994:
@acs Yes, it's http://www.squashtest.org/en

Avfdo1994:
Is there any alternative/workaround of using Bitbucket with Grimoire?

jgbarah:
> Is there any alternative/workaround of using Bitbucket with Grimoire?

If you're interested only in git repos, you can easily use the git backend. If you're interested in pull requests, they are not supported yet, a Perceval backend would be needed.

sergiocasaleiro:
> If you're interested only in git repos, you can easily use the git  backend.

Does this mean interested on using Perceval with Bitbucket without the PRs?
If so, how can we specify the authentication?
Thanks!

jgbarah:
> Does this mean interested on using Perceval with Bitbucket without the PRs?

Yes.

> If so, how can we specify the authentication?

If the repository is public, nothing to do, since Perceval uses git for accessing it, and there are no calls too the Bitbucket API. If the repository is private, if you already have your account configured to access it, it should just work, because Perceval is basically running git clone. Please, let us know if it does not work in either of these cases: to be honest, my experience with analyzing Bitbucket is not that large.

aaronnickovich:
I'd like to contribute a bitbucket interface for this project. bitbucket has similar APIs to github.

sduenas:
That would be great! Please take as reference [github.py](https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/backends/core/github.py) and [gitlab.py](https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/backends/core/gitlab.py) as reference.

valeriocos:
Hi @aaronnickovich, thank you for your interest! If you need some help don't hesitate to ping us!

sarthak0906:
Hello everyone,
Is anyone working on the sonarqube interface for the project? If it is still open I would like to contribute to it.

valeriocos:
Hi @sarthak0906! You're contribution is more than welcome.

[...]

PR #653, opened 2020-04-23
[Bitbucket] Add bitbucket backend to perceval

imnitishng:
This PR adds Bitbucket backend to Perceval.
Issues and Pull requests from bitbucket repositories can be extracted with this iteration.
Based on the discussions here https://github.com/chaoss/grimoirelab-perceval/issues/367

Signed-off-by: Nitish Gupta <imnitish.ng@gmail.com>

### DIFF
--- a/perceval/backends/core/bitbucket.py
+++ b/perceval/backends/core/bitbucket.py
@@ -1,0 +1,650 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2015-2020 Bitergia
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+# Authors:
+#     Nitish Gupta <imnitish.ng@gmail.com>
+
+import json
+import logging
+
+import requests
+from grimoirelab_toolkit.datetime import (datetime_to_utc,
+                                          datetime_utcnow,
+                                          str_to_datetime)
+from grimoirelab_toolkit.uris import urijoin
+
+from ...backend import (Backend,
+                        BackendCommand,
+                        BackendCommandArgumentParser,
+                        DEFAULT_SEARCH_FIELD)
+from ...client import HttpClient, RateLimitHandler
+from ...utils import DEFAULT_DATETIME, DEFAULT_LAST_DATETIME
+
+CATEGORY_ISSUE = "issue"
+CATEGORY_PULL_REQUEST = "pull_request"
+
+BITBUCKET_URL = "https://bitbucket.org/"
+BITBUCKET_API_URL = "https://api.bitbucket.org/2.0"
+
+# Range before sleeping until rate limit reset
+MIN_RATE_LIMIT = 10
+MAX_RATE_LIMIT = 500
+
+MAX_CATEGORY_ITEMS_PER_PAGE = 100
+# PER_PAGE = 100
+
+# Default sleep time and retries to deal with connection/server problems
+DEFAULT_SLEEP_TIME = 1
+MAX_RETRIES = 5
+
+TARGET_ISSUE_FIELDS = ['reporter', 'assignee', 'links']
+TARGET_PULL_FIELDS = ['author', 'closed_by', 'links']
+TARGET_ACTIVITY_FIELDS = ['update', 'approval']
+
+logger = logging.getLogger(__name__)
+
+
+class Bitbucket(Backend):
+
+    version = '0.25.1'
+
+    CATEGORIES = [CATEGORY_ISSUE, CATEGORY_PULL_REQUEST]
+
+    def __init__(self, owner=None, repository=None,
+                 api_token=None, base_url=None,
+                 tag=None, archive=None,
+                 sleep_for_rate=False, min_rate_to_sleep=MIN_RATE_LIMIT,
+                 max_retries=MAX_RETRIES, sleep_time=DEFAULT_SLEEP_TIME,
+                 max_items=MAX_CATEGORY_ITEMS_PER_PAGE, ssl_verify=True,
+                 client_id=None, secret_id=None, refresh_token=None):
+        if api_token is None:
+            api_token = None
+        origin = base_url if base_url else BITBUCKET_URL
+        origin = urijoin(origin, owner, repository)
+
+        super().__init__(origin, tag=tag, archive=archive, ssl_verify=ssl_verify)
+
+        self.owner = owner
+        self.repository = repository
+        self.api_token = api_token
+        self.base_url = base_url
+
+        self.sleep_for_rate = sleep_for_rate
+        self.min_rate_to_sleep = min_rate_to_sleep
+        self.max_retries = max_retries
+        self.sleep_time = sleep_time
+        self.max_items = max_items
+
+        self.client = None
+        self.exclude_user_data = False
+        self._users = {}  # internal users cache
+
+        self.client_id = client_id
+        self.secret_id = secret_id
+        self.refresh_token = refresh_token
+
+    def search_fields(self, item):
+        """Add search fields to an item.
+
+        It adds the values of `metadata_id` plus the `owner` and `repo`.
+
+        :param item: the item to extract the search fields values
+
+        :returns: a dict of search fields
+        """
+        search_fields = {
+            DEFAULT_SEARCH_FIELD: self.metadata_id(item),
+            'owner': self.owner,
+            'repo': self.repository
+        }
+
+        return search_fields
+
+    def fetch(self, category=CATEGORY_ISSUE, from_date=DEFAULT_DATETIME, to_date=DEFAULT_LAST_DATETIME,
+              filter_classified=False):
+        """Fetch the issues/pull requests from the repository.
+
+        The method retrieves, from a Bitbucket repository, the issues/pull requests
+        updated since the given date.
+
+        :param category: the category of items to fetch
+        :param from_date: obtain issues/pull requests updated since this date
+        :param to_date: obtain issues/pull requests until a specific date (included)
+        :param filter_classified: remove classified fields from the resulting items
+
+        :returns: a generator of issues
+        """
+        self.exclude_user_data = filter_classified
+
+        if self.exclude_user_data:
+            logger.info("Excluding user data. Personal user information won't be collected from the API.")
+
+        if not from_date:
+            from_date = DEFAULT_DATETIME
+        if not to_date:
+            to_date = DEFAULT_LAST_DATETIME
+
+        from_date = datetime_to_utc(from_date)
+        to_date = datetime_to_utc(to_date)
+
+        kwargs = {
+            'from_date': from_date,
+            'to_date': to_date
+        }
+        items = super().fetch(category,
+                              filter_classified=filter_classified,
+                              **kwargs)
+
+        return items
+
+    def fetch_items(self, category, **kwargs):
+        """Fetch the items (issues or pull_requests or repo information)
+
+        :param category: the category of items to fetch
+        :param kwargs: backend arguments
+
+        :returns: a generator of items
+        """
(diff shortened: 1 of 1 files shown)


imnitishng:
Hi @valeriocos, here is the first iteration for the backend.
Bitbucket has a different way of handling authorization and retrieving data from access tokens.

In order to use the backend we need 3 tokens,
- Bitbucket Client ID / Key (obtained from [bitbucket](https://bitbucket.org/account/settings/))
- Bitbucket Secret Key (obtained from [bitbucket](https://bitbucket.org/account/settings/))
- Refresh Token (obtained via CURL instructions given below)

Using these 3 keys, the access token is extracted which is used to make API requests.
Access tokens expire in 2 hours ([ref](https://confluence.atlassian.com/bitbucket/oauth-on-bitbucket-cloud-238027431.html)), hence the process to create a new one is mandatory.

For getting refresh token from bitbucket one needs to follow these steps -

- In your browser go to 
https://bitbucket.org/site/oauth2/authorize?client_id={client_id}&response_type=code
(replace {client_id} with the key from bitbucket)
- Authorize under your bitbucket account.
- After that, your browser will be redirected to 
`{your_redirect_link}/?code={code}`
(save this {code} this is what we need)
- After this we need to CURL filling in our details -
`curl -X POST -u {client_id}:{secret_id} https://bitbucket.org/site/oauth2/token -d grant_type=authorization_code -d code={code}`
(use the above extracted code in place of {code})
- The response will be a JSON containing out refresh token.

valeriocos:
Thank you @imnitishng for the PR and details

> Using these 3 keys, the access token is extracted which is used to make API requests.
Access tokens expire in 2 hours (ref), hence the process to create a new one is mandatory.

This may be a really big problem. Is there any other way to bypass the 2 hours limit?
I see that the reference you posted is about `OAuth on Bitbucket Cloud`, however [Bitbucket server](https://confluence.atlassian.com/bitbucketserver/personal-access-tokens-939515499.html#Personalaccesstokens-Usingpersonalaccesstokens) seems to be less restrictive.

What's the difference between Bitbucket Cloud and Bitbucket server?
Are you targetting a specific version of the Bitbucket API?

Thanks!

vchrombie:
Hi @imnitishng 
Nice work.

I just have a few questions.

I understand that you need 3 tokens to fetch the data.
-  client id
- secret key
- refresh token

> Using these 3 keys, the access token is extracted which is used to make API requests.

Do you mean there will be a new token?

> Access tokens expire in 2 hours ([ref](https://confluence.atlassian.com/bitbucket/oauth-on-bitbucket-cloud-238027431.html)), hence the process to create a new one is mandatory.

Will the 3 tokens gonna expire or only the refresh token?

> For getting refresh token from bitbucket one needs to follow these steps -
> 
> * In your browser go to
>   [https://bitbucket.org/site/oauth2/authorize?client_id={client_id}&response_type=code](https://bitbucket.org/site/oauth2/authorize?client_id=%7Bclient_id%7D&response_type=code)
>   (replace {client_id} with the key from bitbucket)
> * Authorize under your bitbucket account.
> * After that, your browser will be redirected to
>   `{your_redirect_link}/?code={code}`
>   (save this {code} this is what we need)
> * After this we need to CURL filling in our details -
>   `curl -X POST -u {client_id}:{secret_id} https://bitbucket.org/site/oauth2/token -d grant_type=authorization_code -d code={code}`
>   (use the above extracted code in place of {code})
> * The response will be a JSON containing out refresh token.

Is it possible to automate this process?

imnitishng:
Hi @vchrombie, thanks.

> Do you mean there will be a new token?

Yes! It is the access token without it we cannot communicate. No other token can be used for communication.

> Will the 3 tokens gonna expire or only the refresh token?

None of the 3 tokens mentioned will expire.
Only the access token we derive will expire in 2 hours. Which will be renewed on every run.

> Is it possible to automate this process?

I tried to do it, but it is not possible since 

> Authorize under your bitbucket account.

we need to grant a yes permission for authorization in the browser at this step. Just like granting OAuth permission for GitHub. So as far as I know it is only possible manually.

imnitishng:
Here is the part where access token is extracted from all the 3 tokens. Have a look.
[ def _extract_access_token(self, client_id, secret_id, refresh_token)](https://github.com/imnitishng/grimoirelab-perceval/blob/d294562d7f9c9a1b086ccf8f23a03e449c2e9291/perceval/backends/core/bitbucket.py#L447)

vchrombie:
Hi @imnitishng 
Thanks for the detailed reply.

I have a few more confirmations from my side.

1. 
> We already have the 3 necessary tokens and everytime perceval is called we create a new access token to interact with bitbucket. 

> Yes! It is the access token without it we cannot communicate. No other token can be used for communication.

> None of the 3 tokens mentioned will expire.
> Only the access token we derive will expire in 2 hours. Which will be renewed on every run.

So, we should just provide the 3 keys, the access token will be generated every time you execute the backend ([bitbucket.py#L447](https://github.com/imnitishng/grimoirelab-perceval/blob/d294562d7f9c9a1b086ccf8f23a03e449c2e9291/perceval/backends/core/bitbucket.py#L447)) which is valid for 2 hours. So, the next time you are executing the backend, a new access token is generated. So there will be a unique access token for every execution of Perceval backend.

2.
> It is highly unlikely that perceval will run for more that 2 hours. 

Reading this gives me another doubt. I am just curious to know what happens if the fetching takes more than 2 hours?

3.
> > What's the difference between Bitbucket Cloud and Bitbucket server?
> 
> Bitbucket server is another ...

Is it something like gitlab? 
self-hosted like https://invent.kde.org/
and under gitlab like https://gitlab.com/amfoss

4.
> > Is it possible to automate this process?
> 
> we need to grant a yes permission for authorization in the browser at this step. Just like granting OAuth permission for GitHub. So as far as I know it is only possible manually.

Thanks for the clarification. Anyways since it is just for once, I think it is fine. I thought this would be the one expiring. I misunderstood, sorry.
Maybe selenium can help you. Just have a look when you are free.

imnitishng:
> So, we should just provide the 3 keys, the access token will be generated every time you execute the backend (bitbucket.py#L447) which is valid for 2 hours. So, the next time you are executing the backend, a new access token is generated. So there will be a unique access token for every execution of Perceval backend.

Yes, you are correct.

> Reading this gives me another doubt. I am just curious to know what happens if the fetching takes more than 2 hours?

401 response will be returned and Perceval run ends.

> Is it something like gitlab?
self-hosted like https://invent.kde.org/
and under gitlab like https://gitlab.com/amfoss

Well I not sure about that. Out of all the info I've found out about Bitbucket Server, I didn't see anywhere that bitbucket server projects are available on the Web. They are hosted in-house and available for the team of people working only.
Only Bitbucket cloud projects are available for everyone to explore on bitbucket.org.

> Maybe selenium can help you. Just have a look when you are free.

Sure, thanks.

valeriocos:
Please keep in mind that Perceval has been conceived to do one thing only, which is fetching data from software repositories. A backend that needs to autogenerate a token doesn't seem to fit in this schema. Furthermore, a token that expires in 2 hours doesn't seem to have been designed to fetch data from a repository, instead his use is probably meant to address other needs (e.g., pushing commits, merging pull requests, etc.).

Before evaluating the next development tasks, it would be convenient to:
- check token usage policy and restrictions.
- check if the API(s) can be queried using user username/password of the user.
- know the difference between the two APIs (do they have the same endpoints, do they return the same information).
- check how to grant a yes permission from low-level librarires (e.g., requests). Selenium (and other similar tools) doesn't seem a viable option since (i) it adds an additional level of complexity in the code, thus more points of failure, (ii) it may require some tuning in the server where the backend runs.

Thanks!

vchrombie:
> * check how to grant a yes permission from low-level librarires (e.g., requests). Selenium (and other similar tools) doesn't seem a viable option since (i) it adds an additional level of complexity in the code, thus more points of failure, (ii) it may require some tuning in the server where the backend runs.

Okay. I suggested it so that it can be a individual script, not included in perceval. Like a gist just to extract the refresh token.

Other than that.
> https://github.com/chaoss/grimoirelab-perceval/pull/653#issuecomment-618839184

Agreed, thanks.

imnitishng:
Hi @valeriocos I understand your concern here is some more info.

> check token usage policy and restrictions.

Access tokens are generated from the Key, Secret and Refresh tokens.
Here is a snippet of the usage scopes Bitbucket allows for users authenticating using these keys.
Screenshot from 2020-04-24 13-33-07
So they serve both read and write purposes. There are no restrictions for reading repository data.

The final keys obtained will be these (they won't expire) (deleted before posting)
Screenshot from 2020-04-24 13-33-16

For rate limit restrictions, we get 1000 queries per hour for any access to `api.bitbucket.org/2.0/repositories/* ` ([more details here](https://confluence.atlassian.com/bitbucket/rate-limits-668173227.html#Ratelimits-Requestlimits))


> check if the API(s) can be queried using user username/password of the user.

Coming to the authentication problem, Bitbucket allows 3 ways to authenticate users -
**Basic**
Basic authentication can be used, querying bitbucket using username and passwords like
`https://username:password@api.bitbucket.org/2.0/repositories/imnitish/libqxt/issues` works.
So we can use basic authentication if you don't like the idea of extracting access tokens.
A drawback to basic authentication is that it cannot be used on accounts with 2-factor-auth / 2-step-verification.

**API Key**
API keys are supported but meant for only team accounts, so an individual user like us cannot get an API key. 

**OAuth2**
This is the secure way bitbucket reccomends for authentication, and this is the way I implemented the backend, it uses access code flow based on [RFC-6749](https://tools.ietf.org/html/rfc6749).
This picture might make things clear.
Screenshot from 2020-04-24 13-36-40
This is the same thing as authenticating with a username(client ID) and password(secret key) but with an added step of generating access token.

Find more info about authentication with Bitbucket API [here](https://developer.atlassian.com/bitbucket/api/2/reference/security).

> know the difference between the two APIs (do they have the same endpoints, do they return the same information).

They both are different APIs meant for different products offered by Atlassian.

Bitbucket Cloud is the product we are targeting. Bitbucket Cloud is an opensource platform hosting projects on the web accessed via URLs. (Eg.- https://bitbucket.org/libqxt/libqxt/) 
Bitbucket Server APIs cannot be used for Bitbucket Cloud.

Bitbucket Server is a different product, which has self-hosted projects among professional teams, the APIs used for those projects hosted under custom domains like `http://example.com/rest/access-tokens/1.0/users/{userSlug}`, not under bitbucket.org.

I hope we are clear on the point that Bitbucket server and Bitbucket Cloud are 2 different products with 2 completely different APIs, that cannot be interchanged with one another. 
I know this is a point raised here https://github.com/chaoss/grimoirelab-perceval/issues/367#issuecomment-548167687. I believe we can add Bitbucket server support as a different backend named `bitbucketserver` in perceval, independent of this backend.

imnitishng:
We can also remove all this complexity and provide instructions to the user on how to generate the access tokens, but they will have to do it every time they run perceval (if they run it after the 2 hour window).
But Bitbucket access tokens, are pretty large, here is an expired one
```
access_token=
9q1wWP3IHN2IuHp427CCADhsDToG9ArZoHY741OnQLOaFRw1zse-v9mFXfKbOkLwE5_DM5JHZJJ0EPFGUp814GIqOW7PK-9XvEf_QkKEFBLU5kyWfOUZnICT
```
so it might not be a good option to go with.

imnitishng:
Any instructions on this @valeriocos?
If you want then I can submit a new PR with username and password authentication method as you suggested.

vchrombie:
Hi @imnitishng, thanks for the work on the bitbucket backend. Do you think you can move the work to a new repository?

Supporting this https://github.com/chaoss/grimoirelab-perceval/pull/645#issuecomment-610832196, it would be great if you can create a new repository and maintain this backend.

Recently, I moved the Zulip Backend to a new repo. The workflows, documentation, and integrations are everything are in place. You can use it as a template repository and create the project `grimoirelab-perceval-bitbucket`. What do you think.? Please let me know if you need any help.

Zulip Backend for Perceval: https://github.com/vchrombie/grimoirelab-perceval-zulip

vchrombie:
Adhering to the [contributing guidelines (#incubating-repositories)](https://github.com/chaoss/grimoirelab/blob/master/CONTRIBUTING.md#incubating-repositories) and after communicating with @imnitishng, this work is moved to a separate repository [perceval-backends/grimoirelab-perceval-bitbucket](https://github.com/perceval-backends/grimoirelab-perceval-bitbucket) and will be maintained over there for some time. We can open this PR again and update it when we want to merge this backend into the main module.

I'll be opening a PR adding the repo to the readme. The pull request will be closed soon if no one has any objection.

Best,
Venu

vchrombie:
Closing this PR in favour of https://github.com/chaoss/grimoirelab-perceval/pull/653#issuecomment-917369234.

PR #675, opened 2020-06-19
[slack] Handle messages without user/bot ID

valeriocos:
The user/bot ID is used by the backend to create the the Perceval document ID. In the scenario where this information is missing (e.g., bot deleted), the document ID cannot be generated and error is thrown in the method `metadata_id`.

This code proposes to set the username as fallback value for the messages without user/bot ID, thus allowing the creation of the Perceval document ID.

Backend version is 0.9.2.

### DIFF
--- a/perceval/backends/core/slack.py
+++ b/perceval/backends/core/slack.py
@@ -61,7 +61,7 @@ class Slack(Backend):
     :param archive: archive to store/retrieve items
     :param ssl_verify: enable/disable SSL verification
     """
-    version = '0.9.1'
+    version = '0.9.2'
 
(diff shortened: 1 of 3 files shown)


perceval/backends/core/slack.py
```diff
@@ -188,8 +188,10 @@ def metadata_id(item):
             nick = item['user']
         elif 'comment' in item:
             nick = item['comment']['user']
-        else:
+        elif 'bot_id' in item:
             nick = item['bot_id']
+        else:
+            nick = 'not_found'
```
sduenas, line 194: Shouldn't be better to use the field `username` that a string like this?

PR #677, opened 2020-07-14
[slack] Use channels.history to retrieve message

valeriocos:
This change is needed since the new endpoint `conversations.history` doesn't include replies among the messages fetched from the API. There isn't enough documentation to understand whether this exclusion is a bug. The document that contains
deprecation info only mentions that `channels.history` is replaced by `conversations.history`: https://api.slack.com/changelog/2020-01-deprecating-antecedents-to-the-conversations-api

As found on the Web, it seems that the endpoint `conversations.history` stopped returning replies since 2020-04-17: https://www.bakejam.com/slacks-conversationshistory-api-ignores-replies/

The `channels.history` endpoint will work until Feb 2021 and won't work with apps created after June 10th 2020 (ref: https://api.slack.com/methods/channels.history).

### DIFF
--- a/perceval/backends/core/slack.py
+++ b/perceval/backends/core/slack.py
@@ -61,7 +61,7 @@ class Slack(Backend):
     :param archive: archive to store/retrieve items
     :param ssl_verify: enable/disable SSL verification
     """
-    version = '0.9.2'
+    version = '0.9.3'
 
     CATEGORIES = [CATEGORY_MESSAGE]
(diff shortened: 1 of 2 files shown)


valeriocos:
@sduenas are you going to merge the PR?

sduenas:
Done. Sorry, I forgot about it.

valeriocos:
no worries

PR #682, opened 2020-07-27
Add support for specifying Gerrit SSH key location, fixes #681

### LINKED ISSUE #681: [Question] Is is possible to run multiple perceval gerrits (with different SSH keys) at the same time?
Hi, here is the problem: I want to run say 10 Gerrit backends in parallel, each of them uses different SSH private keys (`~/.ssh/id_rsa`), is it possible to run them in parallel instead of in sequence (so I can switch keys before each one starts)?
Is there any parameter that can specify a non-default private key location?

lukaszgryglicki:
This adds `--ssh-id-filepath` Perceval Gerrit backend option. You can use it to allow running multiple Gerrit backends in parallel, each having its own SSH private key.

Fixes #681 

Signed-off-by: Łukasz Gryglicki <lukaszgryglicki@o2.pl>

### DIFF
--- a/perceval/backends/core/gerrit.py
+++ b/perceval/backends/core/gerrit.py
@@ -62,6 +62,7 @@ class Gerrit(Backend):
     :param tag: label used to mark the data
     :param archive: archive to store/retrieve items
     :param blacklist_ids: exclude the reviews while fetching
+    :param id_filepath: path to SSH private key
     """
     version = '0.13.1'
 
@@ -74,7 +75,7 @@ class Gerrit(Backend):
 
     def __init__(self, hostname,
                  user=None, port=PORT, max_reviews=MAX_REVIEWS,
-                 disable_host_key_check=False,
+                 disable_host_key_check=False, id_filepath=None,
                  tag=None, archive=None, blacklist_ids=None):
         origin = hostname
 
@@ -82,6 +83,7 @@ class Gerrit(Backend):
         self.hostname = hostname
         self.user = user
         self.port = port
+        self.id_filepath = id_filepath
         self.max_reviews = max(1, max_reviews)
         self.blacklist_ids = blacklist_ids
         self.disable_host_key_check = disable_host_key_check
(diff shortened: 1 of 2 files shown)


lukaszgryglicki:
Solution suggested by @valeriocos [here](https://github.com/chaoss/grimoirelab-perceval/issues/681#issuecomment-663406344).

perceval/backends/core/gerrit.py
```diff
@@ -520,6 +526,7 @@ def setup_cmd_parser(cls):
         group.add_argument('--ssh-port', dest='port',
                            default=PORT, type=int,
                            help="Set SSH port of the Gerrit server")
+        group.add_argument('--ssh-path', dest='path', help="Set SSH path")
```
sduenas, line 529: The name of this argument and variables around the code should be different. The SSH parameter uses the SSH identity file, so I think it will be better something like `ssh-id-filepath`. What do you think?

lukaszgryglicki: OK, will address tomorrow.

lukaszgryglicki: Addressed, tests added, not squashed yet, I'll squash when CI tests pass.

lukaszgryglicki:
@sduenas commits squashed, feedback addressed.

lukaszgryglicki:
@sduenas ping.

lukaszgryglicki:
@sduenas feedback addressed.

perceval/backends/core/gerrit.py
```diff
@@ -296,6 +299,7 @@ class GerritClient():
     :param blacklist_reviews: exclude the reviews of this list while fetching
     :param disable_host_key_check: disable host key controls
     :param port: SSH port
+    :param id_filepath: SSH private key patch
```
sduenas, line 302: There's a typo in `patch`.

lukaszgryglicki: @sduenas fixed.

lukaszgryglicki:
@sduenas fixed.

lukaszgryglicki:
Thanks!!

sduenas:
You are welcome. Thanks for the patch.

Issue #687, opened 2020-09-02
Current 'dulwich' version doesn't work with Python 3.8


jjmerchante:
Using the version of `dulwich` specified in the requirements.txt and running Perceval with Python 3.8 raises the following error:

```
bin/perceval git --latest-items  https://github.com/chaoss/grimoirelab-perceval.git
[2020-09-02 14:34:19,263] - Sir Perceval is on his quest.
[2020-09-02 14:34:19,264] - Fetching latest commits: 'https://github.com/chaoss/grimoirelab-perceval.git' git repository
Traceback (most recent call last):
  File "/home/jjmerchante/.pyenv/versions/perceval/lib/python3.8/site-packages/perceval/backend.py", line 618, in run
    for item in big.items:
  File "/home/jjmerchante/.pyenv/versions/perceval/lib/python3.8/site-packages/perceval/backend.py", line 794, in __fetch
    raise e
  File "/home/jjmerchante/.pyenv/versions/perceval/lib/python3.8/site-packages/perceval/backend.py", line 788, in __fetch
    for item in items:
  File "/home/jjmerchante/.pyenv/versions/perceval/lib/python3.8/site-packages/perceval/backend.py", line 226, in fetch
    for item in self.fetch_items(category, **kwargs):
  File "/home/jjmerchante/.pyenv/versions/perceval/lib/python3.8/site-packages/perceval/backends/core/git.py", line 160, in fetch_items
    commits = self.__fetch_from_repo(from_date, to_date, branches,
  File "/home/jjmerchante/.pyenv/versions/perceval/lib/python3.8/site-packages/perceval/backends/core/git.py", line 279, in __fetch_from_repo
    commits = self.__fetch_newest_commits_from_repo(repo)
  File "/home/jjmerchante/.pyenv/versions/perceval/lib/python3.8/site-packages/perceval/backends/core/git.py", line 316, in __fetch_newest_commits_from_repo
    hashes = repo.sync()
  File "/home/jjmerchante/.pyenv/versions/perceval/lib/python3.8/site-packages/perceval/backends/core/git.py", line 937, in sync
    pack_name, refs = self._fetch_pack()
  File "/home/jjmerchante/.pyenv/versions/perceval/lib/python3.8/site-packages/perceval/backends/core/git.py", line 1105, in _fetch_pack
    client, repo_path = dulwich.client.get_transport_and_path(self.uri)
[...]

PR #688, opened 2020-09-02
Update dulwich package version

jjmerchante:
Running Perceval with Python 3.8 and dulwich <0.19.6 raises an Exception because a method has been deprecated and removed.

This commit updates the version of `dulwich` to the latest one

Related to: https://github.com/chaoss/grimoirelab-perceval/issues/687

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ Run 'perceval <backend> --help' to get information about a specific backend.
 * python3-requests >= 2.7
 * python3-bs4 (beautifulsoup4) >= 4.3
 * python3-feedparser >= 5.1.3
-* python3-dulwich >= 0.18.5
+* python3-dulwich >= 0.20.0
 * grimoirelab-toolkit >= 0.1.4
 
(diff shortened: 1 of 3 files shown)


jjmerchante:
I don't understand why it fails with Python 3.6.7, I run the same commands with the same version on my local machine and installed Perceval successfully.

valeriocos:
Hi @jjmerchante ,  I guess the problem is due to the installation of toolkit (https://github.com/chaoss/grimoirelab-toolkit/issues/28). I saw something similar with ELK.

Hope this helps

sduenas:
It fails in Python 3.5 due to what @valeriocos comments. The failure in 3.6 should be related to this https://github.com/pypa/setuptools/issues/2355.

sduenas:
It works now. There's a new version of **setuptools** that I think that solves the problem. The tests are still failing for 3.5 but it's ok. I think we will drop the support of 3.5 very soon because the end of life is scheduled on 2020-09-13.

PR #694, opened 2020-10-04
[perceval] Add flag `-l/--list` to show list of available backends

### LINKED ISSUE #690: Add option to show the list of available backends dynamically
The command-line client should display the list of available backends. The current list is only shown with `perceval --help` but this list is static.

The idea is to have a flag (`-l / --list`, or similar) to display the list. So for instance, when the user runs the next command it will display the list shown on this example:

```
$ perceval --list

bugzilla
git
github
gitlab
[...]
jira
launchpad
mattermost
```
You can make use of the call to `find_backends` function. This function is called [here](https://github.com/chaoss/grimoirelab-perceval/blob/master/bin/perceval#L104)

imnitishng:
This commit adds `-l` optional argument to show dynamic list of
backends currently available in perceval.

I did some minor changes with the function call [here](https://github.com/chaoss/grimoirelab-perceval/blob/master/bin/perceval#L104) to avoid redundant calls. 

Closes #690

### DIFF
--- a/bin/perceval
+++ b/bin/perceval
@@ -27,6 +27,7 @@
 #     MalloZup <dmaiocchi@suse.com>
 #     Venu Vardhan Reddy Tekula <venuvardhanreddytekula8@gmail.com>
 #     animesh <animuz111@gmail.com>
+#     Nitish Gupta <imnitish.ng@gmail.com>
 #
 
 import argparse
@@ -38,7 +39,7 @@ import perceval.backend
 import perceval.backends.core
 
 PERCEVAL_USAGE_MSG = \
-"""%(prog)s [-g] <backend> [<args>] | --help | --version"""
+"""%(prog)s [-g] <backend> [<args>] | --help | --version | --list"""
 
 PERCEVAL_DESC_MSG = \
 """Send Sir Perceval on a quest to retrieve and gather data from software
@@ -84,6 +85,7 @@ optional arguments:
   -h, --help            show this help message and exit
   -v, --version         show version
   -g, --debug           set debug mode on
+  -l, --list            show available backends
 """
 
 PERCEVAL_EPILOG_MSG = \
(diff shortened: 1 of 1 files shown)


bin/perceval
```diff
@@ -84,6 +85,7 @@ optional arguments:
   -h, --help            show this help message and exit
   -v, --version         show version
   -g, --debug           set debug mode on
+  -l, --list            show available backends
```
sduenas, line 88: Add `--list` option to the line 42 too.

bin/perceval
```diff
@@ -143,6 +147,12 @@ def parse_args():
         parser.print_help()
         sys.exit(1)
 
+    if sys.argv[-1] == '--list' or sys.argv[-1] == '-l':
```
sduenas, line 150: If you run perceval with this flag and an argument you get an error:
```
$ perceval --list test
  File "/home/sduenas/devel/venv/bin/perceval", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
  File "/home/sduenas/devel/grimoirelab/perceval/bin/perceval", line 179, in <module>
    main()
  File "/home/sduenas/devel/grimoirelab/perceval/bin/perceval", line 109, in main
    raise RuntimeError("Unknown backend %s" % args.backend)
RuntimeError: Unknown backend test
```

You should move this code into an `Action` class. You have an example on this [stackoverflow question](https://stackoverflow.com/questions/45060060/argparse-argument-to-print-and-exit-disregarding-positional-arguments). Don't forget to add a `parser.exit()` call to the `__call__` method.

sduenas (changes requested):
Thanks for your PR. Can you address the comments of this review?

bin/perceval
```diff
@@ -117,7 +133,7 @@ def main():
     logging.info("Sir Perceval completed his quest.")
 
 
-def parse_args():
+def parse_args(PERCEVAL_CMDS):
```
sduenas, line 136: Can you rename this to `perceval_cmds`?

sduenas (changes requested):
LGTM but there's a minor change you need to fix first.

sduenas (approved):
LGTM

sduenas:
@imnitishng thanks for the contribution!

PR #698, opened 2020-10-22
Gitlab Client: Parse owner and repository in a url safe way

StingRayZA:
Instead of stitching the 'owner' and 'repository' together with a `%2F` - rather join them with a `/` and let python make them conformant.
This way, if we have any `/` characters embedded in either value, they too will be correctly translated


sduenas:
Thanks for the PR. I need some time to review it. I'll do it this weekend.

StingRayZA:
Hi 👋 
Could I request you label this PR with `hacktoberfest-accepted` ?

StingRayZA:
@sduenas Rebased on master with a reformatted message. 
It also links to chaoss/grimoirelab-elk#946 which is where I started thinking about this issue. I think my hack in that issue is a bit naive, but together with this one - my local copy of grimoirelab is working well now.

sduenas (approved):
LGTM

As I said, I'll merge it when we fix the related problems in ELK.

PR #699, opened 2020-10-23
[tests] Fix pipermail not working search_fields test

sduenas:
The test 'test_search_field'  was not really working. It was trying to access 'example.com' URL without mocking the connection with
'httpretty'. Also, the test was wrong because 'metadata_id' is not a method of 'PipermailList'.


zhquan (approved):
LGTM

PR #700, opened 2020-10-23
[githubql] Fix error `Events not collected for issue...`

zhquan:
This code fixes the error `Events not collected for issue ...'`,
which is caused when fetching MergedEvent on issues because
this event type is only available for pull requests.

Tests updated accordingly.


sduenas (commented):
LGTM

PR #703, opened 2020-11-13
[slack] Fix fetch message from a private user

zhquan:
This code allows fetching message from a private user and set
the `user` data as `None`.


sduenas (changes requested):
It looks ok but we need a test to check this case.

sduenas (changes requested):
@zhquan LGTM but would you mind to change the backend version to `version = '0.9.4'`.

sduenas (approved):
LGTM

Issue #709, opened 2021-01-12
Supybot Backend


Nolski:
# Overview

When running `perceval supybot` I receive a parse error saying a date is expected on line one.

## Further details

The exact output of the command is below:

```bash
$ perceval supybot ritfoss ./supybot/logs/ChannelLogger/freenode/#rit-foss/
[2021-01-12 11:58:22,013] - Sir Perceval is on his quest.
[2021-01-12 11:58:22,016] - Fetching messages of 'ritfoss' from 1970-01-01 00:00:00+00:00
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/dist-packages/perceval/backends/core/supybot.py", line 207, in parse_supybot_log
    for message in parser.parse():
  File "/usr/local/lib/python3.7/dist-packages/perceval/backends/core/supybot.py", line 366, in parse
    ts, msg = self._parse_supybot_timestamp(line)
  File "/usr/local/lib/python3.7/dist-packages/perceval/backends/core/supybot.py", line 387, in _parse_supybot_timestamp
    raise ParseError(cause=msg)
perceval.errors.ParseError: date expected on line 1

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.7/dist-packages/perceval/backend.py", line 618, in run
    for item in big.items:
  File "/usr/local/lib/python3.7/dist-packages/perceval/backend.py", line 794, in __fetch
    raise e
  File "/usr/local/lib/python3.7/dist-packages/perceval/backend.py", line 788, in __fetch
    for item in items:
  File "/usr/local/lib/python3.7/dist-packages/perceval/backend.py", line 226, in fetch
    for item in self.fetch_items(category, **kwargs):
  File "/usr/local/lib/python3.7/dist-packages/perceval/backends/core/supybot.py", line 118, in fetch_items
    for message in self.parse_supybot_log(archive):
  File "/usr/local/lib/python3.7/dist-packages/perceval/backends/core/supybot.py", line 211, in parse_supybot_log
    raise ParseError(cause=cause)
perceval.errors.ParseError: file: ./supybot/logs/ChannelLogger/freenode/#rit-foss/#rit-foss.log; reason: date expected on line 1

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/bin/perceval", line 169, in <module>
    main()
  File "/usr/local/bin/perceval", line 115, in main
    cmd.run()
[...]

PR #712, opened 2021-01-14
Supybot now supports logs without timezone information

gregoriorobles:
Supybot now supports logs without timezone information

Supybot is deprecated and the logs generated by its successor,
Limoria, do not contain timezone information.

Test failing when no timezone information was given has been modified
to pass now.
An additional log file based on the issue report #709 has been added
to the test suite.

Resolves: #709


gregoriorobles:
The good one is #713

PR #713, opened 2021-01-14
Supybot now supports logs without timezone information

gregoriorobles:
Signed-off-by: Gregorio <grex@gsyc.urjc.es>

Supybot now supports logs without timezone information

Supybot is deprecated and the logs generated by its successor,
Limoria, do not contain timezone information.

Test failing when no timezone information was given has been modified
to pass now.
An additional log file based on the issue report #709 has been added
to the test suite.

Resolves: #709

### DIFF
--- a/perceval/backends/core/supybot.py
+++ b/perceval/backends/core/supybot.py
@@ -314,7 +314,7 @@ class SupybotParser:
 
     :param stream: an iterator which produces Supybot log lines
     """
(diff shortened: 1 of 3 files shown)


sduenas (changes requested):
The patch looks great but I need two minor changes before accept it:

- Update the `version` of the backend to `0.10.0` (line 67, file supybot.py).
- Add the message you set in this PR as the body of the commit. Also the first line of the message should be something like `[supybot] Add support for logs without timezone information`.

gregoriorobles:
Solved with pull request #716

PR #714, opened 2021-01-21
Update setup.py

vchrombie:
This commit updates the setup.py with the latest dependencies required for perceval.
It is necessary as it is responsible for failures in the builds of other repos.


sduenas (approved):
LGTM

PR #718, opened 2021-02-12
[stackexchange] Change the fetch order to 'asc'

zhquan:
This change is required to retrieve items incrementally.

Backend version is 0.12.1.


sduenas (approved):
LGTM

PR #731, opened 2021-04-21
[gha] Create release.yml

vchrombie:
This commit adds the release github action which generates a package, tests it with different python versions, publishes the release on GitHub, and uploads the packages to PyPI whenever a tag is generated and pushed.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,103 @@
+name: release
+
+on:
+  push:
+    tags:
+      - '*.*.*'
+      - '*.*.*-*'
+
+jobs:
+
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.x'
+      - name: Install poetry
+        run: |
+          curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/install-poetry.py | python -
+          echo "PATH=$HOME/.poetry/bin:$PATH" >> $GITHUB_ENV
+      - name: Build distributions
+        run: |
+          poetry build
+      - name: Upload distribution artifacts
+        uses: actions/upload-artifact@v2
+        with:
+          name: perceval-dist
+          path: dist
+
+  tests:
+    needs: [build]
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.6, 3.7, 3.8]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Download distribution artifact
+        uses: actions/download-artifact@v2
+        with:
+          name: perceval-dist
+          path: dist
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dev dependencies
+        run: |
+          pip install peodd
+          peodd -o requirements-dev.txt
+          pip install -r requirements-dev.txt
+      - name: Test package
+        run: |
+          PACKAGE_NAME=`(cd dist && ls *whl | cut -f 1 -d "-")` && echo $PACKAGE_NAME
+          pip install --pre --find-links ./dist/ $PACKAGE_NAME
+          cd tests && python run_tests.py
+
+  release:
+    needs: [tests]
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Get release tag
+        id: tag
+        run: |
+          echo ::set-output name=tag::${GITHUB_REF#refs/tags/}
+      - name: Create release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: ${{ steps.tag.outputs.tag }}
+          release_name: ${{ steps.tag.outputs.tag }}
+          body_path: ./releases/${{ steps.tag.outputs.tag }}.md
+
+  publish:
+    needs: [tests]
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Download distribution artifact
+        uses: actions/download-artifact@v2
+        with:
+          name: perceval-dist
+          path: dist
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.x'
+      - name: Install poetry
+        run: |
+          curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/install-poetry.py | python -
+          echo "PATH=$HOME/.poetry/bin:$PATH" >> $GITHUB_ENV
+      - name: Configure pypi credentials
+        env:
+          PYPI_API_TOKEN: ${{ secrets.PYPI_API_TOKEN }}
+        run: |
+          poetry config http-basic.pypi __token__ "$PYPI_API_TOKEN"
+      - name: Publish release to pypi
+        run: |
+          poetry publish


.github/workflows/release.yml
```diff
@@ -0,0 +1,100 @@
...
+        with:
+          name: perceval-dist
+          path: dist
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dev dependencies
+        run: |
+          pip install httpretty>=0.9.6
```
sduenas, line 50: Didn't you find any workaround for this?

vchrombie: I couldn't install only dev dependencies since the feature is not yet released in poetry.

I can use `pip install requirements_tests.txt` but we are looking forward to removing them soon, so I didn't consider the option.
WDYT?

sduenas: I'm not sure if we should keep this PR open until we fix the too problems we have with poetry now:

- Install scripts
- Install dev dependencies only

For the first, there's a PR open that fixes the problem but it isn't integrated with the core of the code yet. For the second one, the fix was implemented but it's not available yet in any of the releases of poetry. Am I right @vchrombie ? Maybe you can complete this with the links to the PRs.

What do you think? Should we wait and see, or mege this?

vchrombie: Yes, you are right @sduenas.

>     * Install scripts

There is an existing PR that could fix this issue https://github.com/python-poetry/poetry-core/pull/40. I see some active work going on here. So, I'm expecting this to completed soon and we could make use of this feature.

>     * Install dev dependencies only

This request was proposed long back and the PR https://github.com/python-poetry/poetry/pull/2917 was merged too, but it is not available for any release till now. I'm not sure when it would be available to use.

> What do you think? Should we wait and see, or merge this?

We are more concerned with the first problem as we cannot use the `perceval` command if we install the tool using poetry. I think once the feature PR is merged, I can open a new PR by adding the configuration.

We currently have a workaround for the second problem (installing the dev dependencies manually) as we don't have another option.

I think we should go ahead and merge this now. We can update the workflows once we have these features in the poetry releases.

WDYT?

sduenas: Ok to me. What should we do with the perceval script then?

vchrombie: > Ok to me. What should we do with the perceval script then?

If we want to generate a release with this workflow, it can cause problems for the users if they are using the pip packages. This is not safe and not recommended.

This means we should not generate the tags until we get the perceval script working with poetry.
Can we generate the releases without the tags (if we merge this PR)? If this is not possible, then we shouldn't merge this at the moment and leave it till we fix the script issue.

@sduenas, do you have any thoughts?

vchrombie:
I think we should not merge this and wait till we fix the issues with the perceval script at least. We can test workflow once everything is ready.

vchrombie:
The PR was blocked because of two main reasons as mentioned in the https://github.com/chaoss/grimoirelab-perceval/pull/731#discussion_r621079285.

>     * Install scripts

We have decided to change the script as an entry point as the script feature in poetry is not yet released and might take some more time. https://github.com/chaoss/grimoirelab-perceval/pull/741

>     * Install dev dependencies only

I wrote a small tool [vchrombie/peodd](https://github.com/vchrombie/peodd) that exports the dev-dependencies from `pyproject.toml` file to `requirements.txt` format. We can make use of this until poetry has the official feature of exporting only the dev-dependencies.

@sduenas, please review this when you are free and let me know your thoughts on the PR.

sduenas:
> The PR was blocked because of two main reasons as mentioned in the [#731 (comment)](https://github.com/chaoss/grimoirelab-perceval/pull/731#discussion_r621079285).
> 
> > ```
> > * Install scripts
> > ```
> 
> We have decided to change the script as an entry point as the script feature in poetry is not yet released and might take some more time. #741

This is already merged.

> 
> > ```
> > * Install dev dependencies only
> > ```
> 
> I wrote a small tool [vchrombie/peodd](https://github.com/vchrombie/peodd) that exports the dev-dependencies from `pyproject.toml` file to `requirements.txt` format. We can make use of this until poetry has the official feature of exporting only the dev-dependencies.
> 
> @sduenas, please review this when you are free and let me know your thoughts on the PR.

I think I didn't understand this part. I don't get why we need this when we create a package and we test it. The package should run with the dependencies that are already in the pyproject file. Can you explain this better?

vchrombie:
Hello @sduenas 

> This is already merged.

Yes, this issue is now solved.

> I think I didn't understand this part. I don't get why we need this when we create a package and we test it. The package should run with the dependencies that are already in the pyproject file. Can you explain this better?

When we build the package using poetry the dev-dependencies are ignored. I think this how it was designed to be. The dev-dependencies don't really matter to the users and it makes sense to me.

```
$ poetry build
Building perceval (0.17.4)
  - Building sdist
  - Built perceval-0.17.4.tar.gz
  - Building wheel
  - Built perceval-0.17.4-py3-none-any.whl

$ PACKAGE_NAME=`(cd dist && ls *whl | cut -f 1 -d "-")` && echo $PACKAGE_NAME
perceval

$ pip install --pre --find-links ./dist/ $PACKAGE_NAME
Looking in links: ./dist/
Collecting perceval
  Using cached perceval-0.17.7-py3-none-any.whl (211 kB)
Collecting urllib3>=1.22
...
Installing collected packages: urllib3, idna, certifi, chardet, requests, six, python-dateutil, grimoirelab-toolkit, PyJWT, pycparser, cffi, cryptography, soupsieve, beautifulsoup4, sgmllib3k, feedparser, dulwich, perceval
Successfully installed PyJWT-2.1.0 beautifulsoup4-4.9.3 certifi-2021.5.30 cffi-1.14.6 chardet-4.0.0 cryptography-3.4.7 dulwich-0.20.23 feedparser-6.0.8 grimoirelab-toolkit-0.2.0 idna-2.10 perceval-0.17.7 pycparser-2.20 python-dateutil-2.8.1 requests-2.25.1 sgmllib3k-1.0.0 six-1.16.0 soupsieve-2.2.1 urllib3-1.26.6
```

grimoirelab-perceval/dist/perceval-0.17.4/setup.cfg
```
install_requires = \
['PyJWT>=1.7.1,<2.0.0',
 'beautifulsoup4>=4.3.2,<5.0.0',
 'cryptography>=3.3.1,<4.0.0',
 'dulwich>=0.20.0,<0.21.0',
 'feedparser>=5.1.3,<6.0.0',
 'grimoirelab-toolkit>=0.1.12,<0.2.0',
 'python-dateutil>=2.6.0,<3.0.0',
 'requests>=2.7.0,<3.0.0',
 'urllib3>=1.22,<2.0']
```

So, now we install the package and there will be a lot of errors while testing as the dependencies which are required to run the tests are not installed.
```
======================================================================
ERROR: test_twitter (unittest.loader._FailedTest)
ImportError: Failed to import test module: test_twitter
  File "/usr/lib/python3.8/unittest/loader.py", line 436, in _find_test_path
    module = self._get_module_from_name(name)
  File "/usr/lib/python3.8/unittest/loader.py", line 377, in _get_module_from_name
    __import__(name)
  File "/home/p0tt3r/dev/sources/grimoirelab-perceval/tests/test_twitter.py", line 33, in <module>
    import httpretty
ModuleNotFoundError: No module named 'httpretty'


Ran 296 tests in 33.036s

FAILED (errors=31)
```

Poetry cannot explicitly install only the dev-dependencies. There was discussion around it and a couple of PRs are merged too but I think there is a lot of work still pending. Reference: https://github.com/python-poetry/poetry/issues/2572

Another workaround could be exporting the dependencies. Again, Poetry cannot explicitly export only the dev-dependencies. There was a discussion around this too, but not sure if they would be interested to add this feature in the poetry export plugin. Reference: https://github.com/python-poetry/poetry/issues/1441

So, to solve this issue, I wrote a small tool that exports the dev-dependencies to a `requirements.txt format` file. As the purpose is only for the CI, we can avoid the burden of maintaining the requirements in different files.

Pleas let me know your thoughts.

sduenas:
I understand what you mean but for the tests when the package is built the dev-dependencies / test-dependencies should be the ones that are in the pyproject.yml. We will install those dependencies from that file in the repository.

For the CI tests - when someone submits a PR or something is merged in a branch - the dev-dependencies should the ones in `poetry` + the dependencies of the other grimoirelab components in git.

What I was saying is to have the latest in a separate `requirements.txt` file. For the CI, first, we will install all the dependencies in the `pyproject.yml` with `poetry`, and then, all of those in that requirements file with `pip`, probably. The requirements file can be named `grimoirelab-dev-depencencies.txt`.

For the build workflow, we will create the package, install it, and install the dev dependencies from the `pyproject` file.

Does it make sense to you?

.github/workflows/release.yml
```diff
@@ -0,0 +1,102 @@
+name: release
+
+on:
+  push:
+    tags: 
+      - '*.*.*'
```
sduenas, line 6: I think we should change this because we might want to have releases with version numbers such as: `x.y.z` or `x.y.z-something`.
The rest of the PR looks ok to me.

vchrombie: @sduenas, I have updated the PR. Please let me know what do you think.

sduenas (approved):
LGTM

vchrombie:
Thanks @sduenas for merging the PR.

> I wrote a small tool [vchrombie/peodd](https://github.com/vchrombie/peodd) that exports the dev-dependencies from `pyproject.toml` file to `requirements.txt` format. We can make use of this until poetry has the official feature of exporting only the dev-dependencies.

We would be using this untill we have some workaround with poetry.

PR #732, opened 2021-04-21
[gha] Update tests workflow

vchrombie:
This commit renames the ci github actions workflow to avoid ambiguity. It also adds a rule to restrict the tests workflow to run when a release tag is pushed to the repository to avoid redundant workflows.


vchrombie:
@sduenas, please review this when you have time.

.github/workflows/tests.yml
```diff
@@ -1,6 +1,14 @@
-name: build
+name: tests
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - '**'
+    tags:
+      - '!*.*.*'
```
sduenas, line 8: Shouldn't be `'!**'`?

vchrombie: It can be that too.
`'!**'` restricts all the tags
`'!*.*.*'` restricts the tags with the pattern `n.n.n`

Do you want me to update the PR with `'!**'`?

sduenas: Yes, please. I think it makes sense to restrict all tags.

vchrombie: Okay

vchrombie:
Updated the PR :+1:

sduenas (approved):
LGTM

Issue #733, opened 2021-04-21
Fail to retrieve user_orgs for organizations with OAuth App access restrictions


jjmerchante:
When retrieving issues from a GitHub repository, it also retrieves information related with users. I found an special issue which creator is an organization: `https://github.com/redis/redis/issues/1531`.

The problem is that the organization (which created the issue) has [OAuth App access restrictions](https://docs.github.com/en/organizations/restricting-access-to-your-organizations-data) to the organization's data.

When you are using Perceval with OAuth tokens, it fails when trying to retrieve information related with the user (organization) because GitHub API returns the following error (403 Unauthorized):
```
curl -H "Authorization: token xxxx" https://api.github.com/users/bobasoba/orgs
{
  "message": "Although you appear to have the correct authorization credentials, the `bobasoba` organization has enabled OAuth App access restrictions, meaning that data access to third-parties is limited. For more information on these restrictions, including how to enable this app, visit https://docs.github.com/articles/restricting-access-to-your-organization-s-data/",
  "documentation_url": "https://docs.github.com/rest/reference/orgs#list-organizations-for-a-user"
}
```
Perceval ouput:
```
...
requests.exceptions.RetryError: HTTPSConnectionPool(host='api.github.com', port=443): Max retries exceeded with url: /users/bobasoba/orgs (Caused by ResponseError('too many 403 error responses',))
[2021-04-21 15:43:38,820 - root - INFO] - Sir Perceval completed his quest.

```
Using personal access tokens, Perceval works as expect.

It would be nice if it continues collecting issues even though it cannot collect some organizations. I will create a PR with a proposal.

PR #734, opened 2021-04-22
[github] Fix user_orgs for organizations with restricted access

jjmerchante:
Return an empty list when requesting a list of orgs of a login and GitHub API returns a 403 Unauthorized.

I wasn't sure why a 403 error code should force a retry, I removed it from the list of `EXTRA_STATUS_FORCELIST`

Fixes: #733

### DIFF
--- a/perceval/backends/core/github.py
+++ b/perceval/backends/core/github.py
@@ -859,6 +859,9 @@ class GitHubClient(HttpClient, RateLimitHandler):
                 orgs = '[]'
             else:
                 raise error
+        except requests.exceptions.RetryError as error:
+            logger.error("Can't get github login orgs: %s", error)
+            orgs = '[]'
 
         self._users_orgs[login] = orgs
 
--- a/tests/test_github.py
+++ b/tests/test_github.py
@@ -2705,6 +2705,18 @@ class TestGitHubBackend(unittest.TestCase):
         github = GitHub("zhquan_example", "repo", ["aaa"])
         _ = [issues for issues in github.fetch()]
 
+        # Check that a RetryError getting user orgs is managed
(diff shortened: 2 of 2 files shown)


sduenas (commented):
Question, what happens if the 403 is returned but on any other call?

jjmerchante:
> Question, what happens if the 403 is returned but on any other call?

I thought that GitHub only returns a 403 when a user is unauthorized, so retrying a 403 would end in RetryLimit.  According to @zhquan, GitHub sometimes returns a 403 when there are network errors and when Perceval retries, it returns a 200, so this implementation is not valid.

I changed the implementation with the other option we have: catch a `RetryLimit` when Perceval requests user orgs. 
The problem here is that I can't get the status code of a `RetryLimit` exception, so I have to catch all the codes. The worst that could happen is that the organizations for a user are always empty (but also Perceval tried to get it 5 times).

sduenas:
> > Question, what happens if the 403 is returned but on any other call?
> 
> I thought that GitHub only returns a 403 when a user is unauthorized, so retrying to get the same resource should end in a `RetryLimit`. According to @zhquan, GitHub sometimes returns 403 when there are network errors, and when Perceval retries, it returns a 200, so this implementation is not valid.
> 
> I changed the implementation with the other option we have: catch a `RetryLimit` when Perceval requests user orgs.
> The problem here is that I can't get the status code of a `RetryLimit` exception, so I have to catch all the codes. The worst that could happen is that the organizations for a user are always empty (but also Perceval tried to get them 5 times).

Can you upload that code so I can see it?

jjmerchante:
Yes, I guess I forgot to push, sorry

sduenas (approved):
LGTM

PR #738, opened 2021-06-18
[docs] Expand docs for implementing new backends

Alch-Emi:
Using the current documentation, it can be pretty difficult to learn how to appropriately implement a new backend for perceval.  Many of the methods that backends are supposed to implement have no documentation explaining what the methods are supposed to do, or even how their arguments are formatted.

Although it is possible to reverse engineer this information from other backends, this process is time consuming, frustrating, and may deter new developers from being able to use perceval with their application.

This commit expands documentation relating to implementing new backends in several ways:

- The class-level documentation for backends has been expanded to list more of the necessary methods and attributes that backends are responsible for implementing, and additionally lists a brief summary of each method.
- Several methods which were previously not listed (_init_client), or listed as optional (metadata_*) have been listed as mandatory.  As far as I can tell, there is no way to create a working backend without implementing these methods, or at least stubbing them.
- Information about class attributes in the class-level documentation has been replaced with a brief summary and a link, with the full text being moved to the attribute itself, making the class-level documentation easier to digest.
- Some of the documentation for class attributes that was sparse or missing altogether has been added
- All* methods and class attributes that are intended to be implemented by subclasses of `Backend` have detailed documentation describing exactly what is expected of the implementor, and giving context for some of the terms or argument names used where necessary.
- Some references to other methods and class attributes in the docs have been converted to links, to make navigating the docs easier.

Some important notes:

- The documentation for `has_resuming` has been left blank.  Despite my efforts, I wasn't able to determine what this was meant to indicate, and could find no examples of it being used in context with which to reverse engineer its purpose.  If someone is able to fill me in on what this method does, I would be happy to add documentation in.
- I haven't been working with perceval very long, and most of the knowledge I am contributing here is based on experimentation, reverse engineering, my own experiences working on perceval-osf, and extrapolating from what documentation existed before hand.  It's very possible that I've made a mistake in describing some of these methods, so please review carefully to make sure I am correct.  If I've made any mistakes, I will be happy to be corrected, and revise my mistake quickly after.

If you have any other advice, or recommended changes, I would be happy to hear!

### DIFF
--- a/perceval/backend.py
+++ b/perceval/backend.py
@@ -59,11 +59,22 @@ class Backend:
     will be named as 'origin'. During the initialization, an `Archive`
     object can be provided for archiving raw data from the repositories.
 
-    Derived classes have to implement `fetch_items`, `has_archiving` and
-    `has_resuming` methods. Otherwise, `NotImplementedError`
-    exception will be raised. Metadata decorator can be used together with
-    fetch methods but requires the implementation of `metadata_id`,
-    `metadata_updated_on` and `metadata_category` static methods.
+    To avoid a :class:`NotImplementedError`, derived classes have to implement
+    or define:
+     - :func:`fetch_items`, to retrieve items from the repository
+     - :func:`has_archiving`, whether this backend supports archives
+     - :func:`has_resuming`, whether this backend supports resuming
+     - :func:`metadata_id`, to produce a unique id from an item
+     - :func:`metadata_updated_on`, to find the last time an item was modified
+     - :func:`metadata_category`, to identify the category of an item
+     - :func:`_init_client`, to initialize the backend's client
+     - :data:`CATEGORIES`, defining the set of categories the backend produces
+     - [Optional] :data:`CLASSIFIED_FIELDS`, to hide certain fields from results
+     - [Optional] :data:`EXTRA_SEARCH_FIELDS`, to add easy access fields to items
+     - [Optional] :data:`ORIGIN_UNIQUE_FIELD`, to enable item blacklisting
+
+    For more information on the details of implementing these methods, please
+    see the docs on each method.
 
     The fetched items can be tagged using the `tag` parameter. It will
     be useful to trace data. When it is set to `None` or to an empty
@@ -73,17 +84,6 @@ class Backend:
     process, this class provides a `version` attribute that each backend
     may override.
 
-    Each backend should implement a class attribute named `CLASSIFIED_FIELDS`.
-    It will allow to filter from items those fields that may be considered
-    sensible or confidential. This attribute is a list of lists.
-    As items returned are dicts that may contain nested dicts, each entry
-    is a list which stores the "path" or nested dicts keys to the field to
-    remove. For example, `['my', 'classified', 'field']` will remove `field`
-    from `item['data']['my']['classified']` dict.
-
-    Classified data filtering and archiving are not compatible to prevent
-    data leaks or security issues.
-
     Each fetch operation generates a summary, available via the property
     `summary`. By default, it includes the last UUID generated, number
     of items fetched, skipped and their sum, plus the min, max and last
@@ -92,32 +92,10 @@ class Backend:
     the summary also includes some extra fields, which can be used by any
     backend to include fetch-specific information.
 
-    Each backend can also provide a set of search fields to simplify query
-    operations (avoiding the manual inspection of the items). The search
-    fields are included in a dict with the following shape:
-
-        {
-            'key-1': value-1,
-            'key-2': value-2,
-            'key-3': value-3
-        }
-
-    These fields are added to the item metadata information in the
-    `search_fields` attribute. By default, `search_fields` contains
-    the id of the item ('item_id': item_id_value), obtained via the
-    method `metadata_id`. However, each backend can set extra search
-    fields using the dict EXTRA_SEARCH_FIELDS. An example of
-    EXTRA_SEARCH_FIELDS is provided below:
-
-        {
-            'project_id': ['fields', 'project', 'id'],
-            'project_key': ['fields', 'project', 'key'],
-            'project_name': ['fields', 'project', 'name']
-        }
-
-    Each key in the dict is a search field to be included in the item
-    metadata information, while the corresponding value is a list that
-    stores the "path" of the search field value within the item.
+    Backends also produce a set of search fields, exposed in the
+    `search_fields` attribute of each item returned by a call to :func:`fetch`.
+    These contain the `item_id`, as well as any number of backend-specific
+    fields.
 
     :param origin: identifier of the repository
     :param tag: tag items using this label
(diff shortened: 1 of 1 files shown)


sduenas:
@Alch-Emi thanks for you PR, this is awesome. I'll review it tomorrow!

perceval/backend.py
```diff
@@ -352,17 +453,68 @@ def has_resuming(cls):
 
     @staticmethod
     def metadata_id(item):
+        """ Produce a unique identifier for an item
```
sduenas, line 456:
```suggestion
        """Produce a unique identifier for an item.
```

perceval/backend.py
```diff
@@ -352,17 +453,68 @@ def has_resuming(cls):
...
+
+        Given one of the items produced by :func:`fetch_items`, produce a
+        unique identifier for that item.  Typically, this is an identifier
```
sduenas, line 459:
```suggestion
        unique identifier for that item. Typically, this is an identifier
```

perceval/backend.py
```diff
@@ -352,17 +453,68 @@ def has_resuming(cls):
...
+        given by the repository itself, such as a commit hash or post id.
+
+        The id should be represented by a string
```
sduenas, line 462:
```suggestion
        The id should be represented by a string.
```

perceval/backend.py
```diff
@@ -352,17 +453,68 @@ def has_resuming(cls):
...
+        """
         raise NotImplementedError
 
     @staticmethod
     def metadata_updated_on(item):
+        """ Determine the last time an item was updated
```
sduenas, line 468:
```suggestion
        """Determine the last time an item was updated.
```

perceval/backend.py
```diff
@@ -352,17 +453,68 @@ def has_resuming(cls):
...
+        identify the last time this item was modified.
+
+        :returns: The timestamp of the last modification, represented as epoch
+            seconds (a UNIX timestamp)
+        """
         raise NotImplementedError
 
     @staticmethod
     def metadata_category(item):
+        """ Identify the category of a given item
```
sduenas, line 480:
```suggestion
        """Identify the category of a given item.
```

perceval/backend.py
```diff
@@ -352,17 +453,68 @@ def has_resuming(cls):
...
+
+        Every item returned by :func:`fetch_items` should belong to exactly one
+        of the categories listed in the `CATEGORIES` class member.  This method
```
sduenas, line 483:
```suggestion
        of the categories listed in the `CATEGORIES` class member. This method
```

perceval/backend.py
```diff
@@ -352,17 +453,68 @@ def has_resuming(cls):
...
+        Note that for all items returned by a call to :func:`fetch_items`, this
+        method is expected to return a category equivalent to the one passed as
+        the `category` argument.
+
+        :returns: One of the strings in `CATEGORIES`
+        """
         raise NotImplementedError
 
     def _init_client(self, from_archive=False):
+        """ Initialize the client to be used by the backend
```
sduenas, line 496:
```suggestion
        """Initialize the client to be used by the backend.
```

perceval/backend.py
```diff
@@ -352,17 +453,68 @@ def has_resuming(cls):
...
+
+        Many backends use a persistent HTTP client to retrieve information from
+        a backend.  This method is called before any calls to
```
sduenas, line 499:
```suggestion
        a backend. This method is called before any calls to
```

perceval/backend.py
```diff
@@ -352,17 +453,68 @@ def has_resuming(cls):
...
+        :func:`fetch_items`, and should be used as an opportunity to initialize
+        the client.
+
+        If the backend chooses to do so, then it should return an instance of
+        its client, which will then be immediately assigned to the `.client`
+        attribute of the backend.
+
+        If the backend chooses not to initialize a client, it may simply `pass`
+        or `return None`.  However, note that the `.client` attribute will
```
sduenas, line 508:
```suggestion
        or return `None`. However, note that the `.client` attribute will
```

perceval/backend.py
```diff
@@ -352,17 +453,68 @@ def has_resuming(cls):
...
+        still be overridden
```
sduenas, line 509:
```suggestion
        still be overridden.
```

perceval/backend.py
```diff
@@ -352,17 +453,68 @@ def has_resuming(cls):
...
+
+        Additionally, the client or the backend should be set to respect
+        archiving, based on the :func:`from_archive`.  If the backend does not
+        report that it :func:`has_archiving`, than this parameter may just be
+        ignored.
```
sduenas, line 514:
```suggestion
        Additionally, the client or the backend should be set to respect
        archiving, based on the :func:`from_archive`. If the backend does not
        report that it :func:`has_archiving`, then this parameter may just be
        ignored.
```

perceval/backend.py
```diff
@@ -352,17 +453,68 @@ def has_resuming(cls):
...
+
+        :returns: A client, or `None`
```
sduenas, line 516:
```suggestion
        :returns: a client, or `None`
```

sduenas (changes requested):
Thanks for your contribution. It's awesome. I added some suggestions to keep the format aligned with the style in the docstrings. Please change your PR and we can merge it. Don't forget to squash your commit and a single one and rebase it with master. Thanks!

sduenas:
>     * The documentation for `has_resuming` has been left blank.  Despite my efforts, I wasn't able to determine what this was meant to indicate, and could find no examples of it being used in context with which to reverse engineer its purpose.  If someone is able to fill me in on what this method does, I would be happy to add documentation in.
> 
@Alch-Emi it's hard to explain but I'll try to do it.

A backend has resuming or the fetching process can be resumed if when it stops, either because it failed or because there aren't more items to retrieve, items can be retrieved again from the point it stopped without data loss.

For example, if the `github` backend fails and it's retrieving data from `chaoss/grimoirelab-elk` repository, I can run again it using the last date of the last item it retrieved and run it again using this date as `from-date` parameter. The backend will start retrieving data from that date and we're sure that there's no a gap of data it didn't retrieve. This doesn't happen with `gerrit` backend.

In gerrit, data is retrieved internally from newest to oldest. If the process fails, it means it has to start from the newest item to retrieve everything it wasn't retrieved before. This newest item can be newer than the first time it run because these items could have been produced while the backend was running. If it fails again the problem becomes bigger because you might have two gaps of information that you didn't retrieve.

It's a bit confusing. I hope you understood it.

Alch-Emi:
I just pushed a commit with your changes!  I really need to unlearn typing double spaces, but I've been doing it since elementary school at this point, and old habits die hard, I guess 😅

I tried writing up the `has_resuming` docs as well, but I recommend you give it a look over to make sure I got everything right.

Thanks so much for the review!

perceval/backend.py
```diff
@@ -118,21 +168,30 @@ class Backend:
...
+    values which should not be included in results, if the field defined here
+    contains them.  For example, if `ORIGIN_UNIQUE_FIELD` were set to
+    `post_id`, then users could pass a list of post ids that should be excluded
+    from results.
+
+    If set to none, blacklisting will be disabled completely.  Otherwise, this
+    should be set to a :class:`OriginUniqueField` containing the number and
+    data type of the field.
+
+    Note:  Origin in this context refers to one site, api, or other remote that
```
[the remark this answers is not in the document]
sduenas: @Alch-Emi this is the only double space you forgot to fix, can you do it?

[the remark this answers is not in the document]
Alch-Emi: Oh thank you!

sduenas (approved):
LGTM

sduenas:
@Alch-Emi thank you so much for your contribution!

PR #742, opened 2021-07-14
[mattermost] Change the key from 'data.metadata.images' to value

zhquan:
This code converts the key (URL) to value when the field
'data.metadata.images' exists, thus the ES index does not have a
field for each image URL and consequently avoid the error
'Limit of total fields [1000] in index [...] has been exceeded'.

From:
```
{
  ...
  "images": {
    "https://.../api/v4/image?...": {
      "width": 1280,
      "height": 640,
      ...
    }
  }
}
```

To:
```
{
  ...
  "images": [
    {
      "url": "https://.../api/v4/image?...",
      "width": 1280,
      "height": 640,
      ...
    }
  ]
}

```

### DIFF
--- a/perceval/backends/core/mattermost.py
+++ b/perceval/backends/core/mattermost.py
@@ -73,7 +73,7 @@ class Mattermost(Backend):
         of connection problems
     :param ssl_verify: enable/disable SSL verification
     """
-    version = '0.4.1'
+    version = '0.5.0'
 
     CATEGORIES = [CATEGORY_POST]
     EXTRA_SEARCH_FIELDS = {
@@ -152,6 +152,9 @@ class Mattermost(Backend):
                 if post['update_at'] < since:
                     fetching = False
                     break
+
+                if 'metadata' in post and 'images' in post['metadata']:
+                    post['metadata']['images'] = self._parse_images(post['metadata']['images'])
 
                 # Fetch user data
                 user_id = post['user_id']
@@ -252,6 +255,12 @@ class Mattermost(Backend):
         # 'order' key.
         for post_id in parsed_posts['order']:
             yield parsed_posts['posts'][post_id]
+
+    def _parse_images(self, images):
+        """Parse images and returns a list of images."""
+
+        list_images = [{**images[i], **{'url': i}} for i in images]
+        return list_images
 
     def _get_or_fetch_user(self, user_id):
         if user_id in self._users:
--- a/tests/data/mattermost/mattermost_posts.json
+++ b/tests/data/mattermost/mattermost_posts.json
@@ -96,7 +96,23 @@
             "type": "",
             "props": {},
             "hashtags": "",
-            "pending_post_id": ""
+            "pending_post_id": "",
+            "metadata": {
+                "images": {
+                    "https://image-one-url.png": {
+                        "width": 1280,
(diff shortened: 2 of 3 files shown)


perceval/backends/core/mattermost.py
```diff
@@ -153,6 +153,12 @@ def fetch_items(self, category, **kwargs):
                     fetching = False
                     break
 
+                metadata = post.get('metadata', None)
+                if metadata:
+                    images = post['metadata'].get('images', None)
+                    if images:
+                        post['metadata']['images'] = self._parse_images(images)
```
sduenas, line 160: I wouldn't use `metadata` word because we have a decorator that already has that name. It shouldn't be a problem but it's better to find another name to avoid unexpected bugs. 
Also, as you don't use the variable, it's better to do:
```suggestion
                if 'metadata' in post and 'images' in post['metadata']
                    post['metadata']['images'] = self._parse_images(images)
```

perceval/backends/core/mattermost.py
```diff
@@ -253,6 +259,12 @@ def _parse_posts(self, raw_posts):
         for post_id in parsed_posts['order']:
             yield parsed_posts['posts'][post_id]
 
+    def _parse_images(self, images):
+        """Parse images and returns a list of images."""
+
+        list_images = [{**images[i], **{"url": i}}for i in images]
```
sduenas, line 265:
```suggestion
        list_images = [{**images[i], **{'url': i}} for i in images]
```

tests/data/mattermost/mattermost_posts.json
```diff
@@ -96,7 +96,17 @@
             "type": "",
             "props": {},
             "hashtags": "",
-            "pending_post_id": ""
+            "pending_post_id": "",
+            "metadata": {
```
sduenas, line 100: Can we add another item, so we check that the loop creating the images data works as expected?

sduenas (changes requested):
Please check my comments. Also you have to update the version number to `0.5.0` because the backend returns different data now.

perceval/backends/core/mattermost.py
```diff
@@ -253,6 +256,12 @@ def _parse_posts(self, raw_posts):
         for post_id in parsed_posts['order']:
             yield parsed_posts['posts'][post_id]
 
+    def _parse_images(self, images):
+        """Parse images and returns a list of images."""
+
+        list_images = [{**images[i], **{'url': i}}for i in images]
```
sduenas, line 262: You missed this part (space in between `}` and `for`.

```suggestion
        list_images = [{**images[i], **{'url': i}} for i in images]
```

tests/test_mattermost.py
```diff
@@ -203,6 +203,25 @@ def test_fetch(self):
             self.assertEqual(post['data']['channel_data']['name'], expected_channel[0])
             self.assertEqual(post['data']['channel_data']['display_name'], expected_channel[1])
 
+        # Check image
+        image_expected = [
```
sduenas, line 207: This should be renamed to `images_expected`.

sduenas (changes requested):
It's ok but just to minor comments you missed.

sduenas (approved):
LGTM

PR #743, opened 2021-07-16
[nntp] Handle not valid date

zhquan:
This code sets the date to the FALLBACK_DATE value when the date is not valid.
FALLBACK_DATE = '1970-01-01'.
Test updated accordingly.


sduenas (changes requested):
I'd rather have these two things in separate PRs. Sorry.

sduenas (approved):
LGTM

PR #748, opened 2021-08-05
[mattermost] Support running with channel names

Alch-Emi:
The current mattermost backend runs using a channel ID in order to select which channel is being scraped.  However, this can get in the way in certain circumstances.  Mainly, the channel ID can't be derived from a mattermost URL like

https://my-mattermost.link/some-team/channels/some-channel

This PR allows the mattermost backend to be called with a team name and a channel name as an alternative to the normal behavior.

This means that tools like grimoirelab-elk would be able to accept a normal mattermost URL instead of requiring a unique repository format, which would improve intuitivity.  Additionally, the channel/team names are easier for users to find than the channel ID, which means a smoother user experience.

This does not break the existing workflow, and the previous usage still exists, just with the ability to alternatively specify names. Similarly, any new function parameters use keyword arguments, as to prevent creating a breaking change.

Love to hear your thoughts!

― Emi

### DIFF
--- a/perceval/backends/core/mattermost.py
+++ b/perceval/backends/core/mattermost.py
@@ -57,15 +57,21 @@ class Mattermost(Backend):
     have enough permissions to read from the given channel.
 
     To initialize this class the URL of the server must be provided.
-    The origin of data will be set using this `url` plus the
-    channel from data is obtained (i.e: https://mattermost.example.com/abcdefg).
+    The origin of data will be set using this `url` plus the channel
+    from data is obtained (i.e: https://mattermost.example.com/abcdefg).
+    If using channel and team names instead of a channel id, this will
+    take the form `url` plus `team` plus `channel`.
+
+    The `team` parameter is only required if providing a channel
+    name instead of a channel ID.
 
     :param url: URL of the server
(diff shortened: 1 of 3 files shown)


Alch-Emi:
Pushed a few amendments to add tests, among a couple other small fixes.  See the commit message for more details

jjmerchante:
Thanks for the PR. Looks good to me!

Add a test in the method `test_setup_cmd_parser` when passing a team name and channel name.

I was wondering if it works for ELK and Mordred. The URL passed to ELK is in the format `https://mymattermost.link abcde123456` (https://github.com/chaoss/grimoirelab-sirmordred#mattermost-).

For this to work with Mordred, I think a change may be required here: https://github.com/chaoss/grimoirelab-elk/blob/00b346623d8f3f0b4de901d25a0ef28e79a02545/grimoire_elk/raw/mattermost.py#L85

Alch-Emi:
Thanks!  I just added that new test!  I was also planning to send a PR to ELK after the next perceval release as well in order to enable support for channel URLs there as well!

sduenas (approved):
LGTM

Issue #752, opened 2021-08-24
Jira Authentication Problem


isikozsoy:
Hello,
I could not understand what should I use for user (Jira's attribute). I tried email and full-name. I could not find another option to use  for user. Can anybody help me with this? Thanks for your time.

vchrombie:
Hi @isikozsoy, thanks for opening the issue.

I think the `user` attribute is the username for the Jira instance.
- [perceval/backends/core/jira.py#L276](https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/backends/core/jira.py#L276)
- [perceval.backends.core.html#perceval.backends.core.jira.JiraClient](https://perceval.readthedocs.io/en/latest/perceval.backends.core.html#perceval.backends.core.jira.JiraClient)

isikozsoy:
Thanks for your response. However, I also checked if I can use username but I think username is  not available in Jira. You can check it from the links.

https://confluence.atlassian.com/cloud/blog/2018/06/say-goodbye-to-usernames-in-atlassian-cloud
https://community.atlassian.com/t5/Jira-questions/How-to-find-username-of-my-jira-account/qaq-p/969298

isikozsoy:
Is it possible to modify the source code for Jira in a way that api token can be used for authorization like Github or GitLab? @vchrombie 

https://support.atlassian.com/atlassian-account/docs/manage-api-tokens-for-your-atlassian-account/

vchrombie:
Thanks for providing the info, @isikozsoy.

> https://confluence.atlassian.com/cloud/blog/2018/06/say-goodbye-to-usernames-in-atlassian-cloud

Didn't know about this. Is this applicable only for the new versions?

> Is it possible to modify the source code for Jira in a way that api token can be used for authorization like Github or GitLab? @vchrombie

It should be possible, but I'm wondering if it would break any existing systems. I would like to keep this issue open for more comments from the other community members.

Thanks for reporting the issue.

vchrombie:
@isikozsoy thanks a lot.

Do you think we can add something to make things clear?
Would you be interested to open a PR for it?

isikozsoy:
You're welcome @vchrombie.

We can use email instead of user and token instead of password in the soure code however this might break existing systems so I think it is better if we just modify the explanation for user and password parameters of Jira class by editing the comment. I can open a PR if you also agree with my suggestion.

vchrombie:
[...]

PR #754, opened 2021-08-28
Clarify parameter explanations of the Jira backend

isikozsoy:
This PR aims at providing a quick fix for the issue #752 by changing the explanation of the parameters for Jira and JiraClient backend.

### DIFF
--- a/perceval/backends/core/jira.py
+++ b/perceval/backends/core/jira.py
@@ -96,8 +96,8 @@ class Jira(Backend):
 
     :param url: JIRA's endpoint
     :param project: filter issues by project
-    :param user: Jira user
-    :param password: Jira user password
+    :param user: JIRA's user email
+    :param password: JIRA's api token
     :param cert: SSL certificate path (PEM)
     :param max_results: max number of results per query
     :param tag: label used to mark the data
@@ -273,8 +273,8 @@ class JiraClient(HttpClient):
 
     :param URL: URL of the JIRA server
     :param project: filter issues by project
-    :param user: JIRA's username
-    :param password: JIRA's password
+    :param user: JIRA's user email
+    :param password: JIRA's api token
     :param cert: SSL certificate
     :param max_results: max number of results per query
     :param archive: an archive to store/read fetched data


perceval/backends/core/jira.py
```diff
@@ -96,8 +96,8 @@ class Jira(Backend):
 
     :param url: JIRA's endpoint
     :param project: filter issues by project
-    :param user: Jira user
```
vchrombie, line 99: I would change this to
```
-      :param user: Jira user
+      :param user: Jira user (email)
```

perceval/backends/core/jira.py
```diff
@@ -96,8 +96,8 @@ class Jira(Backend):
...
-    :param password: Jira user password
```
vchrombie, line 100: I would revert this change because the current implementation is based on password. It is better if we change it only after the implementation is changed to token based.

perceval/backends/core/jira.py
```diff
@@ -273,8 +273,8 @@ class JiraClient(HttpClient):
 
     :param URL: URL of the JIRA server
     :param project: filter issues by project
-    :param user: JIRA's username
```
vchrombie, line 276: `username` might be misleading, so it would be great if we can follow the above comment
```
-      :param user: Jira user
+      :param user: Jira user (email)
```

perceval/backends/core/jira.py
```diff
@@ -273,8 +273,8 @@ class JiraClient(HttpClient):
...
-    :param password: JIRA's password
```
vchrombie, line 277: same as the above comment, revert this change for now and change it only after implementation

vchrombie (changes requested):
Thanks for opening the PR @isikozsoy.
I had some small comments on the PR, please let me know what do you think.

Also, it would be great if you can fix the sign-off issue and maybe amend the commits too. Please don't hesitate to ask for help.

perceval/backends/core/jira.py
isikozsoy: When I used Jira backend in my project, I use email for the user field and api token for the password field. However, it did not work when I used email and password for authentication. Should I still revert this change?

isikozsoy:
_Also, it would be great if you can fix the sign-off issue and maybe amend the commits too. Please don't hesitate to ask for help._

Could you guide me to fix the sign-off issue and to amend the commits please?

isikozsoy:
_Also, it would be great if you can fix the sign-off issue and maybe amend the commits too. Please don't hesitate to ask for help._

Could you guide me to fix the sign-off issue and to amend the commits please?

perceval/backends/core/jira.py
vchrombie: Since the current implementation is based on password, changing the comment to token might not look good. It would be really great if you can work on the feature of changing the basic_auth to token_auth. You can work in the same PR. WDYT?

vchrombie:
Hi @isikozsoy 

> Could you guide me to fix the sign-off issue and to amend the commits please @vchrombie? I also commented on one of your change requests. Thanks.

Sure, you just need to amend the commit with `-s` flag and then force push the branch
```
$ git commit --amend -s
$ git push -f origin jira-backend-description
```

Hope this helps.

sduenas:
This PR has been inactive for many years. Please feel free to reopen it if you resume your work on this.

PR #757, opened 2021-09-06
[github] Extend 'max_items' for all requests

### LINKED ISSUE #755: `max-items` is just used for issues, but not for others
https://github.com/chaoss/grimoirelab-perceval/blob/4a65eceb6463f1847af6c6e8e5da6b527f6c30e7/perceval/backends/core/github.py#L731

The value of `max-items` is only used for `issues` method, so what is the point of setting this value?

For example:
https://github.com/chaoss/grimoirelab-perceval/blob/4a65eceb6463f1847af6c6e8e5da6b527f6c30e7/perceval/backends/core/github.py#L711
when we called `https://api.github.com/repos/apache/spark/pulls/27058/comments?per_page=100&direction=asc&sort=updated`, we got `Server Error`.
But if `per_page` is 50, the response is 200.

It should also be used in others methods so that we can set `per_page` what we want.

zhquan:
This code extends `max_items` for all requests instead of a
default value `VPER_PAGE = 100`.

Fixes #755

### DIFF
--- a/perceval/backends/core/github.py
+++ b/perceval/backends/core/github.py
@@ -111,7 +111,7 @@ class GitHub(Backend):
         of connection problems
     :param ssl_verify: enable/disable SSL verification
     """
-    version = '0.26.0'
+    version = '0.27.0'
 
     CATEGORIES = [CATEGORY_ISSUE, CATEGORY_PULL_REQUEST, CATEGORY_REPO]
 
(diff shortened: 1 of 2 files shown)


sduenas (changes requested):
LGTM but a minor change is needed. `VPER_PAGE` is not used anymore so it's safe to remove it from the code.

sduenas (commented):
I also forgot that you have to increase the backend version number to `0.27.0`.

sduenas (approved):
LGTM

PR #769, opened 2021-12-10
[gha] Validate changelog file on every PR

jjmerchante:
This PR adds a new GitHub action to check if every
pull request includes a changelog file and
validate the content if it is empty or valid YAML
file.

This Pull request depends on https://github.com/Bitergia/release-tools/pull/35

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,0 +1,12 @@
+name: Check changelog file included
+
+on:
+  pull_request:
+    paths:
+      - 'perceval/**'
+
+jobs:
+  check-changelog:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: bitergia/release-tools-check-changelog@master


.github/workflows/changelog.yml
```diff
@@ -0,0 +1,10 @@
+name: Check changelog file included
+
+on:
+  pull_request:
+    branches:
+      - '**'
+
```
vchrombie, line 7: Thanks for opening the PR @jjmerchante 

Coming from the comment https://github.com/chaoss/grimoirelab-toolkit/pull/46#discussion_r778047400

Do you think we can add a similar rule for perceval too?

```
on:
  pull_request:
    paths:
      - 'perceval/'
```

This would avoid the workflow running for all the PRs. Instead, it will be triggered only when there is a change in the source code (perceval folder).

I'm not sure about the `branches` rule. It should be helpful if we are having a developing branch similar to sortinghat and bestiary. Do you have any other advantage that I'm missing?

vchrombie (approved):
LGTM!

@sduenas Do you have any comments?

sduenas (approved):
LGTM

PR #776, opened 2022-02-24
Add configurations of the metadata files to pyproject.toml

vchrombie:
This PR adds the configurations of the metadata files to the `pyproject.toml` file. This is needed to include these files when the package is built using poetry.


sduenas (approved):
LGTM

PR #781, opened 2022-04-19
Drop Python 3.6 support

jjmerchante:
Python 3.6 reached its end of life at the end of 2021. This PR removes its support in Poetry and GitHub actions

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -22,7 +22,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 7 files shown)


vchrombie (changes requested):
Thanks for opening the PR @jjmerchante.
I think the python version for the docs also can be updated to the new version, WDYT?

https://github.com/chaoss/grimoirelab-perceval/blob/6aa6c65923837f8fb1e3036e5ec2673145295a38/.readthedocs.yml#L11

vchrombie (approved):
LGTM!

Do you have any comments @sduenas?

PR #783, opened 2022-05-12
[confluence] Support fetch by spaces.

zhquan:
This code allows fetching confluence by specific spaces. By
default it will fetch the entire instance, but adding the
`--spaces` argument will fetch only these spaces.

New optional argument:
- `--spaces`: List of spaces to fetch

Tests added accordingly.
Version updated to 0.13.0.


jjmerchante:
LGTM

PR #785, opened 2022-05-25
Support package namespaces with Poetry

jjmerchante:
Convert this package in a native namespace package as defined by PEP 420.

It appears that it was not updated when we updated all other Perceval repositories.


sduenas (approved):
LGTM

PR #786, opened 2022-06-01
Update requirements

sduenas:
This PR updates grimoirelab toolkit depedency and creates a file with the dependencies with developers when the want to use the latest version of the Grimoirelab stack.


jjmerchante:
LGTM

PR #787, opened 2022-06-07
[slack] Fix username KeyError

zhquan:
This code fixes the `KeyError: username` when the user was removed.

when a user was deleted, the item doesn't have any user information,
so the "metadata_id" method will generate the identifier using only
the 'ts' value.

### DIFF
--- a/tests/test_slack.py
+++ b/tests/test_slack.py
@@ -858,6 +858,55 @@ class TestSlackClient(unittest.TestCase):
 
         self.assertEqual(user, '{"ok":false,"user":null}')
 
+    def test_metadata_id(self):
+
+        client = Slack('CHANNEL', 'TOKEN')
+
+        # When 'user', 'bot_id', 'comment', or 'username' do not exist.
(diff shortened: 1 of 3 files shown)


tests/test_slack.py
```diff
@@ -858,6 +858,72 @@ def test_private_user(self):
...
+                'id': 'CHANNEL',
+                'name': 'test',
+                'is_channel': True,
+                'num_members': 4
+            }
+        }
+        metadata_id = client.metadata_id(item)
+        expected_id = item['ts']
+
+        self.assertEqual(metadata_id, expected_id)
```
jjmerchante, line 897: Just a minor comment, I think this test is the same as the previous one.

jjmerchante (changes requested):

tests/test_slack.py
zhquan: ups

jjmerchante:
LGTM

PR #790, opened 2022-06-28
[confluence] Fix error when args not passed on 'fetch_items'

sduenas:
When required arguments weren't given in 'fetch_items' the code failed with a KeyError exception. This patch fixes
the error checking if the arguments are given and, if not, setting the default value parameters.


zhquan (approved):
LGTM

PR #796, opened 2022-09-06
[deps] Update flake8 dependency to 4.0.1

jjmerchante:
This PR updates the `flake8` dependency to 4.0.1 to make it consistent with other grimoirelab repositories and keep dependencies more stable.

It also updates how the package is installed in the release because it was not installing the new package.


sduenas (approved):
LGTM

PR #801, opened 2022-09-26
Use `--url` for Gitlab URL argument

xurizaemon:
Hi there! I am taking a look at the Gitlab plugin to familiarize myself with how backend plugins are written.

I noticed the Gitlab code refers to enterprise edition but I think the switch really is for setting the base URL - ie it distinguishes between SaaS Gitlab (gitlab.com) and self-managed (gitlab.example.com).

So I propose renaming `--enterprise-url` to `--url` (and retaining the old label for compatibility).

I expect with this change that `--url https://gitlab.com` should be equivalent to omitting the argument.

Hoping that I haven't misunderstood some other distinction between enterprise and not.

I won't be at all bothered if this PR is not accepted, I'm using it to get familiar as much as to fix a bug 😁 

If there's useful test coverage can be added happy to do that. Excited to see if tests run when I click "Create PR"!


xurizaemon:
Reading [CONTRIBUTING.md](https://github.com/chaoss/grimoirelab-perceval/blob/master/CONTRIBUTING.md) now. It's reminded me that if there was a CHANGELOG I should update it (but Perceval doesn't have a CHANGELOG, should it, or would that go to Grimoire?)

sduenas (changes requested):
@xurizaemon thanks for your PR. Would you mind to fix the flake8 error and squash the commits?

sduenas:
> 

You have to create a new changelog entry on `releases/unreleased` directory. The format is described on the contributing file but you can also use our [release tools](https://github.com/Bitergia/release-tools) to generate it.
If you have any problem, I can create the file and add it to your PR.

xurizaemon:
OK @sduenas thanks for the guidance, that's done now

sduenas (approved):
Thanks for the changes. I rebased and squashed the commits plus I rephrased the entry on the change log. Hope you don't mind.

Thanks for your contribution!

PR #805, opened 2022-10-20
[pyproject] Update build system to poetry-core

jjmerchante:
This PR updates the build system in the pyproject file to poetry-core.

The previous build system was 'setuptools<50' which is not PEP 660 compliant and packages could not be installed in editable mode.


sduenas (approved):
LGTM

PR #815, opened 2023-04-24
Drop Python 3.7 support

jjmerchante:
Python 3.7 reaches the end of life in June 2023. This means it won't receive new updates or patches to fix security issues.
Therefore, this package only works with Python >= 3.8 from now on.

Update the version of the packages


poetry.lock
```diff
@@ -244,98 +227,94 @@ files = [
...
 ]
 
 [package.extras]
-toml = ["toml"]
+toml = ["tomli"]
 
 [[package]]
 name = "cryptography"
-version = "41.0.3"
+version = "40.0.2"
```
sduenas, line 294: Any reason why this should be now lower than on the other version?

sduenas (approved):
LGTM

PR #816, opened 2023-04-28
Update README with the supported platforms

sduenas:
These platforms are Linux and Mac. Windows is still not supported.


jjmerchante:
LGTM

PR #818, opened 2023-05-25
[git] Fetch commits using Git alternates mechanism

jjmerchante:
This PR allows Git to fetch commits from alternate object stores when that mechanism is defined.


sduenas (approved):
LGTM

PR #822, opened 2023-08-29
[ci] Install dev packages using Poetry instead of peodd

jjmerchante:
Poetry supports installing packages from a specific group since version `1.2`. Replace the use of `peodd` with Poetry in the installation of the test packages.


sduenas (approved):
LGTM

PR #831, opened 2023-12-15
Downgrade urllib3 from ^2.0.7 to ^1.26

jjmerchante:
This change is because there is a bug in urllib3 v2 with dulwich for Perceval.


sduenas (approved):
LGTM

PR #836, opened 2024-03-01
Remove NNTP tests not compatible with Python 3.11.8

jjmerchante:
This PR removes the tests for NNTP. It is now deprecated and will be removed in a future version of Perceval. It is not compatible with Python 3.11.8.


sduenas (approved):
LGTM

PR #838, opened 2024-03-07
Recovery functionality for Git

jjmerchante:
This PR implements the recovery functionality in the fetch method of the Git backend.

It also includes the last item search fields in the summary to allow recovering from the last item.

### DIFF
--- a/tests/test_git.py
+++ b/tests/test_git.py
@@ -129,6 +129,19 @@ class TestGitBackend(TestCaseGit):
 
         self.assertEqual(Git.has_resuming(), True)
 
+    def test_metadata(self):
+        """Test that the offset metadata is the commit from the item"""
+
+        new_path = os.path.join(self.tmp_path, 'newgit')
+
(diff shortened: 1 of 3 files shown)


tests/test_git.py
```diff
@@ -706,6 +706,95 @@ def test_fetch_from_file(self):
             self.assertEqual(commit['category'], 'commit')
             self.assertEqual(commit['tag'], 'http://example.com.git')
 
+    def test_fetch_recovery(self):
```
sduenas, line 709: The test looks great but I think you can improve it splitting in on different cases and not having everything tested on the same one.

perceval/backend.py
```diff
@@ -1035,6 +1035,7 @@ def __init__(self):
         self.max_offset = None
         self.last_offset = None
         self.extras = None
+        self.last_search_fields = None
```
sduenas, line 1038: Why is this field needed?

sduenas (approved):
LGTM

PR #841, opened 2024-06-10
[git] Fix file stats for merged files on subdirs

sduenas:
Stats about changes on a file were not reported correctly for files that where moved to a subdirectory. They were reported as an invalid entry and without action associated.

For example, the file `dir/filename` was moved to `dir/subdir/filename`, but it was reported as `dir//filename`. Therefore, the entry of the file `dir/subdire/filename` didn't have the stats about added and deleted lines. This error has been fixed and stats are reported correctly.


jjmerchante:
LGTM

PR #843, opened 2024-06-13
[ci] Fix coveralls parallel run

sduenas:
The error "Can't add a job to a build that is already closed" is fixed assigning a unique name to each parallel run of the tests.


jjmerchante:
LGTM

PR #848, opened 2024-08-28
Simplify `fetch()` and `fetch_items()` in the backends

### LINKED ISSUE #527: Simplify fetch() and fetch_items()
To retrieve items from a data source, `Backend` abstract class implements the method `fetch`. This method creates a generator that will get data calling to `fetch_items`. By default, this method is not implemented because the logic to fetch data is specific for each data source.

Backends also have to override `fetch` method to add specific parameters. This method also needs to call to the parent `fetch` that will all the job. This is a problem because any change in `Backend.fetch()` needs to be spread around all the classes (e.g move `archive` from `__init__` to `fetch`).

In my opinion, `fetch` shouldn't be overiden. The only method backends should implement to add the retrieval logic is `fetch_items`.

vchrombie:
Fixes #527

I trying an attempt to solve this issue. I picked the Gitter backend to get started and see get some feedback on the changes. I will be working on rest of the backends one by one. I am sure this is going to be a huge PR, so I will take some time and add a backend each time to allow time for reviewing.

Let me know if you have any suggestions.


sduenas:
I saw two weeks ago some code to change this problem but I don't see it now. Any issues with it? Is there any way I can help?

vchrombie:
I am taking a stab at this again, will fix the linting issues and remove the changes to tests.

vchrombie:
@sduenas, could you please take a chance to review this approach? What do you think, let me know if you have any suggestions.

PR #852, opened 2024-11-20
[git] Assign stats for files with whitespaces

sduenas:
Git commit stats weren't assigned to the action for those files with whitespaces in their filenames.
Actions didn't have stats and empty stats entries were created, like in a merge commit.

The problem was that the regular expression used to parse the stats considered whitespaces as separators.


jjmerchante (approved):
LGTM

PR #853, opened 2024-12-03
[git] Update references changed on the remote repo only

sduenas:
When repositories are synched, the backend needs to update the local references to the ones set on the remote repository. The backend processed all the references, including those up to date. This made the synchronization very slow.

With this commit, only the references that were updated on the remote repository will be modified locally. We have improved the performance by 2. Repositories are in sync two or more times than before.

Performance data using `https://github.com/eclipse-platform/eclipse.platform.releng.buildtools.git`, when the repository hasn't been updated.

- Before this patch
```
time perceval git https://github.com/eclipse-platform/eclipse.platform.releng.buildtools.git --latest-items

real	0m5,380s
user	0m1,802s
sys	0m2,000s
```

- Using this patch
```
time perceval git https://github.com/eclipse-platform/eclipse.platform.releng.buildtools.git --latest-items

real	0m2,551s
user	0m0,823s
sys	0m0,115s
```

### DIFF
--- a/perceval/backends/core/git.py
+++ b/perceval/backends/core/git.py
@@ -1304,6 +1304,8 @@ class GitRepository:
             self._update_ref(old_ref, delete=True)
 
         # Update new references
+        current_refs = {ref.refname: ref.hash for ref in self._discover_refs()}
+
         for new_ref in refs:
             refname = new_ref.refname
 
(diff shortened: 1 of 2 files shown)


jjmerchante (approved):
LGTM

PR #854, opened 2024-12-03
[tests] Do not print data running 'run_tests.py'

sduenas:
Some tests procuced some noise when the are run with 'run_tests.py' script. Data is send to '/dev/null' or similar.


jjmerchante (approved):
LGTM

Issue #860, opened 2025-03-24
Breaks with Python 3.13


pslldq:
This package doesn't work with Python 3.13, as `nntplib` was [deprecated with Python 3.11 and removed in Python 3.13](https://docs.python.org/3.13/library/nntplib.html) (see also [PEP 594 – Removing dead batteries from the standard library](https://peps.python.org/pep-0594/#nntplib)).

This results in an error when using it:

```
$ python --version
Python 3.13.2
$ python -m venv testenv
$ . ./testenv/bin/activate
(testenv) $ pip install perceval
Collecting perceval
  Using cached perceval-1.1.1-py3-none-any.whl.metadata (16 kB)
Collecting PyJWT<3.0.0,>=2.4.0 (from perceval)
  Using cached PyJWT-2.10.1-py3-none-any.whl.metadata (4.0 kB)
Collecting beautifulsoup4<5.0.0,>=4.3.2 (from perceval)
  Using cached beautifulsoup4-4.13.3-py3-none-any.whl.metadata (3.8 kB)
Collecting cryptography<44.0.0,>=41.0.3 (from perceval)
  Using cached cryptography-43.0.3-cp39-abi3-manylinux_2_28_x86_64.whl.metadata (5.4 kB)
Collecting dulwich<1.0.0,>=0.21.5 (from perceval)
  Using cached dulwich-0.22.8-cp313-cp313-manylinux_2_17_x86_64.manylinux2014_x86_64.whl.metadata (4.9 kB)
Collecting feedparser<7.0.0,>=6.0.8 (from perceval)
  Using cached feedparser-6.0.11-py3-none-any.whl.metadata (2.4 kB)
Collecting grimoirelab-toolkit>=0.3 (from perceval)
  Using cached grimoirelab_toolkit-1.0.9-py3-none-any.whl.metadata (3.0 kB)
Collecting python-dateutil<3.0.0,>=2.6.0 (from perceval)
  Using cached python_dateutil-2.9.0.post0-py2.py3-none-any.whl.metadata (8.4 kB)
Collecting requests<3.0.0,>=2.7.0 (from perceval)
  Using cached requests-2.32.3-py3-none-any.whl.metadata (4.6 kB)
Collecting urllib3<3.0,>=2.2 (from perceval)
  Using cached urllib3-2.3.0-py3-none-any.whl.metadata (6.5 kB)
[...]

PR #861, opened 2025-03-25
Deprecate NNTP backend

jjmerchante:
`nntplib` is no longer part of the Python standard library. It was removed in Python 3.13 after being deprecated in Python 3.11 (PEP 594).

This PR adds a warning when using the NNTP backend to inform users of potential issues in future Python versions (>= 3.13), as nntplib has been removed from the standard library.

This PR also includes Python 3.13 in GitHub action tests

Related to https://github.com/chaoss/grimoirelab-perceval/issues/860

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -22,7 +22,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.9', '3.10', '3.11', '3.12']
+        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
     steps:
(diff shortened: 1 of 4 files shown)


sduenas:
Why don't we show a warning instead of removing it? We can show warning for versions of Pyhton < 3.13 saying the backend will be removed in the near future. For Python >= 3.13 we can just return and do nothing or show an error if someone tries to use it.

What do you think?

jjmerchante:
I updated the PR with your comments, it now raises an ImportError when trying to use the backend with Python 3.13 and a deprecation warning message when using the backend with a lower version.

sduenas (approved):
LGTM

PR #863, opened 2025-03-26
[git] Fix filename parsing for files with brackets

jjmerchante:
Perceval misinterpreted file paths containing curly brackets in the stats section of a log. It incorrectly detects them as renamed or moved. This was causing that the output contains new files that are not associated with any action.

This fix ensures that a file is only considered renamed if the file path contains '=>', reducing false positives of moved files.


sduenas (approved):
LGTM

PR #864, opened 2025-04-08
Update changelog category causing 2.x version

jjmerchante:
Revert the commits from the release to avoid triggering a 2.x version. This PR updates the changelog category from 'removed' to 'deprecated', which previously caused a Perceval 2.x release to be generated.


sduenas (approved):
LGTM

Issue #865, opened 2025-04-25
GitHub the default from date (1970-01-01) is not working


zhquan:
Perceval cannot fetch data using the default from date, but it works since `1970-01-01T08:00:00`. I think this error comes from the GitHub API.

- `1970-01-01T00:00:00+00:00` empty list https://api.github.com/repos/chaoss/grimoirelab-core/issues?state=all&per_page=100&direction=asc&sort=updated&since=1970-01-01T00%3A00%3A00%2B00%3A00
- `1970-01-01T08:00:00+00:00` works https://api.github.com/repos/chaoss/grimoirelab-core/issues?state=all&per_page=100&direction=asc&sort=updated&since=1970-01-01T08%3A00%3A00%2B00%3A00
 
Examples of the executions:
- 
- default or `1970-01-01T00:00:00`
```
$ perceval github chaoss grimoirelab-core -t *** > /dev/null
[2025-04-25 13:58:49,643] - Sir Perceval is on his quest.
[2025-04-25 13:58:50,142] - Summary of results

	   Total items: 	0
	Items produced: 	0
	 Items skipped: 	0

	Last item UUID: 	-
	Last item date: 	-

	Min. item date: 	-
	Max. item date: 	-

	Min. offset: 	-	Max. offset: 	-	Last offset: 	-


[2025-04-25 13:58:50,143] - Sir Perceval completed his quest.

$ perceval github chaoss grimoirelab-core -t *** --from-date 1970-01-01T00:00:00 > /dev/null
[2025-04-25 13:58:43,918] - Sir Perceval is on his quest.
[2025-04-25 13:58:44,386] - Summary of results

	   Total items: 	0
	Items produced: 	0
	 Items skipped: 	0

	Last item UUID: 	-
	Last item date: 	-

	Min. item date: 	-
	Max. item date: 	-

	Min. offset: 	-	Max. offset: 	-	Last offset: 	-


[2025-04-25 13:58:44,389] - Sir Perceval completed his quest.
```
- `1970-01-01T08:00:00`
```
$ perceval github chaoss grimoirelab-core -t *** --from-date 1970-01-01T08:00:00 > /dev/null
[2025-04-25 13:58:30,716] - Sir Perceval is on his quest.
[2025-04-25 13:58:35,533] - Summary of results

	   Total items: 	50
	Items produced: 	50
	 Items skipped: 	0

	Last item UUID: 	1bad5e7589efa3632f48cb88e74a866926894c04
[...]

PR #866, opened 2025-05-05
[GitHub] Update default from-date to avoid issues with GitHub API

jjmerchante:
GitHub API doesn't return issues using the default from date (1970-01-01)

This PR implements a temporary fix to prevent using the epoch time when fetching issues from the GitHub API.

Related to https://github.com/chaoss/grimoirelab-perceval/issues/865


sduenas (approved):
LGTM

PR #869, opened 2025-05-07
[github] Return basic user data when the API does not find the user

jjmerchante:
This PR ensures that when the API returns a 404 for a user, Perceval returns the basic user data available in the related issue, pull request, comment, or review.


sduenas (approved):
LGTM

PR #870, opened 2025-05-16
[git] Fix moved file detection in stats

jjmerchante:
This PR fixes a bug that caused some moved files to be incorrectly detected in the commit file statistics.

Include all the special cases in tests.

### DIFF
--- a/perceval/backends/core/git.py
+++ b/perceval/backends/core/git.py
@@ -571,11 +571,6 @@ class GitParser:
 
     EMPTY_LINE_PATTERN = r"^$"
 
-    RENAME_PREFIX = r"^(?P<prefix>.*/)?"
-    RENAME_FILENAME = r"\{(?P<old_name>.*) => (?P<new_name>(?:(?!}/).)+)?\}"
-    RENAME_SUFFIX = r"(?P<suffix>/[^/]+)?$"
-    RENAME_BRACES_PATTERN = RENAME_PREFIX + RENAME_FILENAME + RENAME_SUFFIX
-
     # Compiled patterns
     GIT_COMMIT_REGEXP = re.compile(COMMIT_PATTERN, re.VERBOSE)
     GIT_HEADER_TRAILER_REGEXP = re.compile(HEADER_TRAILER_PATTERN, re.VERBOSE)
@@ -583,7 +578,6 @@ class GitParser:
     GIT_ACTION_REGEXP = re.compile(ACTION_PATTERN, re.VERBOSE)
     GIT_STATS_REGEXP = re.compile(STATS_PATTERN, re.VERBOSE)
     GIT_NEXT_STATE_REGEXP = re.compile(EMPTY_LINE_PATTERN, re.VERBOSE)
-    GIT_RENAMED_BRACES_REGEXP = re.compile(RENAME_BRACES_PATTERN)
 
     # Git parser status
     (INIT,
@@ -601,8 +595,7 @@ class GitParser:
         self.state = self.INIT
 
         # Aux vars to store the commit that is being parsed
-        self.commit = None
-        self.commit_files = {}
+        self._cleanup_commit()
 
         self.handlers = {
             self.INIT: self._handle_init,
@@ -640,13 +633,20 @@ class GitParser:
 
         commit = self.commit
         commit = remove_none_values(commit)
+
+        self._handle_pending_files()
+
         commit['files'] = [remove_none_values(item)
                            for _, item in sorted(self.commit_files.items())]
 
-        self.commit = None
-        self.commit_files = {}
+        self._cleanup_commit()
 
         return commit
+
+    def _cleanup_commit(self):
+        self.commit = None
+        self.commit_files = {}
+        self.pending_files = {}
 
     def _handle_init(self, line):
         m = self.GIT_NEXT_STATE_REGEXP.match(line)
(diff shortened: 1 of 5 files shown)


perceval/backends/core/git.py
```diff
@@ -764,19 +763,26 @@ def _handle_trailer(self, line):
...
         self.commit_files[filename]['action'] = data['action']
-        self.commit_files[filename]['file'] = filename
+        self.commit_files[filename]['file'] = data['file']
         self.commit_files[filename]['newfile'] = data['newfile']
 
     def _handle_stats_data(self, data):
-        filename = self.__get_old_filepath(data['file'])
+        if ' => ' in data['file']:
+            filename = self.__get_new_filepath(data['file'])
+            if not self.merge_commit and self.commit_files and filename not in self.commit_files:
```
sduenas, line 780: Because this is a long check and hard to understand, maybe it's better to create a function that returns the result of this check:

```
def __should_guess_stats_filename(self):
    return not self.merge_commit and self.commit_files and filename not in self.commit_files
```

sduenas, line 780: Why shouldn't do this for merge commits?  They can also have actions where files are copied, renamed, etc.

perceval/backends/core/git.py
```diff
@@ -791,40 +797,66 @@ def __parse_data_list(self, data, sep=' '):
...
-        """Get the old filepath of a moved/renamed file.
+    def _guess_new_filename(self, rename_path):
+        """
+        Guess the new filename of a moved or renamed file.
+
+        Checks the part before or after a ' => ' in the path
+        (removing braces) against the old and new filenames.
+
+        Return None if the new filename cannot be guessed.
+        """
```
sduenas, line 808:
```suggestion
        """Guess the new filename of a moved or renamed file.
        
        Moved or renamed files can be found in the stats section of the log
        with any of the next patterns:
        
          'old_name => new_name'
          '{old_prefix => new_prefix}/name'
          'name/{old_suffix => new_suffix}'
        
        However, filenames can contain '{', '}', or ' => ' characters that
        can make very difficult to guess the real filenames.
        
        The approach here is to compare the parts before or after a ' => '
        substring, with the filenames parsed on the actions sections, removing
        the '{', and '}' characters from all the filenames. Action filenames contain
        the real name without those meta characters, so it's likely that this
        method will find the real name. When no name is found, the method will
        return 'None'.
        """
```

perceval/backends/core/git.py
```diff
@@ -791,40 +797,66 @@ def __parse_data_list(self, data, sep=' '):
...
+        parts = rename_path.split(' => ')
+        for i in range(1, len(parts)):
```
sduenas, line 810:
```suggestion
        parts = rename_path.split(' => ')
        
        for i in range(1, len(parts)):
```

perceval/backends/core/git.py
```diff
@@ -791,40 +797,66 @@ def __parse_data_list(self, data, sep=' '):
...
+            before = ' => '.join(parts[:i]).replace('{', '')
+            after = ' => '.join(parts[i:]).replace('}', '')
+
+            for files in self.commit_files.values():
+                if not files.get('newfile') or not files.get('file'):
+                    continue
+                old_file = files['file'].replace('{', '')
+                new_file = files['newfile'].replace('}', '')
+                if old_file.startswith(before) and new_file.endswith(after):
+                    return files['newfile']
```
sduenas, line 820:
```suggestion
            for files in self.commit_files.values():
                if not files.get('newfile') or not files.get('file'):
                    continue
                
                old_file = files['file'].replace('{', '')
                new_file = files['newfile'].replace('}', '')
                
                if old_file.startswith(before) and new_file.endswith(after):
                    return files['newfile']
```

perceval/backends/core/git.py
```diff
@@ -791,40 +797,66 @@ def __parse_data_list(self, data, sep=' '):
...
+
+        if rename_path in self.commit_files:
+            # When the file wasn't renamed but contains ' => '
+            return rename_path
```
sduenas, line 824:
```suggestion
        # When the file wasn't renamed but contains ' => '
        if rename_path in self.commit_files:
            return rename_path
```

sduenas (changes requested):
I have some comments and modifications. The PR looks good to me in overall. Very complex scenario.

perceval/backends/core/git.py
jjmerchante: The problem with merge commits is that they behave the opposite way. They show the `numstat` output first and then the `diff-tree`. I've seen some cases where `numstat` reports renames that don't appear in the `diff-tree` output for merge commits. That’s why, in the case of merges, I avoid trying to guess the output.

sduenas: What do you mean with the `diff-tree`? I think we don't get anything like that.

For merge commits, the idea is to find those files that already change on that commit. The merge commit will list all the stats of the commits that are being merged. We parse that but the only files we need to be correct are the ones that change on that commit.

perceval/backends/core/git.py
```diff
@@ -764,25 +764,52 @@ def _handle_trailer(self, line):
...
         self.commit_files[filename]['modes'] = modes
         self.commit_files[filename]['indexes'] = indexes
         self.commit_files[filename]['action'] = data['action']
-        self.commit_files[filename]['file'] = filename
+        self.commit_files[filename]['file'] = data['file']
         self.commit_files[filename]['newfile'] = data['newfile']
 
+    def _handle_pending_files(self):
+        for _, data in self.pending_files.items():
+            if data['file'] in self.commit_files:
```
sduenas, line 780: Due to `data['file']` is used several times, I'd assign this value to a variable for later usage. For small repositories it won't be a problem, but for a large one, it's time that keeps adding up.

Maybe we can even speed it up storing `self.commit_files[filename]` too.

sduenas (changes requested):
The PR looks good. Nevertheless, I added a comment to speed up the process.

sduenas (approved):
LGTM

PR #872, opened 2025-06-17
[jenkins] Add blacklist option for builds

jjmerchante:
Add blacklist functionality for builds in Jenkins integration. To ignore a build, you must define the job name and the build id with the following format: `--blacklist-builds job-id:1234`


sduenas (approved):
LGTM

Issue #875, opened 2025-08-01
[BugzillaREST] requests.exceptions.HTTPError: 400 Client Error: Bad Request for url


VolodymyrBiryuk:
#### System information

- Python 3.12.8
- perceval==1.3.1

#### Issue description
When downloading Bugzilla data, the request fails with the error:

`requests.exceptions.HTTPError: 400 Client Error: Bad Request for url: https://bugzilla.mozilla.org/rest/bug?last_change_time=2023-02-09T03%3A28%3A55Z&limit=25&order=changeddate&include_fields=_all&offset=25&api_key=<your-bugzilla-api-key>`

When executing the URL in the browser, it returns an error as well, however if you increase the offset by amount x (in this example offset=40), it works again.

#### Steps to reproduce the issue

```python
from perceval.backends.core.bugzillarest import BugzillaREST
from datetime import datetime

if __name__ == '__main__':
    init_args={
        'url': 'https://bugzilla.mozilla.org',
        'api_key': '<your-bugzilla-api-key>',
        'max_bugs': 25,
    }

    fetch_args={
        'from_date': datetime(2023, 1, 1)
    }

    backend = BugzillaREST(**init_args)
    iterator = backend.fetch(**fetch_args)
    for i, issue in enumerate(iterator):
        print('i')
```

#### What's the expected result?

- The fetch function should increment offset by amount x until the request returns data again.
- Or take `offset` as initial kwarg in the `fetch` function, so the users can skip it themselves.

#### What's the actual result?

- The fetch function stops with an errro

PR #885, opened 2026-01-21
Handle Bugzilla REST 400 error by stopping pagination

Dev10-sys:
This PR makes BugzillaREST more robust when the API returns
HTTP 400 errors during pagination.

Instead of failing the fetch process, the backend now logs
a warning and stops pagination gracefully by returning an
empty result set.

This matches expected behavior when the API rejects large
offsets and prevents unnecessary crashes.

Related to #875

### DIFF
--- a/perceval/backends/core/bugzillarest.py
+++ b/perceval/backends/core/bugzillarest.py
@@ -459,7 +459,24 @@ class BugzillaRESTClient(HttpClient):
 
         logger.debug("Bugzilla REST client requests: %s params: %s",
                      resource, str(params))
-        r = self.fetch(url, payload=params, headers=headers)
+        try:
+            r = self.fetch(url, payload=params, headers=headers)
+        except requests.exceptions.HTTPError as e:
+            if e.response is not None and e.response.status_code == 400:
+                logger.warning(
+                    "Bugzilla REST returned 400 for resource %s with params %s. "
+                    "This usually indicates an invalid or too-large pagination offset.",
+                    resource, params
+                )
+                raise BackendError(
+                    cause=(
+                        "Bugzilla REST API returned HTTP 400 while paginating. "
+                        "Pagination offsets may be too large or unsupported by the server."
+                    )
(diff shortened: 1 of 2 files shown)


sduenas (changes requested):
Thanks for your patch!

However, I'd rather raise an exception than returning an empty list. An empty list can also mean there's no data retrieved from the given date. When perceval is used as a library it will be hard to know if there was a problem or not.

Dev10-sys:
Thanks for the feedback!

I’ve updated the implementation to raise a BackendError on HTTP 400
instead of returning an empty result set. This avoids ambiguity between
“no data” and “request failure” when perceval is used as a library.

The commit message and behavior have been updated accordingly.
Please let me know if this aligns better with the expected design.

releases/1.4.6.md
```diff
@@ -0,0 +1,3 @@
+## Bug fixes
```
sduenas, line 1: You have to create changelog file. Please read the contributors guidelines: https://github.com/chaoss/grimoirelab/blob/main/CONTRIBUTING_WITH_CODE.md

perceval/backends/core/bugzillarest.py
```diff
@@ -459,7 +459,24 @@ def call(self, resource, params):
...
                      resource, str(params))
-        r = self.fetch(url, payload=params, headers=headers)
+        try:
+            r = self.fetch(url, payload=params, headers=headers)
+        except requests.exceptions.HTTPError as e:
+            if e.response is not None and e.response.status_code == 400:
+                logger.warning(
+                    "Bugzilla REST returned 400 for resource %s with params %s. "
+                    "This usually indicates an invalid or too-large pagination offset.",
+                    resource, params
```
sduenas, line 469:
```suggestion
                    resource, str(params)
```

sduenas (changes requested):
In addition to my comments, you should add a unit test to check this case.

PR #887, opened 2026-02-26
Add secrets manager integration for credential retrieval

alberefe:
## Summary

  - Add support for retrieving backend credentials from external secrets managers (Bitwarden, HashiCorp Vault) instead of passing them on the command line
  - Credentials are fetched in `BackendCommand._pre_init()` and injected into `parsed_args` (user, password, api_token, email, etc.), which backends already consume via HTTP headers/session auth
  - Enable secrets manager arguments (`--secrets-manager`, `--item-name`, `--token-field`, etc.) across 14 backends: bugzilla, bugzillarest, confluence, discourse, gerrit, git, github, gitlab, gitter, googlehits, groupsio, jenkins, rocketchat,
  stackexchange
  - Add optional `hashicorp-manager` dependency group (`hvac >= 2.3.0`) in pyproject.toml
  - Add documentation in README with usage examples for Bitwarden and HashiCorp Vault
  - Add tests for the included functionality

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,6 +143,138 @@ A Perceval Docker image is available at
 Detailed information on how to run and/or build this image can be found
 [here](https://github.com/chaoss/grimoirelab-perceval/tree/main/docker/images/).
 
+## Secrets Manager
+
+Perceval supports retrieving credentials from a secrets manager instead of
+passing them directly on the command line. This is useful for automated
+pipelines and environments where storing credentials in plain text is not
+acceptable.
+
+The following backends support secrets manager authentication: `bugzilla`,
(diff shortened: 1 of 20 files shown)


sduenas (commented):
This looks promising. My main concern is that this only works using perceval from the command line and not as library.  [chronicler](https://github.com/chaoss/grimoirelab-chronicler/) and mordred are two examples of perceval used as a library.

I think it would be better if perceval handles the access to the credentials on the backend and not in the backend command. What do you think?

alberefe:
Closed it cause I'm making changes to adapt to what you said, also including changes to toolkit to modularize the functions and make it so the changes to perceval and others are minimal and everything stays inside toolkit as much as possible.

PR #897, opened 2026-04-28
Add secrets manager integration for credential retrieval 

alberefe:
Integrate Perceval with external secrets managers (Bitwarden, HashiCorp Vault) so credentials can be resolved at runtime instead of being passed as plain-text CLI arguments.                                                                                                                                     
                  
  - Add --secrets-manager, --item-name, and provider-specific CLI flags                                                                                                                                                                                                                                               
  to BackendCommandArgumentParser
  - Add _pre_init() and _build_manager() in BackendCommand to orchestrate credential resolution via grimoirelab-toolkit's resolve_credentials()                                                                                                                                                                                                                                              
  - Enable secrets manager support on 14 backends
  - Add hvac as optional dependency group                                                                                                                                                                                                                                                                             
  - Usage in README

### DIFF
--- a/perceval/backend.py
+++ b/perceval/backend.py
@@ -617,12 +617,14 @@ class BackendCommandArgumentParser:
 
     def __init__(self, backend, from_date=False, to_date=False, offset=False,
                  basic_auth=False, token_auth=False, archive=False,
-                 aliases=None, blacklist=False, ssl_verify=False):
+                 aliases=None, blacklist=False, ssl_verify=False,
+                 secrets_manager=False):
         self._from_date = from_date
         self._to_date = to_date
         self._archive = archive
         self._backend = backend
         self._ssl_verify = ssl_verify
+        self._secrets_manager = secrets_manager
 
         self.aliases = aliases or {}
         self.parser = argparse.ArgumentParser()
@@ -672,6 +674,9 @@ class BackendCommandArgumentParser:
         if ssl_verify:
             group.add_argument('--no-ssl-verify', dest='ssl_verify', action='store_false',
                                help="disable SSL verification")
+
+        if secrets_manager:
+            self._set_secrets_manager_arguments()
 
         self._set_output_arguments()
 
@@ -749,6 +754,39 @@ class BackendCommandArgumentParser:
         group.add_argument('--json-line', dest='json_line', action='store_true',
                            help="produce a JSON line for each output item")
 
+    def _set_secrets_manager_arguments(self):
+        """Activate secret manager arguments parsing"""
+
+        group = self.parser.add_argument_group('secrets manager arguments')
+        group.add_argument('--secrets-manager', dest='secrets_manager',
+                           choices=['bitwarden', 'hashicorp'],
+                           help="Secrets manager service to use for credential retrieval")
+        group.add_argument('--secret-name', dest='secret_name',
+                           help="Name of the secret entry in the secrets manager")
+
+        bw_group = self.parser.add_argument_group('bitwarden arguments')
+        bw_group.add_argument('--bw-client-id', dest='bw_client_id',
+                              default=os.environ.get('PERCEVAL_BW_CLIENT_ID'),
+                              help="Bitwarden API client ID (env: PERCEVAL_BW_CLIENT_ID)")
+        bw_group.add_argument('--bw-client-secret', dest='bw_client_secret',
+                              default=os.environ.get('PERCEVAL_BW_CLIENT_SECRET'),
+                              help="Bitwarden API client secret (env: PERCEVAL_BW_CLIENT_SECRET)")
+        bw_group.add_argument('--bw-master-password', dest='bw_master_password',
+                              default=os.environ.get('PERCEVAL_BW_MASTER_PASSWORD'),
+                              help="Bitwarden master password (env: PERCEVAL_BW_MASTER_PASSWORD)")
+
+        hc_group = self.parser.add_argument_group('hashicorp vault arguments')
+        hc_group.add_argument('--vault-url', dest='vault_url',
+                              default=os.environ.get('PERCEVAL_VAULT_URL'),
+                              help="HashiCorp Vault URL (env: PERCEVAL_VAULT_URL)")
+        hc_group.add_argument('--vault-token', dest='vault_token',
+                              default=os.environ.get('PERCEVAL_VAULT_TOKEN'),
+                              help="HashiCorp Vault authentication token (env: PERCEVAL_VAULT_TOKEN)")
+        hc_group.add_argument('--vault-certificate', dest='vault_certificate',
+                              default=os.environ.get('PERCEVAL_VAULT_CERTIFICATE'),
+                              help="Path to CA certificate for HashiCorp Vault TLS verification "
+                                   "(env: PERCEVAL_VAULT_CERTIFICATE)")
+
 
 class BackendCommand:
     """Abstract class to run backends from the command line.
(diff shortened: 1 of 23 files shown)


alberefe:
I'll be fixing this later please wait before you check so every test and dependency is working fine!

perceval/backend.py
```diff
@@ -749,6 +754,31 @@ def _set_output_arguments(self):
...
+                           help="Bitwarden master password")
+
+        # HashiCorp Vault credentials
+        group.add_argument('--vault-url', dest='vault_url',
+                           help="HashiCorp Vault URL")
+        group.add_argument('--vault-token', dest='vault_token',
+                           help="HashiCorp Vault authentication token")
+        group.add_argument('--vault-certificate', dest='vault_certificate',
+                           help="Path to CA certificate for HashiCorp Vault TLS verification")
+
```
sduenas: I think it would be good if by default these parameters are loaded from env variables.

sduenas: Also, both variable groups are incompatible. I think you should create a different group for each one, but not sure if that's possible using argparse.

perceval/backend.py
```diff
@@ -822,8 +852,76 @@ def run(self):
...
+            from grimoirelab_toolkit.credential_manager.bw_manager import BitwardenManager
+            return BitwardenManager(bw_client_id, bw_client_secret, bw_master_password)
+
+        elif manager_type == 'hashicorp':
+            vault_url = getattr(self.parsed_args, 'vault_url', None)
+            vault_token = getattr(self.parsed_args, 'vault_token', None)
+            vault_certificate = getattr(self.parsed_args, 'vault_certificate', None)
+
+            if not all([vault_url, vault_token]):
```
sduenas, line 926: Isn't the certificate something always required?

sduenas (changes requested):
Please check my initial comments. Also, my concern is this only makes perceval work if you use the backend command but not the library. Please evaluate if it makes sense to integrate it within `fetch` or other methods, or if on the other hand, developers should use grimoirelab toolkit instead.

perceval/backend.py
alberefe: It is not necessary. Depending on the value on the cert, it may skip the TLS verification or not. Or use the system default CA bundle (if None or True). 

We could make it mandatory for more security.

perceval/backend.py
alberefe: The thing is that sirmordred → elk → _pre_init()     -   So right now sirmordred (the orchestrator of everything) is calling the current implementation. It has been made this way cause the rest of grimoirelab calls this cli integration to execute perceval.

If a developer wants, they could use it as a standalone library, but right now grimoirelab process runs through this cli integration and that's the reason it's been inserted in there. 

I'm open to other ways of doing it! Just trying to integrate it while changing the least amount of code possible.

alberefe: I've created different groups for the arguments, so at least in --help they show separate, but I can't make them mutually exclusive with argparse. 

Also implemented the default load from env vars (will upload the changes once we finish discussing the chagnes related to the cli and backend integartion)

perceval/backend.py
alberefe: Not always. Depending if it's not included it skips the TLS verification (False), uses the system CA bundle (True or None) or uses the users inserted certificate.

Would you change it so it's mandatory?

perceval/backend.py
alberefe: It has been inserted this way into perceval cause SirMordred calls Elk, which calls the cli integration to use perceval. 

The toolkit library can be used as a standalone, but this in particular has been done this way cause it's the flow that grimoirelab follows. 

Regarding the parameters:
    - default has been added
    - separated in groups for when using --help, but I can't make them mutually exclusive.

Would you change something about how it's used in the whole grimoirelab system?

perceval/backend.py
sduenas: It's ok this way, I was just wondering. Maybe, with a comment should be enough because for bitwarden you check the three parameters while here you check two.

perceval/backend.py
```diff
@@ -749,6 +754,38 @@ def _set_output_arguments(self):
...
+        group = self.parser.add_argument_group('secrets manager arguments')
+        group.add_argument('--secrets-manager', dest='secrets_manager',
+                           choices=['bitwarden', 'hashicorp'],
+                           help="Secrets manager service to use for credential retrieval")
+        group.add_argument('--item-name', dest='item_name',
+                           help="Name of the item in the secrets manager")
+
+        bw_group = self.parser.add_argument_group('bitwarden arguments')
+        bw_group.add_argument('--bw-client-id', dest='bw_client_id',
+                              default=os.environ.get('PERCEVAL_BW_CLIENT_IDt '),
```
sduenas:
```suggestion
                              default=os.environ.get('PERCEVAL_BW_CLIENT_ID'),
```

perceval/backend.py
sduenas: > The thing is that sirmordred → elk → _pre_init() - So right now sirmordred (the orchestrator of everything) is calling the current implementation. It has been made this way cause the rest of grimoirelab calls this cli integration to execute perceval.
> 
> If a developer wants, they could use it as a standalone library, but right now grimoirelab process runs through this cli integration and that's the reason it's been inserted in there.
> 
> I'm open to other ways of doing it! Just trying to integrate it while changing the least amount of code possible.

Ok. Then let's start with running it only from the command line and we can think later if we change how this feature works.

perceval/backend.py
alberefe: fixed

perceval/backend.py
alberefe: Not sure what you mean by that. To leave  a  comment to make it clear?

sduenas: It doesn't really matter. Let's leave it like this.

sduenas (changes requested):
Please fix the lint errors and add unit test for this functionality.

alberefe:
Getting those errors cause I think it's importing an old version of Toolkit from before it had the las PR changes, is that correct? Not sure how to deal with that or if I'm even correct, but not sure what to do with it.

sduenas:
> Getting those errors cause I think it's importing an old version of Toolkit from before it had the las PR changes, is that correct? Not sure how to deal with that or if I'm even correct, but not sure what to do with it.

We fixed it in #900. Please rebase your branch.

alberefe:
> Please fix the lint errors and add unit test for this functionality.



> Please fix the lint errors and add unit test for this functionality.

Linting errors should be fixed and made tests for the env functinoality.

poetry.lock
```diff
@@ -435,65 +435,65 @@ toml = ["tomli ; python_full_version <= \"3.11.0a6\""]
 
 [[package]]
 name = "cryptography"
-version = "48.0.0"
+version = "47.0.0"
```
sduenas: Why this downgrade?

poetry.lock
```diff
@@ -632,22 +632,19 @@ sphinx-basic-ng = "*"
 
 [[package]]
 name = "grimoirelab-toolkit"
-version = "1.2.6"
+version = "1.2.5"
```
sduenas: Another downgrade. Be sure you are generating the look file with the correct python version.

tests/test_backend.py
```diff
@@ -1122,6 +1136,151 @@ def test_remove_empty_category(self):
...
+        providers = ['bitwarden', 'hashicorp']
+
+        for provider in providers:
+            args = ['--secrets-manager', provider,
+                    '--item-name', 'test-item']
+
+            parsed_args = parser.parse(*args)
+            self.assertEqual(parsed_args.secrets_manager, provider)
+
+    def test_parse_secrets_manager_no_field_args(self):
```
sduenas: I don't understand why this test is needed.

tests/test_backend.py
```diff
@@ -1122,6 +1136,151 @@ def test_remove_empty_category(self):
...
+                                              secrets_manager=True)
+
+        args = ['--secrets-manager', 'bitwarden',
+                '--item-name', 'test-credentials',
+                '--user-field', 'username']
+
+        with self.assertRaises(SystemExit):
+            parser.parse(*args)
+
+    def test_secrets_manager_disabled_by_default(self):
```
sduenas: Rename the test because it's not disabled by default

tests/test_backend.py
```diff
@@ -1122,6 +1136,151 @@ def test_remove_empty_category(self):
...
+        parser = BackendCommandArgumentParser(MockedBackend,
+                                              secrets_manager=False)
+
+        args = ['--secrets-manager', 'bitwarden']
+
+        # Should fail parsing unknown argument
+        with self.assertRaises(SystemExit):
+            parser.parse(*args)
+
+    def test_secrets_manager_arguments_passed_to_backend(self):
```
sduenas: You already wrote this test. The test function name doesn't do what the code does.

tests/test_backend.py
```diff
@@ -1122,6 +1136,151 @@ def test_remove_empty_category(self):
...
+                                                  secrets_manager=True)
+            args = ['--secrets-manager', 'hashicorp',
+                    '--item-name', 'test-item']
+            parsed_args = parser.parse(*args)
+
+        self.assertEqual(parsed_args.vault_url, 'http://vault:8200')
+        self.assertEqual(parsed_args.vault_token, 'hvs.env_token')
+        self.assertEqual(parsed_args.vault_certificate, '/path/to/ca.crt')
+
+    def test_cli_flags_override_env_vars(self):
```
sduenas: You shouldn't test it because this is a feature already available on argparse.

tests/test_backend.py
```diff
@@ -1140,6 +1299,245 @@ def convert_cmd_output_to_json(filepath):
                 buff += line
 
 
+class TestBackendCommandPreInit(unittest.TestCase):
```
sduenas, line 1280: I not sure what the purpose of this set of tests is. They are checking that arguments are parsed but not really injected into the command and replaced using the credentials manager. I think a good set of tests would be that you mock the access to the credential manager and returns the real credentials before running perceval.

sduenas (commented):
Please check my comments. The tests fail because hvac is not installed.

tests/test_backend.py
alberefe: You're right, went over my head.

I modified it to test actual perceval work mocking the toolkit answers.

alberefe:
Fix the poetry.lock file (needed to update some packages) and the testing now properly tests perceval functioning and not toolkit.

releases/unreleased/add-secrets-manager.yml
```diff
@@ -0,0 +1,8 @@
+title: 'Add secrets manager integration for credential retrieval'
+category: added
+notes: >
+    Perceval can now retrieve backend credentials from external
+    secrets managers (Bitwarden, HashiCorp Vault) instead of
+    passing them directly on the command line. This is useful
+    for automated pipelines and environments where storing
+    credentials in plain text is not acceptable.
```
jjmerchante: The format must be:
```
---
title: xxx
category: added
author: Name <email>
issue: null
notes: >
  xxx
```

perceval/backend.py
```diff
@@ -822,8 +860,76 @@ def run(self):
                 logger.exception(f"Error!: {e}", exc_info=self.debug)
 
     def _pre_init(self):
-        """Override to execute before backend is initialized."""
-        pass
+        """Override to execute before backend is initialized.
```
jjmerchante, line 863: Some backends (git, groupsio, hyperkitty, pipermail) don't call `super()._pre_init()` in their `_pre_init()` implementations, and this method won't be executed. Could you fix that?

perceval/backend.py
```diff
@@ -749,6 +754,39 @@ def _set_output_arguments(self):
...
                            help="produce a JSON line for each output item")
 
+    def _set_secrets_manager_arguments(self):
+        """Activate secret manager arguments parsing"""
+
+        group = self.parser.add_argument_group('secrets manager arguments')
+        group.add_argument('--secrets-manager', dest='secrets_manager',
+                           choices=['bitwarden', 'hashicorp'],
+                           help="Secrets manager service to use for credential retrieval")
+        group.add_argument('--item-name', dest='item_name',
```
jjmerchante: Defining `--item-name` as the parameter name could mean many things. What about renaming it to `--secret-name` or something more specific?

Please update it in the README as well.

jjmerchante (changes requested):
Overall, it looks good to me. Just some minor changes

perceval/backend.py
alberefe: Fixed!

perceval/backend.py
alberefe: Updated everywhere I could find!

alberefe:
So, I fixed the things you commented, but I am not sure how to proceed with the poetry.lock conflict. I updated poetry to match the version of main, but now I'm getting other conflicts and not sure if leaving it to you. Not fully understand what the problem is.

jjmerchante (approved):
LGTM

sduenas (changes requested):
Everything looks good to me now. However, would you mind to move the documentation from the README to a file on `/docs/perceval`? I think it's too long for the readme file. You can leave a brief description of its usage and the rest, move it to a file there. You can also link the file from the README for those who need more info.

alberefe:
Created a new docs file and linked to it from the README.md.

sduenas (approved):
LGTM

PR #899, opened 2026-05-05
[github] GitHub App installation token for public repos

### LINKED ISSUE #898: Make installation check optional when using GitHub App credentials to query public data
### Summary

When Percival is configured to authenticate via a GitHub App (application ID + private key), the current token generation logic includes a check that verifies the App is installed in the GitHub organization or account that owns the target repository. This check is unnecessary for querying public data and unnecessarily limits the usability of GitHub Apps.

### Current Behavior

When using a GitHub App to generate an access token, Percival validates that the App is installed in the org or account owning the repository before proceeding. If the App is not installed there, the request fails, even when the target data is entirely public.

### Expected Behavior

GitHub permits applications to query public repository data regardless of where they are installed, in the same way a Personal Access Token (PAT) can be used to access public data on repositories where the token owner has no specific privileges. The installation check should not be a hard requirement for this use case.
GitHub adjusts rate limits based on the identity of the requester, but does not restrict access to public data based on App installation scope.

### Motivation

GitHub Apps offer meaningful advantages over PATs:

* Higher rate limits, which matter at scale
* Finer-grained access control, scoped to specific repositories and permissions
* Better auditability, as actions are attributed to the App rather than a personal account

The current hard requirement effectively prevents users from leveraging these benefits in common real-world scenarios (e.g. querying public data across organizations where installing the App is not feasible or desired). Making the check configurable would give users a practical, robust alternative to PAT-based authentication.

jjmerchante:
Use a non-matching GitHub App installation token when no installation exists for the target repo owner, enabling access to public data with higher rate limits.

Fixes https://github.com/chaoss/grimoirelab-perceval/issues/898

### DIFF
--- a/perceval/backends/core/github.py
+++ b/perceval/backends/core/github.py
@@ -1040,9 +1040,13 @@ class GitHubClient(HttpClient, RateLimitHandler):
         return private_key
 
     def _get_installation_id(self, headers):
-        """Get installation ID given the GitHub login
+        """Return the GitHub App installation ID for the configured owner.
 
-        :param headers: request headers with JWT token
+        For private repositories, the installation must match the owner.
+        For public repositories, any available installation ID is sufficient.
+
+        If no installation matches the owner, the first available installation
+        is returned (if any).
 
         :returns: Installation ID
         """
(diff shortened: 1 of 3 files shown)


sduenas (approved):
LGTM

PR #900, opened 2026-05-05
[tests] Add dev dependencies installation in tests workflow

jjmerchante:
This change adds a step to install development dependencies using Poetry in the GitHub Actions workflow. This ensures that all necessary packages are available for running tests.


sduenas (approved):
LGTM

Issue #903, opened 2026-05-09
Phabricator/Phorge backend uses some frozen/deprecated API endpoints


aklapper:
The Phorge Maniphest backend at https://github.com/aklapper/grimoirelab-perceval/blob/main/perceval/backends/core/phabricator.py uses deprecated Phorge Conduit API endpoints. Quoting that code:

```
    # Methods
    MANIPHEST_TASKS = 'maniphest.search'
    MANIPHEST_TRANSACTIONS = 'maniphest.gettasktransactions'
    PHAB_PHIDS = 'phid.query'
    PHAB_USERS = 'user.query'
```

* `user.query` will in the foreseeable phuture get superseded by `user.search`, see https://we.phorge.it/T16625
* `maniphest.gettasktransactions` in the long run should get replaced with `transaction.search`

jjmerchante:
Hi @aklapper,

Thanks for the heads up.

We have updated the users API endpoint for Phabricator.

The `transaction.search` endpoint returns results that differ from `maniphest.gettasktransactions`, which is why we decided to keep the existing endpoint unchanged.

Furthermore, we created a [pull request](https://github.com/chaoss/grimoirelab-perceval/pull/909) with a new Phorge backend that includes all Phabricator functionality, but uses the new `transaction.search` endpoint.

This ensures that existing functionality relying on Phabricator is not broken.

aklapper:
Thank you!

PR #908, opened 2026-05-22
[phabricator] Update Phabricator user API method

jjmerchante:
Update the user query method from 'user.query' to 'user.search' because the first is deprecated.

Related to https://github.com/chaoss/grimoirelab-perceval/issues/903


sduenas (approved):
LGTM

PR #916, opened 2026-08-05
Add ETag conditional requests to the GitHub backend

### LINKED ISSUE #915: GitHub backend re-fetches pull requests without conditional requests (ETag / If-None-Match)
### Summary

The GitHub backend fetches pull requests by listing issues (`GET /repos/{owner}/{repo}/issues?state=all&sort=updated`) and then fetching each PR and its sub-resources (`GET /pulls/{number}`, reviews, comments, commits, requested reviewers). All of these go through `HttpClient`'s `requests.Session`, which sends no `If-None-Match` and never reads `ETag`. On repeated collection runs, unchanged pull requests are re-downloaded in full. The `--from-date` filter narrows only the *issues list* step; each matched PR's detail resources are re-fetched unconditionally regardless.

### Where

- Base client sends unconditional `session.get()` and never handles `304`: [`perceval/client.py`](https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/client.py) — `HttpClient._fetch_from_remote`.
- PR fetch path: [`perceval/backends/core/github.py`](https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/backends/core/github.py) — `GitHubClient.pulls()` / `issues()` / `fetch_items()` and per-PR `fetch()`.

### Impact on users

GitHub's REST API supports conditional requests: sending the previously returned `ETag` as `If-None-Match` returns `304 Not Modified` when nothing changed, and **`304` responses do not count against the token's primary rate limit**. Perceval already tracks and sleeps on rate limits; conditional requests would let repeated runs revalidate unchanged pages/PRs cheaply instead of spending full budget and bandwidth re-downloading them. For users collecting many repositories or running frequent incremental updates, this reduces rate-limit waits and speeds up collection.

### Suggested change

In `GitHubClient.fetch()`, keep a per-URL `{url: ETag}` map, send `If-None-Match` on repeat GETs, and on a `304` return the previously cached response body. `304` responses still carry `X-RateLimit-*`, so the existing rate-limit handling continues to work; the archive store should skip `304` (no body). No new runtime dependency is required (`requests` is already used). An in-run/session ETag store is a small first step; persisting ETags across runs is a natural follow-up. Collected data stays identical while rate-limit pressure and bandwidth drop.

joshfree:
## Summary

The GitHub backend currently issues every REST request unconditionally. The
`ETag` returned by the GitHub API is never stored or replayed, so each
collection run re-downloads pull requests, issues and their sub-resources in
full even when nothing has changed upstream.

This adds HTTP conditional-request (`ETag` / `If-None-Match`) support to
`GitHubClient`:

- The client caches the `ETag` of each GET response, keyed by URL and query
  parameters.
- On the next identical GET it sends the cached value as `If-None-Match`.
- When the API answers `304 Not Modified`, the cached response body is served,
  avoiding a full re-download.

Rate-limit accounting continues to read the headers of the live response
(including the `304`), so throttling behaviour is unchanged. Per the GitHub
REST API docs, a `304 Not Modified` response does not count against the
token's primary rate limit, so repeated runs consume less rate-limit budget
and bandwidth while producing identical data.

The conditional cache is automatically disabled when an archive is in use (both
while recording and while replaying), so archived/replayed responses keep their
full bodies.

## Testing

- Added `test_conditional_request_not_modified`, which verifies that the first
  request carries no `If-None-Match` header, the second sends the cached
  `ETag`, and a `304` response is served from cache with the original body
  while the rate limit is updated from the live response.
- `python -m unittest tests.test_github` passes locally.

Closes #915

### DIFF
--- a/perceval/backends/core/github.py
+++ b/perceval/backends/core/github.py
@@ -635,6 +635,8 @@ class GitHubClient(HttpClient, RateLimitHandler):
     # API headers
     HAUTHORIZATION = 'Authorization'
     HACCEPT = 'Accept'
+    HIF_NONE_MATCH = 'If-None-Match'
+    HETAG = 'ETag'
 
     # Resource parameters
     PSTATE = 'state'
@@ -669,6 +671,12 @@ class GitHubClient(HttpClient, RateLimitHandler):
         self.max_items = max_items
         self.github_app_id = github_app_id
         self.github_app_pk_filepath = github_app_pk_filepath
+
+        # Conditional-request (ETag) cache: {request key: (etag, response)}.
+        # Only used for live GET requests without an archive, so recorded and
+        # replayed responses keep their full body.
+        self._etag_cache = {}
+        self._etag_enabled = (not from_archive) and (archive is None)
 
         if base_url:
             base_url = urijoin(base_url, 'api', 'v3')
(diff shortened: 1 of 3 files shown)


sduenas (approved):
Thanks for the PR! Nice feature.

LGTM

